PR #10, opened 2020-05-11
Add Redux

grahamhoyes:
Adds Redux and Redux-Toolkit primarily.

I've added a sample slice in `slices/users/userSlice.js`, to serve as a basis for how we structure slices. That one can probably be updated and reused once we have a whoami endpoint.

Tests are included for the selectors and reducers. I decided not to test that the `fetchUserById` thunk dispatches the correct actions to the store, since the behaviour for a simple thunk like this one that just returns a promise is very well defined and clear in the documentation: https://redux-toolkit.js.org/api/createAsyncThunk.


grahamhoyes:
> Personally prefer browser native solutions like fetch but axios definitely is more beginner friendly - shouldn't increase our bundle sizes too much so nbd

Fetch doesn't receive cross-site cookies: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API.

In production that won't be a problem, but it will be a major issue for local development when the backend is running on eg :8000 and the frontend on :3000. Since we're using session authentication (or that's the plan anyway, we can discuss tomorrow), we need to be able to set and receive cross-site cookies in development environments.

grahamhoyes:
> Love these hooks, but we should be firm on how we want to access the redux api from within our components. We should stick with these hooks if we're not gonna use connect(). There are some subtle differences from what I've read between these hooks and connect but should not be a big deal at all for us. Slight possibility that a few extra rerenders might happen here and there but again I don't care and nbd

From what I could find, there are some edge cases where connect is more efficient in terms of unnecessary renders if you provide it with extra options, but with just the default `mapStateToProps` and `mapDispatchToProps` its practically identical to `useState` and `useDispatch`.

I think the hooks are much easier to wrap your head around for beginners, and reduce the amount of boilerplate code you need. We'll create a coding practices doc of sorts tomorrow where we can write down all these decisions.

rayraykay:
> I think the hooks are much easier to wrap your head around for beginners, and reduce the amount of boilerplate code you need. We'll create a coding practices doc of sorts tomorrow where we can write down all these decisions.

Totally agree - I think it's fine if you were to just put that in a standards doc without even the meeting

grahamhoyes:
> Should make it very clear to the group that this is how we want our future reducers/selectors/async functions with a dispatch to be written. I'm not a particularly draconian guy and I don't think anyone else is either, and I think this form will work well for us, seems pretty standard anyway.
> 
> One thing to note is that we should get redux-saga in sooner than later, or stick with the toolkit. I've found this: https://redux-toolkit.js.org/api/createAsyncThunk
> ^don't know much about it but it might do the trick for everything we're trying to do. Saga might be overkill if we just want its async function + dispatch capabilities.

I decided against redux-saga, since it seemed overkill for 99% of our use cases, and since redux-thunk is supported by redux-toolkit right out of the box. We can always revisit it later (with admitedly more difficulty than doing it now like you said), but I don't see much value in adding it now when we might never need it.

Thanks for the link to `createAsyncThunk` - I'll take a peak. The 3 lifecycle types it created are exactly the ones I created in the users slice, I like the consistency in naming this would bring. I just want to check what the action creater it creates is named so we can dispatch them elsewhere if necessary (though I suspect the convention is that you shouldn't be dispatching them anywhere outside of the thunk).

hackathon_site/dashboard/frontend/.yarnrc
```diff
@@ -0,0 +1 @@
+--install.frozen-lockfile true
```
BrennyChao, line 1: does this mean we can just run yarn install from now on instead of the full command or it has nothing to do with our files?

hackathon_site/dashboard/frontend/src/App.js
```diff
@@ -1,22 +1,34 @@
...
+const ConnectedApp = () => (
+    <ReduxProvider store={store}>
+        <StylesProvider injectFirst>
+            <UnconnectedApp />
+        </StylesProvider>
+    </ReduxProvider>
+);
 
-export default App;
+export default ConnectedApp;
```
BrennyChao, line 34: what is App doing diff between connected and unconnected

rayraykay: This is all about Redux - a major library that we're integrating into the front end.

hackathon_site/dashboard/frontend/.yarnrc
grahamhoyes: That's the hope! As long as you're running yarn v1, this should work (we'll need a new solution for yarn 2)

PR #11, opened 2020-05-12
Footer

BrennyChao:


hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+// import "./Footer.module.scss";
```
grahamhoyes, line 2: Can remove

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.js
```diff
@@ -0,0 +1,18 @@
...
+import styles from "./Footer.module.scss";
+
+const Footer = () => (
+    <Container maxWidth={false} disableGutters={true}>
+        <footer className={styles.footer}>
+            <Typography variant="caption" display="block" gutterBottom>
+                Hardware Signout
+            </Typography>
+            <Typography variant="caption" display="block" gutterBottom>
+                © 2020 - build by the web team at MakeUofT (University of Toronto)
```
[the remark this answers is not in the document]
grahamhoyes: Library to make testing when using `new Date()` possible: https://sinonjs.org/releases/v1.17.6/fake-timers/

hackathon_site/dashboard/frontend/src/__tests__/Footer.test.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
```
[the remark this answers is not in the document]
rayraykay: It actually is a common convention from what I found. I personally am used to seeing tests separately, in non-UI code as well. 

I'll change it though - I can't think of a better reason than convention and finding tests easily is a good argument. I also don't really care and am sure the others aren't big on it as well.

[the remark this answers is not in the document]
rayraykay: Also would prefer to have a separate `__tests__` folder but react-scripts actually doesn't look for tests outside `src/`, which is strange but wtv, maybe there's a setting to override it  but couldn't find it and too lazy

hackathon_site/dashboard/frontend/src/__tests__/Footer.test.js
```diff
@@ -0,0 +1,13 @@
...
+import { render } from "@testing-library/react";
+import Footer from "../components/general/Footer/Footer";
+
+jest.mock("@material-ui/core/Typography");
```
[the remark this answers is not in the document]
rayraykay: No reason actually - I'm assuming brenny and aki haven't done much unit testing besides what they've seen in 2nd year ECE so I wanted to go over the basic ropes of unit testing with them and what unit testing exactly is.

I'll change it back, per your suggestions on the other comment.

hackathon_site/dashboard/frontend/src/__tests__/Footer.test.js
```diff
@@ -0,0 +1,13 @@
...
+
+describe("Footer", () => {
+    test("renders", () => {
+        const { asFragment } = render(<Footer />);
+
+        expect(asFragment()).toMatchSnapshot();
```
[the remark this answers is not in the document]
rayraykay: I agree - taking out the mock because just wanted to show the other two what's going on with mocking since it will likely be used in our code.

With regards to the date testing - I think that's better than what's there but I don't see too much point in it besides just testing the date, which we would be mocking anyway. I'll go over creating a mock for the date nonetheless when Brenny makes her updates to the date text, and I'll show them a way of mocking the Date natively without having to install `fake-timers`.

hackathon_site/dashboard/frontend/src/__tests__/Footer.test.js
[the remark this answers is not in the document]
grahamhoyes: Would definitely agree that if we were packaging and distributing just the react source as a package, it would make sense to have tests outside (Material UI does that, for example). But if they have to be in src, might as well make them easy to get to.

We probably can't do anything about `react-scripts` not looking outside of src without ejecting the app.

hackathon_site/dashboard/frontend/src/__tests__/Footer.test.js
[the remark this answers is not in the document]
grahamhoyes: Apologies, I didn't realize that these tests were just for demonstration purposes! Didn't mean to impose.

If the date is mocked (natively or with sinon), and the mock for Typography is removed so the component fully renders, then a snapshot test would be just as good :+1:

hackathon_site/dashboard/frontend/src/__tests__/Footer.test.js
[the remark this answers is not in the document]
grahamhoyes: Roger - I missed that this was for demonstration purposes, good call!

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.js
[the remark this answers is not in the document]
BrennyChao: > Also typo (in the Figma too), should be `built by` not `build by`

LOLLL I just copied from figma

hackathon_site/dashboard/frontend/src/App.js
```diff
@@ -15,6 +16,7 @@ function App() {
                     <p>IEEeeeeeee</p>
                 </div>
             </StylesProvider>
+            <Footer />
```
lisa-sa-li, line 19: Put this above `</StylesProvider>`

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.module.scss
```diff
@@ -0,0 +1,11 @@
+@import "./../../../assets/abstracts/mixins";
+@import "./../../../assets/abstracts/variables";
+
+.footer {
+    text-align: center;
+    background-color: color(secondary);
+    color: color(white);
+    height: 64px;
```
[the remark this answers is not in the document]
BrennyChao: Am I supposed to press resolve conversation, or do you do that, or do we just leave it on here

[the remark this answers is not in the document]
lisa-sa-li: you can press it so you can keep track of what fixes you've completed and how much you have left

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import Typography from "@material-ui/core/Typography";
+import Container from "@material-ui/core/Container";
+import styles from "./Footer.module.scss";
+
+const Footer = () => {
+    let date = new Date();
+
+    if (date.getFullYear() === 2020) {
+        var outputMessage =
```
grahamhoyes, line 10: Stick to using ES6 `let` or `const` instead of `var`. A simpler way of writing this would be:
```javascript
const yearRange = date.getFullYear() === 2020 ? "" : " - " + date.getFullYear();
const outputMessage = "© 2020" + yearRange + " built by the web team at MakeUofT (University of Toronto)";
```

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
```diff
@@ -0,0 +1,38 @@
...
+                return fakeYear;
+            }
+        };
+
+        global.Date = fakeDateClass;
+        const { getByText } = render(<Footer />);
+        const footerText = getByText(/built/, { exact: false });
+
+        expect(footerText).toHaveTextContent(
+            `© 2020 - ${fakeYear} built by the web team at MakeUofT (University of Toronto)`
```
grahamhoyes, line 35: What if the current year is 2020? There should be a test for that as well that the copright just has 2020.

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+import Footer from "./Footer";
+
+describe("Footer", () => {
+    test("renders", () => {
+        const { asFragment } = render(<Footer />);
+        expect(asFragment()).toMatchSnapshot();
```
grahamhoyes, line 9: Since the snapshot has the year in it, this test will fail in 2021. You'll need to mock the date for all tests. See below for a shorter way to mock the date. You may also want to do it outside of the `test` functions (but within `describe`, that way it will apply to all of these tests, then you can just override it for the ones you need it to be a different date.

hackathon_site/dashboard/frontend/src/App.js
grahamhoyes: This needs to be done

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
```diff
@@ -0,0 +1,38 @@
...
+    });
+
+    test("date", () => {
+        const fakeDate = new Date("2021-05-17");
+        const fakeYear = "2021";
```
grahamhoyes, line 14: This doesn't do anything. If we were to change the year in `fakeDate` above, this variable would be wrong and the test would fail. Have the test below check for `fakeDate.getFullYear()`

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
```diff
@@ -0,0 +1,38 @@
...
+        const fakeDateClass = class extends Date {
+            constructor() {
+                super(fakeDate);
+            }
+
+            static now() {
+                return null;
+            }
+
+            getFakeYear() {
```
grahamhoyes, line 25: This doesn't do anything

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
```diff
@@ -0,0 +1,38 @@
...
+    test("renders", () => {
+        const { asFragment } = render(<Footer />);
+        expect(asFragment()).toMatchSnapshot();
+    });
+
+    test("date", () => {
+        const fakeDate = new Date("2021-05-17");
+        const fakeYear = "2021";
+
+        const fakeDateClass = class extends Date {
```
grahamhoyes, line 16: An easier way to do all of the date mocking is:
```javascript
const fakeYear = 2021;
global.Date.getFullYear = jest.fn(() => fakeYear);
```

With that, you can remove this class and the `global.Date = fakeDateClass`, and the `fakeDate` object above.

Rather than mocking the entire date class, this keeps things simple by mocking just the one function this component needs.

hackathon_site/dashboard/frontend/src/App.js
BrennyChao: mmmm my bad I misunderstood, will fix

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.js
BrennyChao: I was looking into it, so `let` and `const ` doesn't allow you to change its values? and that's why you wanted to use those two instead `var` ?

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
BrennyChao: If I were to add a mock for the date function would I still need the 2 tests

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
BrennyChao: so I would do a snapshot, mock the function and do 2 tests for each case where one test would be 2020 and the other one where the year is not 2020?

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.js
grahamhoyes: `const` doesn't let you modify values ("constant" type). Most of the time `const` is what you'll want to use for components etc. `let` does, and in most cases behaves similarly to `var`, just more precise in what scopes it lives at (block scope). Some ES6 linters will also complain about the user of `var`, so we should avoid using it.

Here's a documentation link on `let`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
grahamhoyes: Indeed, you should have two separate tests for this behaviour, because there are two cases: When the year is 2020 (and the copyright just has 2020) and when the year is not 2020 (and the copyright has 2020 - <year>)

grahamhoyes: Implicitly, if you mock the year as 2020 in your first snapshot test, that covers the first case (the second is covered in this test). I would still recommend writing a separate test to be explicit about what's being tested, but that's up to you since it would be redundant.

hackathon_site/dashboard/frontend/src/components/general/Footer/Footer.test.js
grahamhoyes: You need to mock the date function before you render the component; so you should be mocking it for all tests.

See below, technically if you mock the year as 2020 in this test, you've covered the case where the date is 2020 and just need the case where it is not 2020 below. 

Up to you if you'd like to make the 2020 case explicit like the one below, it's not strictly necessary as long as you mock the date as 2020 here.

PR #13, opened 2020-05-13
Sponsor card

lisa-sa-li:


hackathon_site/dashboard/frontend/src/setupTests.js
```diff
@@ -3,3 +3,14 @@
...
+// Fixes the error I get: "matchMedia not present, legacy browsers require a polyfill"
+window.matchMedia =
+    window.matchMedia ||
+    function () {
+        return {
+            matches: false,
+            addListener: function () {},
+            removeListener: function () {},
+        };
+    };
```
BrennyChao, line 16: Is this the test you use if any images are included ??

lisa-sa-li: I added this to fix the error "matchMedia not present, legacy browsers require a polyfill" I was getting when I was trying to run my test. It might be because my yarn version is too ahead of yall. I'll downgrade and try again.

BrennyChao: ohh okay thanks

grahamhoyes:
Needs tests for the Dashboard page and Navbar component. The dashboard scss file is also in the wrong folder, so the page won't build.

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.test.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import Dashboard from "./Dashboard";
+
+it("renders correctly when the dashboard appears with the sponsor card", () => {
+    const { asFragment } = render(<Dashboard />);
+    expect(asFragment()).toMatchSnapshot();
```
grahamhoyes, line 7: If you look at the snapshot for this test, it's massive because it's also rendering everything within this component (testing-library doesn't have shallow rendering). 

A better test (that's more inline with RTL's guiding principles) would be to look for the text you expect to be in eg the title of each card and make sure it's there.

PR #14, opened 2020-05-31
Set up API app and add Authentication endpoint

grahamhoyes:
The API lives under `/api/`.

Authentication is handled by [dj-rest-auth](https://github.com/jazzband/dj-rest-auth), and lives under `/api/auth/`.

No automated tests at this point, since we're using entirely out-of-the-box functionality.


hackathon_site/hackathon_site/settings/base.py
```diff
@@ -25,11 +25,22 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = bool(int(os.environ.get("DEBUG", 0)))
 
+CORS_ORIGIN_REGEX_WHITELIST = [
```
albogdan: you don't need this if you have it in the else statement

albogdan: nvm u removed it in a later commit

PR #20, opened 2020-06-07
Dashboard tables

lisa-sa-li:
For the items prop:
- For checked out items, it takes in the keys: url, name, qty
- Returned items:  url, name, qty, time, condition
- Order pending: url, nae, reqQty, grantQty

The prop status can be: null, "ready", "pending", or "error"


hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.js
```diff
@@ -0,0 +1,256 @@
...
+                    label="Please visit the tech station"
+                    className={styles.chipRed}
+                />
+            );
+        default:
+            return null;
+    }
+};
+
+export const ItemTable = ({ items, title, status }) => (
```
grahamhoyes, line 219: Split this out into 3 separate components (exported so you can use a named imports) for each of the 3 tables. That way you can eliminate most of the switch statement functions above, and not rely on hard-coded strings that are subject to change.

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.js
```diff
@@ -37,6 +89,11 @@ const Dashboard = () => {
                 ))}
                 <ConnectedSponsorCard />
             </Grid>
+            {orderStatus !== null && (
```
grahamhoyes, line 92: `orderStatus && ...`

_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-cayman
```
grahamhoyes, line 1: Remove this file and add it to `.gitignore`

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.js
```diff
@@ -37,6 +89,11 @@ const Dashboard = () => {
...
+                <ItemTable title="Order pending" items={itemsP} status={orderStatus} />
+            )}
+            <ItemTable title="Checked out items" items={itemsC} status={orderStatus} />
+            <ItemTable title="Returned items" items={itemsR} status={orderStatus} />
```
grahamhoyes, line 96: A simple test to check that these are on the page would be helpful (use a `getByText` on the table title or something).

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
```diff
@@ -1,54 +1,52 @@
 import React from "react";
 import { render } from "@testing-library/react";
-import ItemTable from "./ItemTable";
+import {
+    PendingTable,
+    ReturnedTable,
+    CheckedOutTable,
+} from "components/dashboard/ItemTable/ItemTable";
 import { itemsP, itemsR, itemsC } from "pages/Dashboard/Dashboard";
```
grahamhoyes, line 8: Eventually this is going to need to be moved somewhere else, eg into a mock data file once this is hooked up to redux. Doesn't have to be done now. I created a `testing` folder in my PR, so `testing/mockData.js` would probably be helpful.

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
```diff
@@ -1,54 +1,52 @@
...
 
 it("renders correctly when there's an order pending and orange chip", () => {
```
grahamhoyes, line 10: Wrap each of these in a `describe` block to indicate which component is being tested. Makes the file a little bit cleaner, and jest outputs will be grouped by component that way rather than showing every test if they all pass. For example:

```javascript
describe("<PendingTable />", () => {
    it("renders correctly when there's an order pending and orange chip", () => {
        ....
    });

    it(..., () => {
        ...
    });
});

describe("<ItemTable />", () => {
    ...
});
```

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.js
```diff
@@ -0,0 +1,245 @@
...
+                        <col className={styles.widthFixed} />
+                        <col className={styles.widthFixed} />
+                        <col className={styles.widthBuffer} />
+                    </colgroup>
+                    <TableHead>
+                        <TableRow>
+                            <TableCell />
+                            <TableCell align="left">Name</TableCell>
+                            <TableCell align="center">Info</TableCell>
+                            <TableCell align="right">Qty</TableCell>
```
grahamhoyes, line 88: Could probably use the full "Quantity" everywhere unless width is at a premium

PR #22, opened 2020-06-07
Sheets

1999akinori:
For rendering and testing use this for component
\\\ <SideSheetR detail={testDetail} cart={testFunc}/> \\\

copy below to app.js

import SideSheetR from "components/general/SideSheetR/SideSheetR";

const testDetail = {
    name: "Arduino",
    type: "red",
    total: 30,
    available: 29,
    img: "https://www.filterforge.com/more/help/images/size200.jpg",
    tags: "MCU, FPGA",
    manufacturer: "Canakit",
    model_num: "Model 3B+",
    datasheet: "link",
    notes: ["- For micropython ask for image", "- randomnerdtutorials.com"],
    constraints: ["- Max 1 of this item", "- Max 3 microcontroller labelled red"],
    quantity: 3,
};

const testFunc = () => {
alert("testing");
}


hackathon_site/dashboard/frontend/src/App.js
```diff
@@ -6,20 +6,37 @@ import { Provider as ReduxProvider } from "react-redux";
...
+    notes: ["- For micropython ask for image", "- randomnerdtutorials.com"],
+    constraints: ["- Max 1 of this item", "- Max 3 microcontroller labelled red"],
+    quantity: 3,
+};
 
 const UnconnectedApp = () => {
     return (
         <div className="App">
             <Router>
-                <Header />
```
grahamhoyes, line 18: This should stay as the `Header` (which itself contains the `NavBar`)

hackathon_site/dashboard/frontend/src/assets/images/icons/BackArrow.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="16" viewBox="0 0 24 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
grahamhoyes, line 1: Use a MUI icon instead:
```javascript
import ArrowBackIcon from '@material-ui/icons/ArrowBack';
```

hackathon_site/dashboard/frontend/src/assets/images/icons/redtype.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
grahamhoyes, line 1: It would get bloated to upload a bunch of circle SVGs for each category color, you can use some circle-looking icon from MUI and set the color with css based off your `type` prop. Here's one that might work:
```javascript
import FiberManualRecordIcon from '@material-ui/icons/FiberManualRecord';
```

You can search through all the icons here: https://material-ui.com/components/material-icons/

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+import Select from "@material-ui/core/Select";
+import LaunchIcon from "@material-ui/icons/Launch";
+import Grid from "@material-ui/core/Grid";
+
+import styles from "./SideSheetR.module.scss";
+import { ReactComponent as BackArrow } from "../../../assets/images/icons/BackArrow.svg";
+import { ReactComponent as CircleType } from "../../../assets/images/icons/redtype.svg";
+
+const createQuantityList = (number) => {
+    var entry = [];
```
grahamhoyes, line 17: Stick to using `let` or `const` (`let` in this case), not `var`.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+import FormControl from "@material-ui/core/FormControl";
+import Select from "@material-ui/core/Select";
+import LaunchIcon from "@material-ui/icons/Launch";
+import Grid from "@material-ui/core/Grid";
+
+import styles from "./SideSheetR.module.scss";
+import { ReactComponent as BackArrow } from "../../../assets/images/icons/BackArrow.svg";
+import { ReactComponent as CircleType } from "../../../assets/images/icons/redtype.svg";
+
+const createQuantityList = (number) => {
```
grahamhoyes, line 16: This is somewhat inefficient: you have a loop to create a list of ints, which you then `.map` over below. You could just put the for loop directly inline with your `CartSheet` component below.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+};
+
+const CartSheet = ({ cart, detail }) => {
+    const [item, setItem] = React.useState("");
+
+    const handleChange = (value) => {
+        setItem(value);
+    };
+
+    if (cart) {
```
grahamhoyes, line 33: This is fine, but a shorthand that's usually used instead is:

```javascript
return (
    cart && (
        // Everything you return below
    )
);
```

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+                            >
+                                {val}
+                            </MenuItem>
+                        ))}
+                    </Select>
+                </FormControl>
+                <Button
+                    variant="contained"
+                    className={styles.cartButton}
+                    onClick={cart}
```
grahamhoyes, line 52: What is the `cart` prop supposed to be? If it's a function (which this implies it is), give it a more descriptive name about what it's supposed to do.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+            </div>
+            <div className={styles.bodyinfo}>
+                <Typography variant="h3">Datasheet</Typography>
+                <Link
+                    href={detail.datasheet}
+                    rel="noopener"
+                    color="inherit"
+                    underline="none"
+                    target="_blank"
+                    style={{ display: "flex" }}
```
grahamhoyes, line 83: Try not to use inline styles, add a class for this into your scss file.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+                        onClick={() => {
+                            handleChange(false);
+                        }}
+                        role="close"
+                    />
+                    <Typography variant="h2" className={styles.title}>
+                        Product Overview
+                    </Typography>
+                </div>
+                <HeaderSheet detail={detail} />
```
grahamhoyes, line 157: Rather than passing `detail` down into all of these components, I would recomment breaking out props for exactly what information you need and only passing that in. It makes it easier to trace which data is being used where.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+            <div
+                onClick={() => {
+                    handleChange(true);
+                }}
+            >
+                click me
+            </div>
+            <Drawer anchor="right" open={toggle}>
+                <div className={styles.topsheet}>
+                    <BackArrow
```
grahamhoyes, line 146: See my comment above about using an MUI icon instead of an SVG.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+            </div>
+            <img src={detail.img} alt="product" />
+        </div>
+    );
+};
+
+const SideSheetR = ({ detail, cart }) => {
+    const [toggle, setToggle] = React.useState(false);
+
+    const handleChange = (change) => {
```
grahamhoyes, line 131: This may just be a placeholder, but give functions like this a better name, for example `toggleCardState`. Though this function is not really necessary, you could just do inline:
```
onClick={() => {setToggle(false)}}
```

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+        setToggle(change);
+    };
+
+    return (
+        <Grid>
```
grahamhoyes, line 136: Does this Grid need to be here, or can it be a Container? You don't have any `<Grid item>`s.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+                    <Typography variant="h3">Tags</Typography>
+                    <Typography>{detail.tags}</Typography>
+                </div>
+            </div>
+            <img src={detail.img} alt="product" />
+        </div>
+    );
+};
+
+const SideSheetR = ({ detail, cart }) => {
```
grahamhoyes, line 128: Use a full name, `SideSheetRight`.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.test.js
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+import SideSheetR from "./SideSheetR";
+import { render, fireEvent, waitForElementToBeRemoved } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+
+/**
+ * Test the quantity log thing
```
grahamhoyes, line 7: Still need to do this?

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.test.js
```diff
@@ -0,0 +1,77 @@
...
+            <SideSheetR detail={testDetail} cart={addCartMock} />
+        );
+        //Open Drawer
+        const openDrawer = getByText("click me");
+        fireEvent.click(openDrawer);
+
+        const returnArrow = getByRole("close");
+        fireEvent.click(returnArrow);
+
+        await waitForElementToBeRemoved(() => getByText("Product Overview"));
```
grahamhoyes, line 60: Great use of `waitForElementToBeRemoved`!

Could you flesh this test out to have an `expect(...).toBeInTheDocument()`? Technically this will work though, since `getByText` will throw an error if it's not there - so it's up to you if you want to change this for consistency.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+    const handleChange = (value) => {
+        setItem(value);
+    };
+
+    if (cart) {
+        return (
+            <div className={styles.cart}>
+                <FormControl className={styles.form}>
+                    <Select value={item} role="selected">
+                        {createQuantityList(detail.quantity).map((val, i) => (
```
grahamhoyes, line 38: You can do a for loop directly inline here.

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
grahamhoyes: Rename the folder and files too

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.test.js
1999akinori: yes I still need to do this

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.test.js
1999akinori: So replace `getByText` with `expect()toBeInTheDocument()`? That gives me an error saying the element you are waiting to be removed has already been removed.

hackathon_site/dashboard/frontend/src/assets/abstracts/_variables.scss
```diff
@@ -13,6 +13,7 @@ $colors: (
     orangeLight: #ffe3b4,
     red: #b00020,
     redLight: #f9e0e0,
+    grey: #ebf0ef,
```
lisa-sa-li, line 16: remove this cause this color is already in light2

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.module.scss
```diff
@@ -0,0 +1,55 @@
...
+@import "./../../../assets/abstracts/variables";
+
+.topsheet {
+    background-color: color(light2);
+    display: flex;
+    padding: 24px 50px;
+    width: 100%;
+    position: fixed;
+    .backarrow {
+        margin: 0 15px;
```
lisa-sa-li, line 11: change to `margin: auto 15px auto 0;`

hackathon_site/dashboard/frontend/src/assets/images/icons/BackArrow.svg
lisa-sa-li: sad okay :'( Graham convinced me to use toe MUI one

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
lisa-sa-li, line 146: Wrap the icon in <IconButton>

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+    } else {
+        return null;
+    }
+};
+
+const BodySheet = ({ detail }) => {
+    return (
+        <div className={styles.bodysheet}>
+            <div className={styles.bodyinfo}>
+                <Typography variant="h3">Manufacturer</Typography>
```
lisa-sa-li, line 68: The font size on the cart is 14px. Change all these to variant="body2"

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+
+    const handleChange = (value) => {
+        setItem(value);
+    };
+
+    if (cart) {
+        return (
+            <div className={styles.cart}>
+                <FormControl className={styles.form}>
+                    <Select value={item} role="selected">
```
lisa-sa-li, line 37: Set the default value of the input to 1 and give it a label of "Qty". The input is also not the style in the mockup. The examples on the MUI select page is the same

hackathon_site/dashboard/frontend/src/components/general/SideSheetR/SideSheetR.js
```diff
@@ -0,0 +1,165 @@
...
+                                onClick={() => handleChange(val)}
+                                key={i}
+                                role="quantity"
+                            >
+                                {val}
+                            </MenuItem>
+                        ))}
+                    </Select>
+                </FormControl>
+                <Button
```
lisa-sa-li, line 49: The button is missing a shadow, top and bottom padding, a hover. Graham's login PR has the correct styles to refer to

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.js
```diff
@@ -21,6 +21,28 @@ import {
 } from "testing/mockData";
 import { hackathonName } from "constants.js";
 
+import SideSheetRight from "components/general/SideSheetRight/SideSheetRight";
+import ProductOverview from "components/general/ProductOverview/ProductOverview";
+
+const testD = {
```
grahamhoyes, line 27: 1. Give this a better name than `testD`
2. If you'd like to use this, put it under `testing/mockData.js` and import it with the rest, above

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.js
```diff
@@ -61,6 +83,9 @@ const Dashboard = () => {
                 <PendingTable items={itemsPending} status={orderStatus} />
                 <CheckedOutTable items={itemsCheckedOut} />
                 <ReturnedTable items={itemsReturned} />
+                <SideSheetRight title="ProductOverview">
+                    <ProductOverview detail={testD} addCartFunction={testF} />
```
grahamhoyes, line 87: you could just pass in `alert`

hackathon_site/dashboard/frontend/src/components/general/SideSheetRight/SideSheetRight.test.js
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    const addCartMock = jest.fn();
+
+    test("render", () => {
+        const { asFragment } = render(
+            <SideSheetRight detail={testDetail} addCartFunction={addCartMock} />
+        );
+
+        expect(asFragment()).toMatchSnapshot();
```
grahamhoyes, line 35: Would also get rid of the snapshot test, it's not super helpful. Some good things to test would be:
- Displays the children
- Displays the title

And all the other stuff about visibility in my other comments.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,196 @@
...
+                        />
+                    ))}
+                </div>
+            </div>
+            <img src={img} alt="product" />
+        </div>
+    );
+};
+
+const ProductOverview = ({ detail, addCartFunction }) => {
```
grahamhoyes, line 173: minor, would make the second prop `addToCart`

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,196 @@
...
+                <Typography variant="body2">Constraints</Typography>
+                {constraints.map((constraint, i) => (
+                    <Typography key={i}>{constraint}</Typography>
+                ))}
+            </div>
+        </div>
+    );
+};
+
+const HeaderSheet = ({ type, name, total, available, categories, img }) => {
```
grahamhoyes, line 147: Is this something that's standard across all the drawers we'll have (cc @lisa-sa-li)? If so, it could be standardized and moved into `SideSheetRight.js` as well

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,196 @@
...
+
+import { Formik } from "formik";
+
+const createQuantityList = (number) => {
+    let entry = [];
+
+    for (let i = 1; i <= number; i++) {
+        let valueInString = i + "";
+        entry.push(
+            <MenuItem key={i} role="quantity" value={valueInString}>
```
grahamhoyes, line 22: `value={i.toString()}`

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,196 @@
...
+};
+
+export const CartForm = ({
+    availableQuantity,
+    handleSubmit,
+    handleChange,
+    requestFailure,
+    values: { quantity },
+}) => {
+    if (!handleSubmit) {
```
grahamhoyes, line 38: This makes it hard to work on the component while developing, it can be removed. We'll know pretty quickly when an error gets thrown that `handleSubmit` is undefined.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,196 @@
...
+        return false;
+    }
+
+    return (
+        handleSubmit && (
```
grahamhoyes, line 43: This is redundant with the check above, but I would remove it as well. There shouldn't be any real case where this is created without `handleSubmit`.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,196 @@
...
+                    >
+                        ADD TO CART
+                    </Button>
+                </form>
+            </>
+        )
+    );
+};
+
+export const EnhancedCartForm = ({ handleCart, requestFailure, availableQuantity }) => {
```
grahamhoyes, line 78: `handleCart` is sort of a weird prop name. Something like `handleSubmit` is fine, and we can pass in the right function on an outer layer.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,196 @@
...
+    const onSubmit = (formikValues) => {
+        handleCart(formikValues.quantity);
+    };
+
+    return (
+        handleCart && (
```
grahamhoyes, line 84: As above, don't worry about checking that `handleCart` is defined

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
grahamhoyes: Seems I was a little bit confused about what this is. Could you give it a name other than `HeaderSheet`, since it is more than just the header

hackathon_site/dashboard/frontend/src/api/api.js
```diff
@@ -1,9 +1,7 @@
 import axios from "axios";
 
 export const SERVER_URL =
-    process.env.NODE_ENV === "development"
-        ? process.env.REACT_APP_DEV_SERVER_URL.replace(/\/$/, "")
-        : "";
+    process.env.NODE_ENV === "development" ? process.env.REACT_APP_DEV_SERVER_URL : "";
```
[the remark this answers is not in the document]
1999akinori: yeah I made the wrong temporary fix and forgot to fix after i set the environment variable.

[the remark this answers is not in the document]
1999akinori: its all good now :+1:

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
1999akinori: I had this because nothing renders if I don't return false

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
1999akinori: In the figma there are product overviews without an add cart button so I added this for those cases.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
+import React from "react";
```
grahamhoyes, line 1: I think it would make sense to move this component under the `inventory` folder, rather than `general`.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
...
+                img={detail.img}
+            />
+            <DetailInfoSection
+                manufacturer={detail.manufacturer}
+                model_num={detail.model_num}
+                datasheet={detail.datasheet}
+                notes={detail.notes}
+                constraints={detail.constraints}
+            />
+            <CartForm
```
grahamhoyes, line 220: This should be the `EnhancedCartForm`

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
...
+    handleSubmit,
+    requestFailure,
+    availableQuantity,
+}) => {
+    const onSubmit = (formikValues) => {
+        handleSubmit(formikValues.quantity);
+    };
+
+    return (
+        handleSubmit && (
```
grahamhoyes, line 81: No need to have `handleSubmit &&` - you can assume that it would be defined

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
...
+                >
+                    ADD TO CART
+                </Button>
+            </form>
+        </>
+    );
+};
+
+export const EnhancedCartForm = ({
+    handleSubmit,
```
grahamhoyes, line 72: Generally speaking, this should have a more descriptive prop name than just `handleSubmit`, and it would probably match what we call the redux action. For now this is fine since we don't know what that will be called, just wanted to call that out.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
...
+                </Typography>
+                <Link
+                    href={datasheet}
+                    rel="noopener"
+                    color="inherit"
+                    underline="none"
+                    target="_blank"
+                    className={styles.bodyinfoDataSheet}
+                >
+                    <LaunchIcon></LaunchIcon>
```
grahamhoyes, line 146: `<LaunchIcon />`

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
...
+                    <Typography>Link</Typography>
+                </Link>
+            </div>
+            <div className={styles.bodyinfo}>
+                <Typography variant="body2" className={styles.heading}>
+                    Notes
+                </Typography>
+                {notes.map((note, i) => (
```
grahamhoyes, line 154: Notes in the models is just a text field, not a list, so this can just spit the content onto the page like the rest of the fields https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/hardware/models.py#L28

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
...
+                    <Typography key={i} variant="body2">
+                        {note}
+                    </Typography>
+                ))}
+            </div>
+        </div>
+    );
+};
+
+const MainSection = ({ type, name, total, available, categories, img }) => {
```
grahamhoyes, line 164: would rename the `available` prop to `quantityAvailable`, just so it's not confused as a bool

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.module.scss
```diff
@@ -0,0 +1,64 @@
...
+            background-color: color(blueLight);
+            margin: 8px 2.5px;
+        }
+    }
+    & img {
+        margin: 0 11px;
+    }
+}
+
+.detailinfosection {
```
grahamhoyes, line 33: Minor, but give some level of distinction between the words like `.detailInfoSection`. Same for some of the other class names in this file

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
+import React from "react";
+import SideSheetRight from "../SideSheetRight/SideSheetRight";
```
grahamhoyes, line 2: For things that aren't actually being tested, use the absolute import `components/general/SideSheetRight/SideSheetRight`

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
...
+import { ProductOverview, EnhancedCartForm } from "./ProductOverview";
+import {
+    render,
+    fireEvent,
+    waitForElementToBeRemoved,
+    waitFor,
+    waitForElement,
```
grahamhoyes, line 9: This has been deprecated

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
...
+} from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
```
grahamhoyes, line 11: Did you need to import this? This should be loaded by default iirc

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
...
+    test("render", () => {
+        const addCartMock = jest.fn();
+
+        const { asFragment } = render(
+            <SideSheetRight title="Product Overview" isVisible={true}>
+                <ProductOverview detail={testDetail} addCartFunction={addCartMock} />
+            </SideSheetRight>
+        );
+
+        expect(asFragment()).toMatchSnapshot();
```
grahamhoyes, line 38: The snapshot this produced is just an empty div, because of how drawers work (the same will happen for other pop-up components like Dialogs).

I would stick to testing things concretely, and try to avoid snapshot testing if you can. Think of the things that the user would be looking for when using your component (eg all the information is there, buttons work, etc), and test those things explicitly.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
...
+            <MenuItem key={i} role="quantity" value={i.toString()}>
+                {i}
+            </MenuItem>
+        );
+    }
+
+    return entry;
+};
+
+export const CartForm = ({
```
grahamhoyes, line 30: Call this `AddToCartForm`. We'll wind up having more forms to do with the cart (eg for actually submitting it), so this could get confusing.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,228 @@
...
+    };
+
+    return (
+        handleSubmit && (
+            <Formik
+                initialValues={{ quantity: "1" }}
+                onSubmit={onSubmit}
+                validateOnBlur={false}
+                validationOnChange={false}
+                validationSchema={false}
```
grahamhoyes, line 87: You should set some actual validation schema using `Yup`. In your case, the schema would look something like:

```javascript
export const ERROR_MESSAGES = {
   quantityMissing: "Quantity is required"
}

const addToCartFormSchema = Yup.object().shape({
    quantity: Yup.number().required(ERROR_MESSAGES.quantityMissing),
});
```

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
...
+    render,
+    fireEvent,
+    waitForElementToBeRemoved,
+    waitFor,
+    waitForElement,
+} from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+
+describe("Product Overview", () => {
+    const testDetail = {
```
grahamhoyes, line 14: Add this to `testing/mockData.js` with a descriptive name, and import from there. That way we can import it wherever we need it as part of tests.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
...
+        );
+
+        expect(asFragment()).toMatchSnapshot();
+    });
+
+    test("add to cart button calls the correct function", () => {
+        const addCartMock = jest.fn();
+
+        const { getByText, getByRole } = render(
+            <SideSheetRight title="Product Overview" isVisible={true}>
```
grahamhoyes, line 45: Do you actually need to have `SideSheetRight` as part of these tests? You should be able to test a lof of the components within `ProductOverview` (especially the form) individually

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
...
+                <ProductOverview detail={testDetail} addToCart={addCartMock} />
+            </SideSheetRight>
+        );
+
+        const button = getByText("ADD TO CART");
+
+        fireEvent.click(button);
+        expect(addCartMock).toHaveBeenCalledTimes(2);
```
grahamhoyes, line 53: I think this is being caused by not using the Enhanced form in `ProductOverview`. 

However, you are also right about having `onSubmit` set for both the form and the submit button is redundant (and the Enhanced component may just be magically fixing it). It's up to you which you want to do; since the button has `type="submit"` the form's `onSubmit` handler will be called, so you can either remove the `onClick` from the button or the `onSubmit` from the form. Removing from the button probably makes the most sense.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
...
+        const { asFragment } = render(
+            <SideSheetRight title="Product Overview" isVisible={true}>
+                <ProductOverview detail={testDetail} addCartFunction={addCartMock} />
+            </SideSheetRight>
+        );
+
+        expect(asFragment()).toMatchSnapshot();
+    });
+
+    test("add to cart button calls the correct function", () => {
```
grahamhoyes, line 41: This test can just be a part of the `EnhancedCartForm` tests, it doesn't need anything that comes from the rest of the components.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/__snapshots__/ProductOverview.test.js.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
grahamhoyes, line 1: Not a useful snapshot, see above

hackathon_site/dashboard/frontend/src/components/general/SideSheetRight/SideSheetRight.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import Drawer from "@material-ui/core/Drawer";
+import Typography from "@material-ui/core/Typography";
+import IconButton from "@material-ui/core/IconButton";
+import ArrowBackIcon from "@material-ui/icons/ArrowBack";
+import styles from "./SideSheetRight.module.scss";
+
+const SideSheetRight = ({ children, title, isVisible, handleClose }) => {
+    return (
+        <div>
```
grahamhoyes, line 10: This `div` is unnecessary

hackathon_site/dashboard/frontend/src/components/general/SideSheetRight/SideSheetRight.test.js
```diff
@@ -0,0 +1,47 @@
...
+import "@testing-library/jest-dom/extend-expect";
+
+// test for visibility after react is connected to redux store
+
+describe("SideSheetRight", () => {
+    test("title is displayed", () => {
+        const titleTest = "random";
+
+        const { getByText } = render(
+            <SideSheetRight title={titleTest} isVisible={true}></SideSheetRight>
```
grahamhoyes, line 13: When a component has no children, call it like

```javascript
<SideSheetRight title={titleTest} isVisible={true} />
```

hackathon_site/dashboard/frontend/src/components/general/SideSheetRight/SideSheetRight.test.js
```diff
@@ -0,0 +1,47 @@
...
+                handleClose={handleCloseMock}
+            />
+        );
+
+        // close the sidesheetright
+        const backArrow = getByRole("close");
+        fireEvent.click(backArrow);
+
+        // confirm the mock handler function was called
+        expect(handleCloseMock.mock.calls.length).toBe(1);
```
grahamhoyes, line 45:
```javascript
expect(handleCloseMock).toHaveBeenCalledTimes(1);
```

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
grahamhoyes: Actually, something I just remembered - when a form is submitted, it causes the page to reload. You can notice this at the moment by clicking the add to cart button.

This behaviour may be removed once you are using the Enhanced form in the drawer, if formik removes the default behaviour. If it doesn't, then we'll want to have the `onClick` on the button, and remove the `onSubmit` on the form (but I'm pretty sure formik does prevent the default form submit action).

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,132 @@
...
+        );
+
+        const button = getByText("ADD TO CART");
+
+        fireEvent.click(button);
+        expect(addCartMock).toHaveBeenCalledTimes(2);
+    });
+});
+
+describe("<EnhancedCartForm />", () => {
```
grahamhoyes, line 57: The tests this will wind up needing:
- Doesn't call handleSubmit when an option isn't selected, and displays the error message
- When a value is selected, it submits and calls `handleSubmit` with the correct values (you're close on this one).

1999akinori: @lisa-sa-li mentioned that the default option is 1

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
1999akinori: You're right formik removed the refreshing behavior

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
1999akinori: does this also apply to the constraints which is also use `.map`

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
grahamhoyes: Ah, good catch! In those cases, we'd actually want to selectively render the entire form in the `ProductOverview` section, rather than at this level. `ProductOverview` would just have 

```javascript
{addToCart && <CartForm ... />}
```

The reasoning being we can save having to process the formik logic if we know that the form itself isn't going to be there.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
grahamhoyes: From another comment above, since you're trying to capture the case where the form isn't necessary, move this check out into the `ProductOverview` component.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.js
grahamhoyes: Nope - constraints are actually a list. They're derived from the max per team counts for the item and category... some massaging into a single array will be required when we're processing the API response.

hackathon_site/dashboard/frontend/src/components/general/ProductOverview/ProductOverview.test.js
grahamhoyes: Sounds good, as long as there's no way to actually clear the value then you can skip that test (but I would still include it in a yup schema, just for completeness).

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,232 @@
...
+                <Alert className={styles.alert} variant="filled" severity="error">
+                    {requestFailure.message}
+                </Alert>
+            )}
+            <form className={styles.form} onSubmit={handleSubmit}>
+                <FormControl variant="outlined" className={styles.formControl}>
+                    <InputLabel>Qty</InputLabel>
+                    <Select
+                        value={quantity}
+                        role="selecter"
```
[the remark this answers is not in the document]
1999akinori: it looks like it is no longer active

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,229 @@
...
+                    disableElevation
+                >
+                    ADD TO CART
+                </Button>
+            </form>
+        </>
+    );
+};
+
+export const EnhancedCartForm = ({
```
grahamhoyes, line 79: `EnhancedAddToCartForm`, to be consistent with above

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.js
```diff
@@ -0,0 +1,79 @@
+import React from "react";
+import { ProductOverview, EnhancedCartForm } from "./ProductOverview";
+import { render, fireEvent, waitFor } from "@testing-library/react";
+import { productInformation } from "../../../testing/mockData";
```
grahamhoyes, line 4: Use an absolute import `"testing/mockData";`

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.js
```diff
@@ -0,0 +1,229 @@
...
+import MenuItem from "@material-ui/core/MenuItem";
+import FormControl from "@material-ui/core/FormControl";
+import Select from "@material-ui/core/Select";
+import LaunchIcon from "@material-ui/icons/Launch";
+import InputLabel from "@material-ui/core/InputLabel";
+import Chip from "@material-ui/core/Chip";
+import Alert from "@material-ui/lab/Alert";
+
+import styles from "./ProductOverview.module.scss";
+import * as Yup from "yup";
```
grahamhoyes, line 14: Minor again, but move the Yup and formik imports above the styles import. The general pattern I like is to have all imports from external libraries up top (some people even alphebetize them, I don't care for that), a line break, then the imports from your own modules. Helps when looking for things.

PR #26, opened 2020-06-12
Create pages

lisa-sa-li:


hackathon_site/dashboard/frontend/src/pages/NotFound/NotFound.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+// import styles from "./NotFound.module.scss";
+import Header from "components/general/Header/Header";
+import Typography from "@material-ui/core/Typography";
+
+const NotFound = () => {
```
grahamhoyes, line 6: Need to plug this page into the router. Something at the bottom like:

```jsx
<Route exact path="/not-found"  />
<Redirect to="/not-found" />
```

PR #33, opened 2020-06-22
Added nav drawer for mobile view and quantity indicator on the "cart" nav button 

lisa-sa-li:
Deleted notification button and drop down, added it to branch `nav-with-notification`

Navbar now works on mobile, with a drawer activated by a menu button. Navbar requires a new prop, cartQuantity.


hackathon_site/dashboard/frontend/src/components/general/Header/Header.js
```diff
@@ -5,39 +5,71 @@ import AppBar from "@material-ui/core/AppBar";
...
+                </Hidden>
+                {showNavbar && <Navbar cartQuantity={cartQuantity} />}
+
+                <Drawer
+                    anchor="left"
+                    open={mobileOpen}
+                    onClose={toggleMenu}
+                    className={styles.headerDrawer}
+                >
+                    <div className={styles.headerDrawerTop}></div>
```
grahamhoyes, line 63: If you have an empty div (or any other tag), write it as 
```jsx
<div className={styles.headerDrawerTop} />
```

hackathon_site/dashboard/frontend/src/components/general/Header/Header.module.scss
```diff
@@ -30,4 +30,27 @@
...
         }
     }
+
+    &Drawer {
+        &Top {
+            width: 100%;
+            height: 56px;
+            display: block;
+            background-color: color(light1);
+            width: 300px;
```
grahamhoyes, line 40: you have `width: 100%` at the top.

I don't think you should be specifying the width here either way, in one div and expecting the rest to grow to it. Though for whatever CSS magic reason, I can't set `width: 300px` on `&Drawer` and have it behave.

hackathon_site/dashboard/frontend/src/components/general/Header/Header.js
```diff
@@ -5,39 +5,71 @@ import AppBar from "@material-ui/core/AppBar";
...
+                    <div className={styles.headerDrawerHackathon}>
+                        <HackathonTitle />
+                        <Typography variant="body2">{userEmail}</Typography>
+                    </div>
+                    <div className={styles.headerDrawerDivider}></div>
+                    <NavbarDrawer />
```
grahamhoyes, line 69: You should maybe have the check for `showNavbar` here. I got it to show up by setting the state default true and going to the login page, I don't think it's possible to make that happen normally though. 

The drawer unmounts when it's closed, so there isn't a performance difference.

hackathon_site/dashboard/frontend/src/components/general/Header/Header.js
```diff
@@ -5,39 +5,71 @@ import AppBar from "@material-ui/core/AppBar";
...
+    const [mobileOpen, setMobileOpen] = React.useState(false);
+
+    const toggleMenu = () => {
+        setMobileOpen(!mobileOpen);
+    };
+
+    return (
+        <AppBar className={styles.header} position="sticky">
+            <Toolbar className={styles.headerToolbar}>
+                <Hidden implementation="css" smDown={!!showNavbar}>
```
grahamhoyes, line 37: Functionality wise, I have no issues setting this to the JS implementation when I tested. I'd rather we do that, since it reduces how much we have to render (which I think offsets the extra JS processing to check the page width).

hackathon_site/dashboard/frontend/src/components/general/Header/Header.js
```diff
@@ -5,39 +5,71 @@ import AppBar from "@material-ui/core/AppBar";
...
+                    <div className={styles.headerLogo}>
+                        <HackathonTitle />
+                    </div>
+                </Hidden>
+                <Hidden implementation="css" mdUp>
```
grahamhoyes, line 42: Ditto about changing this to js implementation.

hackathon_site/dashboard/frontend/src/components/general/Header/Header.js
```diff
@@ -5,39 +5,71 @@ import AppBar from "@material-ui/core/AppBar";
...
+                    open={mobileOpen}
+                    onClose={toggleMenu}
+                    className={styles.headerDrawer}
+                >
+                    <div className={styles.headerDrawerTop}></div>
+                    <div className={styles.headerDrawerHackathon}>
+                        <HackathonTitle />
+                        <Typography variant="body2">{userEmail}</Typography>
+                    </div>
+                    <div className={styles.headerDrawerDivider}></div>
```
grahamhoyes, line 68: This is unnecessary. Set

```css
border-bottom: 1px solid color(light2);
margin-bottom: 20px;
```

on `headerDrawerHackathon`.

hackathon_site/dashboard/frontend/src/components/general/Header/Header.js
```diff
@@ -5,39 +5,71 @@ import AppBar from "@material-ui/core/AppBar";
...
+                            aria-label="Menu"
+                            className={styles.headerIconBtn}
+                            onClick={toggleMenu}
+                            data-testid="menuBtn"
+                        >
+                            <Menu />
+                        </IconButton>
+                    )}
+                </Hidden>
+                {showNavbar && <Navbar cartQuantity={cartQuantity} />}
```
grahamhoyes, line 55: This has it's own `Hidden` inside. Can we take advantage of the ones in this outer component instead?

hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.js
```diff
@@ -103,4 +98,60 @@ const Navbar = () => (
     </nav>
 );
 
-export default Navbar;
+export const NavbarDrawer = () => (
+    <nav className={styles.navMobile}>
+        <div>
+            <Link to={"/"}>
```
grahamhoyes, line 104: Are all of these routes the same here an in the regular `<Navbar>`, and will that always be the case?

Seems uncessary to me to duplicate all this code if so. You could pull this out to it's own component, with a bit of logic to determine whether `styles.navBtn` or `styles.navMobileBtn` gets applied. Eg
```jsx
const btnClass = mobile ? styles.navMobileBtn : styles.navBtn 
```

hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.module.scss
```diff
@@ -5,22 +5,56 @@
...
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    &Btn {
+        height: 50px;
+        width: 100%;
+        display: flex;
+        justify-content: left;
+        padding: 0px 15px;
```
grahamhoyes, line 36: Minor, but my IDE yells when you specify a unit after 0, since you don't need to. There's a lot of places we already do this though.

hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.js
```diff
@@ -2,31 +2,30 @@ import React from "react";
...
 import { ReactComponent as Inventory } from "assets/images/icons/Hardware.svg";
 // Components
 import Button from "@material-ui/core/Button";
 import IconButton from "@material-ui/core/IconButton";
 import Hidden from "@material-ui/core/Hidden";
 
-const Navbar = () => (
+export const Navbar = ({ cartQuantity }) => (
     <nav className={styles.nav}>
         <Hidden implementation="css" smDown>
```
grahamhoyes, line 19: Likewise to my comment on the header component, can we leverage the `Hidden` there instead of needing it here too?

hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.test.js
```diff
@@ -1,9 +1,24 @@
...
+import { Navbar, NavbarDrawer } from "./Navbar";
 import { withRouter } from "testing";
 
-it("renders correctly when all icons appear", () => {
-    const { asFragment } = render(withRouter(<Navbar />));
-    expect(asFragment()).toMatchSnapshot();
+describe("<Navbar />", () => {
+    it("renders correctly when all icons appear", () => {
+        const { getByText } = render(withRouter(<Navbar cartQuantity={1} />));
+        expect(getByText("Dashboard")).toBeInTheDocument();
```
grahamhoyes, line 9: You're missing buttons here it seems (orders, teams, reports, inventory). I think you'd be fine just using a snapshot for these tests.

hackathon_site/dashboard/frontend/src/testing/mockData.js
```diff
@@ -93,6 +93,9 @@ export const itemsPending = [
 
 export const orderStatus = "pending";
 
+// User email for nav drawer
+export const userEmail = "graham@email.com";
```
grahamhoyes, line 97: I'm honored

hackathon_site/dashboard/frontend/src/components/general/Header/Header.test.js
```diff
@@ -3,18 +3,20 @@ import { render } from "@testing-library/react";
...
-test("renders header with navbar", () => {
-    const { getByText, getByTestId } = render(withRouter(<Header />));
-    expect(getByText(/Hackathon Name/i)).toBeInTheDocument();
-    expect(getByTestId("headerLogo")).toBeInTheDocument();
-    expect(getByText("Dashboard")).toBeInTheDocument(); // Checking if navbar is in header
-});
+describe("<Header />", () => {
+    test("renders header with navbar", () => {
+        const { getByText, getByTestId } = render(withRouter(<Header />));
+        expect(getByText(/Hackathon Name/i)).toBeInTheDocument();
```
grahamhoyes, line 9: We should probably make "Hackathon Name" a constant somewhere. It's going to be a pain to have to replace it in so many places eventually (if you don't want to do this now, that could be a separate piece of work / PR)

lisa-sa-li:
> _No description provided._

hackathon_site/dashboard/frontend/src/components/general/Header/Header.js
```diff
@@ -5,39 +5,84 @@ import AppBar from "@material-ui/core/AppBar";
...
-                        alt="Hackathon logo"
-                        data-testid="headerLogo"
-                    />
-                    <Typography variant="h6" data-testid="hackathonName">
-                        Hackathon Name
-                    </Typography>
+                    <HackathonTitle />
                 </div>
-            </Hidden>
-            <Hidden implementation="css" mdUp>
```
lisa-sa-li, line 27: Used css implementation for now so my test doesnt fail

grahamhoyes:
Not a huge fan of the css implementation of `Hidden`, but I definitely also cannot get it to work with the test suite, so it's fine.

We'll need to connect either the Header or the Navbar to redux to get the cart quantity at some point. Could also move the drawer open/close state into redux if we connected the Header, but that's pretty pointless and just extra work.

PR #34, opened 2020-06-23
Add cart status and category color

grahamhoyes:
Adds a new status "In Cart" for hardware, and changes the "Pending" option to "Submitted".

This lets us track the cart of a team across members, rather than just in the redux store. It also means that to submit an order, all we have to do is set "In Cart" orders to "Submitted" (recall that an "Order" is actually just one item, we don't group them into actual orders).


grahamhoyes:
Closing because:
- We don't need color in the end
- Want hardware to order to be many-to-many, with a custom through model for linking incidents.

PR #35, opened 2020-06-24
Hardware and Category API

albogdan:
Added serializers for Hardware and Category
Added fixtures for Hardware and Category
Added API endpoints for Hardware and Category


hackathon_site/hardware/api_urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path, include
+from hardware import views
+
+
+app_name = "hardware"
+
+urlpatterns = [
+    path("hardware/", views.HardwareListView.as_view()),
```
grahamhoyes, line 8: you should add a `name` attribute to each of these paths:

```python
```

I like to name resource endpoints with plurals, but keep the names singular like that. Read more about why you want to use `name` here: https://docs.djangoproject.com/en/3.0/topics/http/urls/#naming-url-patterns

hackathon_site/hardware/api_urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path, include
```
grahamhoyes, line 1: `include` is unused

hackathon_site/hardware/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+            "datasheet",
+            "quantity_available",
+            "notes",
+            "max_per_team",
+            "picture",
+            "categories",
+            "quantity_remaining",
+        )
+
+    def get_quantity_remaining(self, obj):
```
grahamhoyes, line 24: You can make this a static method (I also like to type annotate for code hints): 

```
@staticmethod
def get_quantity_remaining(obj: Hardware):
    return (
        obj.quantity_available
        - Order.objects.filter(items__hardware__id=obj.id).count()
    )
```

hackathon_site/hardware/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+        return (
+            obj.quantity_available
+            - Order.objects.filter(items__hardware__id=obj.id).count()
```
grahamhoyes, line 27: This isn't entirely correct, because:
- We should be counting items that have been returned as available again
- When items are just in their cart, they have not been checked out yet and are still available for other teams to snag. We decided on this so people can't just hoard things in their cart.

hackathon_site/hardware/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+        )
+
+
+class CategorySerializer(serializers.ModelSerializer):
+    hardware = HardwareSerializer(many=True, read_only=True)
```
grahamhoyes, line 32: This wasn't in our spec, and we don't need it here anyway. Once filters are added to the Hardware endpoint, we can filter them by category that way. This will end up just being extra computational and response payload overhead.

hackathon_site/hardware/views.py
```diff
@@ -1,3 +1,26 @@
 from django.shortcuts import render
 
+from hardware.models import Hardware, Category
+from rest_framework import generics, mixins
+from hardware.serializers import HardwareSerializer, CategorySerializer
+
+
 # Create your views here.
+class HardwareListView(
+    mixins.ListModelMixin, mixins.CreateModelMixin, generics.GenericAPIView
```
grahamhoyes, line 10: `mixins.CreateModelMixin` isn't being used. Remove it untill it's necessary

hackathon_site/hardware/views.py
```diff
@@ -1,3 +1,26 @@
 from django.shortcuts import render
```
grahamhoyes, line 1: Unused, remove

hackathon_site/hardware/views.py
```diff
@@ -1,3 +1,26 @@
...
+):
+    queryset = Hardware.objects.all()
+    serializer_class = HardwareSerializer
+
+    def get(self, request, *args, **kwargs):
+        return self.list(request, *args, **kwargs)
+
+
+class CategoryListView(
+    mixins.ListModelMixin, mixins.CreateModelMixin, generics.GenericAPIView
```
grahamhoyes, line 20: Ditto, `mixins.CreateModelMixin` is unused

hackathon_site/hardware/fixtures/hardware.yaml
```diff
@@ -0,0 +1,146 @@
+-   model: hardware.category
```
grahamhoyes, line 1: Nice. Can you add something to the readme about how to load this, and whether or not it is loaded during testing?

hackathon_site/hardware/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+        fields = (
+            "id",
+            "name",
+            "model_number",
+            "manufacturer",
+            "datasheet",
+            "quantity_available",
+            "notes",
+            "max_per_team",
+            "picture",
```
grahamhoyes, line 19: This guy is a little bit weird coming from the fixture. For the value `"/home/pictures"`, the API returns `"http://localhost:8000/api/hardware/hardware/home/pictures"`. If this is something to fix another time (which it probably is, since we need to configure file uploads), we should call it out in a ticket.

albogdan: Created new issue.

hackathon_site/hardware/serializers.py
albogdan: Is this within the class or no?

grahamhoyes: Yup, exactly where it is, just with the decorator on top. Static methods don't need the "self" parameter.

hackathon_site/hardware/test_api.py
```diff
@@ -0,0 +1,97 @@
...
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = response.json()
+
+        self.assertEqual(expected_response, data["results"][0])
+
+
+class CategoryListViewTestCase(APITestCase):
+    def setUp(self):
+        self.password = "foobar123"
```
grahamhoyes, line 63: Since we're going to have to setup the user and login method for every test ever, it would make sense to define a mixin for that somewhere, perhaps in `hackathon_site/tests.py`. Something like:

```python
class SetupUserMixin:
    def setUp(self):
        self.password = "foobar123"
        self.user = User.objects.create_user(
            username="foo@bar.com",
            password="foobar123",
            first_name="Test",
            last_name="Bar",
            email="foo@bar.com",
        )

    def _login(self):
```

Then this test class would become eg:

```python
class CategoryListViewTestCase(SetupUserMixin, APITestCase):
    def setUp(self):
        super().setUp()
        
        self.category = ...
```

hackathon_site/hardware/test_api.py
```diff
@@ -0,0 +1,97 @@
...
+        self.view = reverse("api:hardware:category-list")
+
+    def _login(self):
+        self.client.login(username=self.user.username, password=self.password)
+
+    def test_user_not_logged_in(self):
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_category_success(self):
```
grahamhoyes, line 83: As above, `test_category_get_success`, (or just `test_get_success`, since the class name indicates the endpoint in question) in case we have other methods on the list view

hackathon_site/hardware/tests.py
```diff
@@ -1,3 +1,71 @@
...
+            "datasheet": "/datasheet/location/",
+            "quantity_available": 4,
+            "quantity_remaining": 4,
+            "notes": None,
+            "max_per_team": 1,
+            "picture": "picture/location",
+        }
+
+        data = hardware_serializer.data
+        self.assertCountEqual(
```
grahamhoyes, line 36: This is redundant if you're asserting equality between `data` and `expected_response`

hackathon_site/hardware/tests.py
```diff
@@ -1,3 +1,71 @@
 from django.test import TestCase
 
-# Create your tests here.
+from hardware.models import Hardware, Category
+from hardware.serializers import HardwareSerializer, CategorySerializer
+
+
+class HardwareSerializerTestCase(TestCase):
```
grahamhoyes, line 7: There should be extra tests here for the different cases that `get_quantity_remaining()` can have. 

- With and without any OrderItems
- With and without any returned items
- With and without anything in the cart

You can combine those together into just a few tests, definitely don't need 8 cases.

hackathon_site/hardware/tests.py
```diff
@@ -1,3 +1,71 @@
...
+        category_serializer = CategorySerializer(self.category)
+        data = category_serializer.data
+        self.assertCountEqual(
+            data.keys(), ("id", "name", "max_per_team", "unique_hardware_count")
+        )
+        expected_response = {
+            "id": 1,
+            "name": "category",
+            "max_per_team": 4,
+            "unique_hardware_count": 0,
```
grahamhoyes, line 69: Should we have a test where this comes out to > 0?

hackathon_site/hardware/fixtures/hardware.yaml
grahamhoyes: Reminder of this, will be helpful for the people less experienced with django

PR #54, opened 2020-06-27
Django test environment

grahamhoyes:
- Adds a new settings file for use during testing, that uses an in-memory sqlite database instead of postgres for testing.
- Updates the python github action to run django tests. Read more about Django's tests [here](https://docs.djangoproject.com/en/3.0/topics/testing/overview/).

Usage:
```bash
python manage.py test --settings=hackathon_site.settings.ci
```

The caveat is that there may be inconsistencies between sqlite and postgres, that would pose an issue in production but will not be caught during testing. These should be caught by a staging environment, and by running your code before you merge it.


grahamhoyes:
> Can't we have test settings and base settings - why do we need base in the init?

The base settings file was redundant, and it was something that I added to begin with (django by default comes with a `settings.py`, not a settings module like I made in preparation for this). Since `__init__.py` just imported from base, I removed the unnecessary layer of complexity. New settings files like `ci.py` can override the default settings from `__init__.py` as required.

PR #55, opened 2020-06-27
Added swagger for managing the API

albogdan:


hackathon_site/hackathon_site/urls.py
```diff
@@ -15,8 +15,41 @@
...
 from django.contrib import admin
 from django.urls import path, include
+from django.conf.urls import url
+
+from rest_framework import permissions
+from drf_yasg.views import get_schema_view
+from drf_yasg import openapi
+
+schema_view = get_schema_view(
+    openapi.Info(
```
grahamhoyes, line 25: Put actual information in here

hackathon_site/hackathon_site/urls.py
```diff
@@ -15,8 +15,41 @@
...
+        r"^swagger(?P<format>\.json|\.yaml)$",
+        schema_view.without_ui(cache_timeout=0),
+        name="schema-json",
+    ),
+    url(
+        r"^swagger/$",
+        schema_view.with_ui("swagger", cache_timeout=0),
+        name="schema-swagger-ui",
+    ),
+    url(
```
grahamhoyes, line 52: Unnecessary, we only need one of swagger ui or redoc (and imo swagger is better)

hackathon_site/hackathon_site/urls.py
```diff
@@ -15,8 +15,41 @@
...
+    public=True,
+    permission_classes=(permissions.AllowAny,),
+    #    url="/api/",
+)
+
 
 urlpatterns = [
     path("admin/", admin.site.urls),
     path("api/", include("api.urls", namespace="api")),
+    url(
```
grahamhoyes, line 42: We don't really need the json/yaml view of the schema, so imo I would leave it out. The Swagger UI is just there at the moment to provide convienent documentation for us.

hackathon_site/hackathon_site/urls.py
```diff
@@ -15,8 +15,41 @@
...
+schema_view = get_schema_view(
+    openapi.Info(
+        title="Snippets API",
+        default_version="v1",
+        description="Test description",
+        terms_of_service="https://www.google.com/policies/terms/",
+        contact=openapi.Contact(email="contact@snippets.local"),
+        license=openapi.License(name="BSD License"),
+    ),
+    public=True,
```
grahamhoyes, line 33: If `public=False`, only endpoints that the user has access to are shown. My preference would be to have that be the case if this is kept live in production.

hackathon_site/hackathon_site/urls.py
```diff
@@ -15,8 +15,41 @@
...
+    permission_classes=(permissions.AllowAny,),
+    #    url="/api/",
```
grahamhoyes, line 35: I see that this is being weird, and that we would need the full URL path. Let's see if the base path fixes itself magically once some other endpoints are included.

hackathon_site/hackathon_site/settings/base.py
```diff
@@ -57,6 +57,7 @@
     "applications",
     "event",
     "hardware",
+    "drf_yasg",
```
grahamhoyes, line 60: Minor, I would put this above all our apps (so right between `dj_rest_auth` and `dashboard`). Just makes it a bit cleaner to pick out our apps from the list if they're all at the bottom.

hackathon_site/hackathon_site/urls.py
```diff
@@ -15,8 +15,41 @@
...
+from rest_framework import permissions
+from drf_yasg.views import get_schema_view
+from drf_yasg import openapi
+
+schema_view = get_schema_view(
+    openapi.Info(
+        title="Snippets API",
+        default_version="v1",
+        description="Test description",
+        terms_of_service="https://www.google.com/policies/terms/",
```
grahamhoyes, line 29: TOS, contact, and license are all optional and can be left out, too

PR #60, opened 2020-06-30
Connected Navbar to store to get pathname. Navbar button is active if you're on that page.

lisa-sa-li:


hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.js
```diff
@@ -11,73 +12,81 @@ import { ReactComponent as Inventory } from "assets/images/icons/Hardware.svg";
...
-            </Button>
-        </nav>
-    );
-};
+const UnconnectedNavbar = ({ cartQuantity, pathname }) => (
+    <nav className={styles.nav}>
+        <div className={styles.navFlexDiv}>
+            <Link to={"/"}>
+                <Button
+                    className={pathname === "/" ? styles.navActive : styles.navBtn}
```
grahamhoyes, line 20: It would be nice to test that this is actually happening. A possibility:
```javascript
it("Adds the active class to only the active page", () => {
        const pagesAndPaths = [
            ["Dashboard", "/"],
            ["Orders", "/orders"],
            // ...
        ];

        pagesAndPaths.map(([label, path]) => {
            const { getByText, container } = render(
                withStore(withRouter(<Navbar pathname={path} />))
            );

            expect(getByText(label).closest("button").className).toMatch(
                new RegExp(styles.navActive)
            );

            expect(container.querySelectorAll(".navActive").length).toBe(1);
        });
    });
```

RTL really doesn't want you to be querying and using classes or ids in your test, since it abstracts away from what the user sees. However, in this case I think it is appropriate, and there isn't any other thing we could do.

grahamhoyes: Had to do a
```javascript
import { styles } from "./Navbar.module.scss";
```
at the top, of course

grahamhoyes: Seems that multiple calls to `render` without cleaning up will keep adding to the DOM, so this will fail for anything beyond the first map item. I can't get [cleanup](https://testing-library.com/docs/react-testing-library/api#cleanup) to work on my end, but you can give it a try. The other solution is to just map `pagesAndPaths` and have the `it` call with the `render` and `expect`s inside it.

hackathon_site/dashboard/frontend/src/testing/helpers.js
```diff
@@ -7,3 +7,9 @@ import store from "slices/store";
 export const withRouter = (component) => <BrowserRouter>{component}</BrowserRouter>;
 
 export const withStore = (component) => <Provider store={store}>{component}</Provider>;
+
+export const withStoreAndRouter = (component) => (
```
grahamhoyes, line 11: Would flip the order of the internal components arround, currently it's acting as `withRouterAndStore` I think semantically. Or just have it call `withStore(withRouter(component))` directly.

hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.test.js
```diff
@@ -11,17 +11,17 @@ describe("<Navbar />", () => {
...
+        ["Orders", "/orders"],
+        ["Teams", "/teams"],
+        ["Reports", "/reports"],
+        ["Inventory", "/inventory"],
+        [`Cart (${cartQuantity})`, "/cart"],
+    ];
 
-        pagesAndPaths.map(([label, path]) => {
+    pagesAndPaths.map(([label, path]) => {
+        it("Adds the active class to only the active page", () => {
```
grahamhoyes, line 24: Can you update this message with a detail about which one it's doing? Otherwise the output will be this same message 6 times. Something like 

```javascript
it(`Adds the active class to ${label} when on ${path}`, ...)
```

Issue #62, opened 2020-07-04
GET: /api/event/users/user/


grahamhoyes:
Return the current request user, or 401 Unauthorized if they aren't logged in.

The model for this endpoint is actually the `event.Profile` model, which has the user within it.

Fields:
- Profile:
    - User:
        - First name
        - Last name
        - Email
        - Groups (serialize to strings rather than ints)
    - Status
    - ID Provided
    - Attended
    - Acknowledge Rules
    - E signature

PR #63, opened 2020-07-04
Current User Endpoint

grahamhoyes:
Solves #62 

- Adds an endpoint to get the current user (actually, their profile): `GET /api/event/users/user/`.
- Sets Token Authentication as the first default authentication class. This allows DRF to return `HTTP 401 Unauthorized` for unauthorized requests, which will make the frontend logic simpler. There's a note in the API readme about this.


hackathon_site/event/models.py
```diff
@@ -1,8 +1,9 @@
 from django.db import models
-from django.contrib.auth.models import User
-
+from django.contrib.auth import get_user_model
 import uuid
 
+User = get_user_model()
```
albogdan, line 5: Why do we need to do it this way and not just import the class?

hackathon_site/event/serializers.py
```diff
@@ -0,0 +1,37 @@
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Group
+from rest_framework import serializers
+
+from event.models import Profile
+
+User = get_user_model()
```
albogdan, line 7: If you're calling this again instead of importing it from the model won't it make two unconnected copies of the class

hackathon_site/event/serializers.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class GroupSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Group
+        fields = ("id", "name")
+
+
+class UserSerializer(serializers.ModelSerializer):
+    groups = GroupSerializer(many=True, read_only=True)
```
albogdan, line 17: Groups - shouldn't this be team; or is this a Django auth class

hackathon_site/hackathon_site/settings/__init__.py
```diff
@@ -122,7 +122,8 @@
 
 REST_FRAMEWORK = {
     "DEFAULT_AUTHENTICATION_CLASSES": [
-        "rest_framework.authentication.SessionAuthentication"
+        "rest_framework.authentication.TokenAuthentication",
+        "rest_framework.authentication.SessionAuthentication",
```
albogdan, line 126: I thought we were going with only token auth?

hackathon_site/event/models.py
grahamhoyes: In Django, it's possible to override the default user model: https://docs.djangoproject.com/en/3.0/topics/auth/customizing/#referencing-the-user-model

Using `get_user_model()` instead of `from django.contrib.auth.models import User` lets forks of this repo specify a custom user model in the settings before they first migrate. The migration file was smart enough to reference `settings.AUTH_USER_MODEL` to begin with, so no changes are needed there: https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/event/migrations/0001_initial.py#L78.

We'll need to change this elsewhere like in the `applications` app.

hackathon_site/event/serializers.py
grahamhoyes: Why do you mean "two unconnected copies of the class"? All `get_user_model()` does in our case is return `django.contrib.auth.models.User`, the default user model.

I could change the import to be `from event.models import Profile, User`, but it is functionally identical.

hackathon_site/event/models.py
```diff
@@ -38,5 +38,11 @@ class Profile(models.Model):
     created_at = models.DateTimeField(auto_now_add=True, null=False)
     updated_at = models.DateTimeField(auto_now=True, null=False)
 
+    def save(self, *args, **kwargs):
```
albogdan, line 41: Is this called by Django automatically? I don't see you calling it anywhere in the model

hackathon_site/hackathon_site/settings/__init__.py
grahamhoyes: The react hardware site still uses session authentication. Having token authentication enabled just gives us another option, and makes `HTTP 401 Unauthorized` possible.

hackathon_site/api/README.md
```diff
@@ -0,0 +1,11 @@
+# API
+This app is largely a wrapper to coordinate the API endpoints for the various other apps.
+
+Generally speaking, each app will implement it's own serializers, views, and URL routes for the API. Views may be in `views.py`, or a dedicated `api_views.py` to differentiated them from regular user-facing views. API url routes should be put in a standard url config file named `api_urls.py`. These are included by `api.urls` with the appropriate namespaces, which itself is included in the base urlconf `hackathon_site.urls` with the `api` namespace.
```
albogdan, line 4: nit: "differentiate" not "differentiated"

hackathon_site/event/models.py
grahamhoyes: This will be called when you do things like

```python
profile = Profile(**kwargs)
profile.save()
```
or
```python
profile = Profile.objects.create(**kwargs)
```

There are a whole host of things that will call either of these methods, like `.create()` or `.update()` for serializers. The notable exceptions are [`bulk_create()`](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#bulk-create) and [`bulk_update()`](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#bulk-update), which we won't be using for this model.

hackathon_site/event/serializers.py
grahamhoyes: Will update to use that import, since it's what the tests also do. It doesn't seem like the best practice since it's almost importing another import, but it makes things clearer and reduces some boilerplate.

hackathon_site/event/views.py
```diff
@@ -1,3 +1,27 @@
...
+
+class CurrentUserAPIView(generics.GenericAPIView, mixins.RetrieveModelMixin):
+    """
+    View to handle API interaction with the current user's Profile
+    """
+
+    serializer_class = ProfileSerializer
+
+    def get_object(self):
+        if self.request.user.is_anonymous:
```
grahamhoyes, line 15: Add a permission class `IsAuthenticated` instead of doing this check manually.

PR #68, opened 2020-07-05
Broken/lost button linking to new page

lisa-sa-li:
"Report broken/lost" button added to CheckedOutItems table. Button links to "Incident Form" page that will contain the form.


hackathon_site/dashboard/frontend/src/App.js
```diff
@@ -47,6 +48,7 @@ const UnconnectedApp = () => {
                     <Route exact path="/reports" component={Reports} />
                     <Route exact path="/inventory" component={Inventory} />
                     <Route exact path="/cart" component={Cart} />
+                    <Route exact path="/Incidentform" component={IncidentForm} />
```
grahamhoyes, line 51: Minor, I don't really like the inconsistency with the caps in this URL but not the others.

hackathon_site/dashboard/frontend/src/assets/abstracts/_exports.scss
```diff
@@ -3,5 +3,5 @@
 :export {
     //  Exports to plug into MUI's theming system
     primary: color(primary);
-    secondary: color(secondary);
+    secondary: color(red);
```
grahamhoyes, line 6: Are you sure you want to do this? This will make MUI use red for all secondary colors (backgrounds, buttons, font colors, etc).

I would instead set the warning and error palettes to be what you want: https://material-ui.com/customization/palette/. You can define them here, then add them to the theme in App.js

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.js
```diff
@@ -117,7 +117,16 @@ export const UnconnectedCheckedOutTable = ({ items, isVisible, toggleVisibility
                                         </IconButton>
                                     </TableCell>
                                     <TableCell align="right">{row.qty}</TableCell>
-                                    <TableCell className={styles.widthBuffer} />
+                                    <TableCell
+                                        align="right"
+                                        className={styles.widthButton}
+                                    >
+                                        <Link to={"/IncidentForm"}>
```
grahamhoyes, line 124: We can't just link them to there, we also need to pass information about which item they are reporting. My suggestion would be something like:
- We have somewhere in the store that stores an incident report
- When this button is clicked, dispatch an action to the store that sets the relevant item, then pushes to the incident form
- If the incident form page is loaded and that value in the store isn't set, send them back to the dashboard

You can leave as is for now, just want to call out that this will change

grahamhoyes: In preparation for that change, I would recommend passing in a prop `handleIncidentReport` or something along those lines that takes in an id (which you can get from `row.id` I believe), and making this a button that calls that. Down the line that would be a redux action / thunk that gets passed in.

In the tests, you can then pass in a spy for that prop and see that it gets called when one of the report broken/lost buttons is clicked.

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
```diff
@@ -77,14 +79,25 @@ describe("<UnconnectedCheckedOutTable />", () => {
...
+                <UnconnectedCheckedOutTable items={itemsCheckedOut} isVisible={false} />
+            )
         );
         expect(getByText(/checked out items/i)).toBeInTheDocument();
         expect(getByText(/show all/i)).toBeInTheDocument();
         itemsCheckedOut.map(({ name }) => {
             expect(queryByText(name)).toBeNull();
         });
     });
+    it("Goes to the incident form when you click the 'Report broken/lost' button", () => {
```
grahamhoyes, line 92: The test does not follow from this description.

lisa-sa-li: oops sorry forgot to change that

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.js
grahamhoyes: Actually, to add to this, I would also pass in `push` as a prop, which the connected component should get in `mapDispatchToProprs` from `connected-react-router`. The `onClick` for the button that goes here should call `reportIncident` (which will be a regular redux action, not a thunk), and then will call `push("/incidentreport")` or whatever the page is called.

The alternative would be to make a `handleIncidentReport` thunk that calls `push`, but that's extra work and testing for the same behaviour.

hackathon_site/dashboard/frontend/src/App.js
lisa-sa-li: should I use kebob or camel case

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
```diff
@@ -142,6 +150,21 @@ describe("Connected tables", () => {
         );
     });
 
+    it("Calls reportIncident when the 'Report broken/lost' button is clicked", () => {
+        const oneRow = [
```
grahamhoyes, line 154: Can you take one of the items from the mocked data instead? That way if we change the structure of the API response, we just need to update the mock data to see what breaks (given we lack integration tests at the moment)

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
```diff
@@ -142,6 +150,21 @@ describe("Connected tables", () => {
...
+            { id: 1, url: "https://i.imgur.com/IO6e5a6.jpg", name: "Arduino", qty: 6 },
+        ];
+        const reportIncidentSpy = jest.fn();
+        const { getByText } = render(
+            withStore(
+                <CheckedOutTable items={oneRow} reportIncident={reportIncidentSpy} />
```
grahamhoyes, line 160: We should also test that `push` was called. Two ways you could do that:
- Actually pass in a mock store to `withStore`, and check that the push action was dispatched.
- Use `UnconnectedCheckedOutTable` instead, and pass in a spy for `push` as well as `reportIncident` (and move it to the right `describe()` section of the test). Would recommend doing that, since there's no need for this test to be on the connected component.

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.js
```diff
@@ -117,7 +122,20 @@ export const UnconnectedCheckedOutTable = ({ items, isVisible, toggleVisibility
                                         </IconButton>
                                     </TableCell>
                                     <TableCell align="right">{row.qty}</TableCell>
-                                    <TableCell className={styles.widthBuffer} />
+                                    <TableCell
+                                        align="right"
+                                        className={styles.widthButton}
+                                    >
+                                        <Button
+                                            color="secondary"
```
grahamhoyes, line 130: Since this is supposed to be red, will this be resolved by #73?

lisa-sa-li: yup, should be red now

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
```diff
@@ -85,9 +85,34 @@ describe("<UnconnectedCheckedOutTable />", () => {
             expect(queryByText(name)).toBeNull();
         });
     });
+
+    it("Calls 'push' when 'Report broken/lost' is clicked ", () => {
```
grahamhoyes, line 89: Trailing space hiding in here

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
```diff
@@ -142,6 +167,19 @@ describe("Connected tables", () => {
         );
     });
 
+    it("Calls reportIncident when the 'Report broken/lost' button is clicked", () => {
```
grahamhoyes, line 170: Move this up into the block with the unconnected table, and switch to using the unconnected component. You could add this right inline with the `it("Calls 'push' when 'Report broken/lost' is clicked ")` test, just by adding `expect(reportIncident).toHaveBeenCalled()`.

Once this is a real redux action, it won't be a prop on the connected component, so this test won't end up here anyway.

PR #74, opened 2020-07-06
Refactored DashCard, TeamCard, CartSummary to create TitledPaper

lisa-sa-li:
- CartSummary takes in `cartQuantity` (same as navbar)
- TitledPaper takes in `title` and `children`
- Updated DashCard, TeamCard, CartSummary component and tests


hackathon_site/dashboard/frontend/src/components/general/TitledPaper/TitledPaper.test.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import TitledPaper from "./TitledPaper";
+
+describe("<TitledPaper />", () => {
+    const titleTest = "Test title";
+    const childTest = "Test child";
+    it("matches snapshot", () => {
```
grahamhoyes, line 8: Imo, this test isn't really necessary. I think in general we should move away from snapshot tests of components like this, since it doesn't add much value. Thoughts?

hackathon_site/dashboard/frontend/src/components/general/TitledPaper/TitledPaper.test.js
```diff
@@ -0,0 +1,26 @@
...
+        expect(asFragment()).toMatchSnapshot();
+    });
+    it("renders text given to it", () => {
+        const { getByText } = render(
+            <TitledPaper title={titleTest}>{childTest}</TitledPaper>
+        );
+        expect(getByText(titleTest)).toBeInTheDocument();
+        expect(getByText(childTest)).toBeInTheDocument();
+    });
+    it("renders no text if not given anything", () => {
```
grahamhoyes, line 21: This is a strange test. There's no possible way that `titleTest` and `childTest` could even get into the component, so this test can't possibly fail. 

Perhaps a useful test in it's place would be that the `Typography` component isn't rendered if no title is given (you'd probably need to give it a `data-testid`).

hackathon_site/dashboard/frontend/src/components/general/TitledPaper/__snapshots__/TitledPaper.test.js.snap
```diff
@@ -0,0 +1,16 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
grahamhoyes, line 1: As above, I don't think this is a particularly helpful snapshot to have, but I defer to your judgement.

hackathon_site/dashboard/frontend/src/components/dashboard/TeamCard/__snapshots__/TeamCard.test.js.snap
```diff
@@ -3,7 +3,7 @@
 exports[`TeamCard render 1`] = `
 <DocumentFragment>
   <h2
-    class="MuiTypography-root MuiTypography-h2 MuiTypography-noWrap"
+    class="MuiTypography-root title MuiTypography-h2 MuiTypography-noWrap"
```
grahamhoyes, line 6: This is a good example of why I don't think snapshot tests are helpful anymore, there wasn't any reason good reason this test should have failed to call out an incorrect change.

hackathon_site/dashboard/frontend/src/components/dashboard/TeamCard/TeamCard.js
```diff
@@ -1,34 +1,29 @@
...
-                    </Container>
-                ))}
-
-                <Container className={styles.lastRow}>
-                    <Button color="primary" onClick={handleEditTeam}>
-                        Edit
-                    </Button>
+        <TitledPaper title={title}>
+            {members.map((member, i) => (
+                <Container className={styles.name} key={i}>
```
grahamhoyes, line 14: Something I learned recently, keys when you map are actually supposed to be globally unique, and this won't be if we did the same thing elsewhere on the dashboard. Something like 
```javascript
key={`team-card-${i}`}
```
should work, assuming it won't complain about there being a space in the key. Or you could use `${member}` in there.

(I know this isn't actually a thing you introduced in this PR, but might as well call it out and fix now)

hackathon_site/dashboard/frontend/src/components/dashboard/DashCard/DashCard.js
```diff
@@ -1,43 +1,34 @@
...
+const DashCard = ({ title, content }) => (
+    <TitledPaper title={title}>
+        {content.map((listItem, i) => (
+            <Link
+                href={listItem.url}
+                rel="noopener"
+                color="inherit"
+                underline="none"
+                target="_blank"
+                key={i}
```
grahamhoyes, line 26: See comment above about keys

hackathon_site/dashboard/frontend/src/components/cart/CartSummary/CartSummary.js
```diff
@@ -0,0 +1,31 @@
...
+import TitledPaper from "components/general/TitledPaper/TitledPaper";
+
+const CartSummary = ({ cartQuantity }) => (
+    <TitledPaper title="Cart Summary">
+        <Container className={styles.qty}>
+            <Typography variant="body2">Quantity</Typography>
+            <Typography variant="body2">{cartQuantity}</Typography>
+        </Container>
+        <Typography variant="body2" className={styles.msg}>
+            Your entire team’s order is here. Before you submit, refresh the page if
```
grahamhoyes, line 15: This smart apostrophy spooks me, was it copy-pasted from somewhere?

lisa-sa-li: probably copied from the Figma, I'll update

hackathon_site/dashboard/frontend/src/components/dashboard/TeamCard/TeamCard.js
grahamhoyes: Jokes, I'm wrong about this. Keys just have to be unique amongs siblings, they recommend you don't use the map index for other reasons: https://reactjs.org/docs/lists-and-keys.html#keys

Perhaps making it
```
key={`${member}-${i}`}
```
would be good, since that doesn't just use the map index and gets around the edge case that two people have the same name.

PR #89, opened 2020-07-12
Add Notistack

### LINKED ISSUE #66: Add notistack
Add notistack to the react app: https://github.com/iamhosseindhv/notistack. Will be helpful for error messages as well as the limited notifications we support.

grahamhoyes:
Resolves #66.

Adds [notistack](https://github.com/iamhosseindhv/notistack). Connects snackbars to the redux ui store with the `SnackbarNotifier` component, so that snacks can be dispatched from redux thunks (for example, in response to an API failure).

It is still possible to display snacks using notistacks' `enqueueSnackbar` (used by `SnackbarNotifier`), but that should be avoided so that we have a source of truth for the snackbars in the redux store.

Adds the following actions to the ui slice:
- `displaySnackbar({message, options})`: Display a snackbar. See the [docs](https://iamhosseindhv.com/notistack/api#enqueuesnackbar-options) for possible options.
- `dismissSnackbar({key})`: Dismissed a snackbar by unique key (which is set as an optional prop on `options`). If no key is given, dismiss all snackbars.
- `removeSnackbar({key})`: Remove a snackbar from the store, to be used after it has been dismissed to cleanup.

To see it in action locally, add a component like this somewhere, for example in `App.js`:
```javascript
import { displaySnackbar as displaySnackbarAction } from "slices/ui/uiSlice";
import { useDispatch } from "react-redux";

const SnackButton = () => {
    const dispatch = useDispatch();

    const displaySnackbar = (...args) => {
        dispatch(displaySnackbarAction(...args));
    };

    return (
        <button
            onClick={() => {
                displaySnackbar({
                    message: "Hello there!",
                    options: {
                        variant: "success",
                    },
                });
            }}
        >
            Dispatch a snackbar
        </button>
    );
};
```


hackathon_site/dashboard/frontend/src/components/general/SnackbarNotifier/snackbarnotifier.test.js
```diff
@@ -0,0 +1,73 @@
+import React from "react";
```
lisa-sa-li, line 1: filename should be camelcase

PR #92, opened 2020-07-22
Update current user endpoint to use user serializer

### LINKED ISSUE #91: Make Current User API View return the user with nested profile
Currently, `/api/event/users/user/` uses a profile serializer, with user nested. This is a problem, because users without profiles - such as admins and tech team - won't be able to log in if this endpoint is the check.

Switch this around to use a user serializer, with profile as a nested object. We can check if that object is not null for whether to let the user access the participant dashboard (in addition to permissions).

grahamhoyes:
Resolves #91 

Updates the /api/event/users/user/ endpoint to return the current user, with nested profile, rather than the other way around. This way users without profiles (tech team, admins) can still use this endpoint.

### DIFF
--- a/hackathon_site/event/serializers.py
+++ b/hackathon_site/event/serializers.py
@@ -10,25 +10,23 @@ class GroupSerializer(serializers.ModelSerializer):
         fields = ("id", "name")
 
 
-class UserSerializer(serializers.ModelSerializer):
-    groups = GroupSerializer(many=True, read_only=True)
-
-    class Meta:
-        model = User
-        fields = ("id", "first_name", "last_name", "email", "groups")
-
-
 class ProfileSerializer(serializers.ModelSerializer):
-    user = UserSerializer(read_only=True)
-
     class Meta:
         model = Profile
         fields = (
             "id",
-            "user",
             "status",
             "id_provided",
             "attended",
             "acknowledge_rules",
             "e_signature",
         )
+
+
+class UserSerializer(serializers.ModelSerializer):
+    profile = ProfileSerializer(read_only=True)
+    groups = GroupSerializer(many=True, read_only=True)
+
+    class Meta:
+        model = User
+        fields = ("id", "first_name", "last_name", "email", "profile", "groups")
--- a/hackathon_site/event/views.py
+++ b/hackathon_site/event/views.py
@@ -1,7 +1,7 @@
 from rest_framework import generics, mixins
 
-from event.models import Profile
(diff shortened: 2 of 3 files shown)


hackathon_site/event/views.py
```diff
@@ -1,18 +1,21 @@
...
 
-    serializer_class = ProfileSerializer
+    queryset = User.objects.all()
+    serializer_class = UserSerializer
 
     def get_object(self):
-        return generics.get_object_or_404(Profile.objects.all(), user=self.request.user)
+        queryset = self.get_queryset()
+
+        return generics.get_object_or_404(queryset, id=self.request.user.id)
```
albogdan, line 18: What's the point of doing 

    queryset = User.objects.all()
    queryset = self.get_queryset()
    return generics.get_object_or_404(queryset, id=self.request.user.id)

when you could just do 

    return generics.get_object_or_404(User.objects.all(), id=self.request.user.id)

hackathon_site/event/serializers.py
```diff
@@ -10,25 +10,23 @@ class Meta:
...
             "status",
             "id_provided",
             "attended",
             "acknowledge_rules",
             "e_signature",
         )
+
+
+class UserSerializer(serializers.ModelSerializer):
+    profile = ProfileSerializer(read_only=True)
```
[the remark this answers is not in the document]
grahamhoyes: The default behaviour is many=None, which is the same as False

hackathon_site/event/views.py
grahamhoyes: `queryset` is an attribute that serializers should have (or override `self.get_queryset()`), it was on omission on my part to hard code `User.objects.all()` into that method the first place.

Various generic reasons include:
- Some things, like the `RetreiveModelMixin`, use the default implementation of `.get_object` which filters the queryset on url kwargs: https://github.com/encode/django-rest-framework/blob/master/rest_framework/generics.py#L75
- It's helpful to have the queryset defined in one place in case you need to modify it. In this case, I should be doing a `.select_related("profile").prefetch_related("groups")`. If the view used it anywhere else, I wouldn't have to update it.
- Sometimes you might want to actually override `self.get_queryset()` instead, to eg return a different queryset depending on url arguments

grahamhoyes: Actually in this case, prefetch_related doesn't help, but often it will save a lot of queries when dealing with many-to-many fields

PR #94, opened 2020-07-23
Add django-debug-toolbar

### LINKED ISSUE #93: Add Django Debug Toolbar
https://django-debug-toolbar.readthedocs.io/en/latest/

This is always helpful to have to make sure queries are being crafted efficiently, amongst other things. Make sure that it works on the DRF browsable API it minimum, and is not accessible outside of debug mode.

grahamhoyes:
Resolves #93 

Adds django-debug-toolbar, which is great for debugging and especially queryset optimization when writing API endpoints, since it shows you what SQL queries are ran, and points out which ones are duplicates of the same type of query.

You'll need to run `python manage.py collectstatic` to make this work.

Also, this uses django's regular templating engine in the background. If we switch to Jinja, we'll need to set the admin site and debug_toolbar to use django's templating engine.

### DIFF
--- a/hackathon_site/hackathon_site/urls.py
+++ b/hackathon_site/hackathon_site/urls.py
@@ -18,11 +18,8 @@ from django.urls import path, include
 from django.conf.urls import url
 from django.conf import settings
 
-from rest_framework import permissions
 from drf_yasg.views import get_schema_view
 from drf_yasg import openapi
-
-PUBLIC = bool(settings.DEBUG)
 
 schema_view = get_schema_view(
     openapi.Info(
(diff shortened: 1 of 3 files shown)


hackathon_site/hackathon_site/urls.py
```diff
@@ -18,19 +18,16 @@
...
-PUBLIC = bool(settings.DEBUG)
-
 schema_view = get_schema_view(
     openapi.Info(
         title="Hardware Hackathon API",
         default_version="v1",
         description="API Endpoint Visualization for Hardware Hackathon",
     ),
-    public=PUBLIC,
+    public=bool(settings.DEBUG),
```
grahamhoyes, line 30: In case anyone is curious about this, I moved this inline to reduce confusion beyond this. On my first attempt, I figured that the `PUBLIC` variable was `not settings.DEBUG` (without actually reading it), so I tried to setup the debug toolbar with

```python
if not PUBLIC:
  ...
```

which follows semantically imo.

PR #101, opened 2020-08-01
Add more styling for ProductOverview and edge case when 0 is in stock

### LINKED ISSUE #99: Add styling for ProductOverview and add edge case for when there's 0 in stock

lisa-sa-li:


hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.js
```diff
@@ -65,15 +69,37 @@ describe("<EnhancedAddToCartForm />", () => {
...
+    });
+
+    test("dropdown values are minimum between quantityAvailable and maxConstraint", () => {
+        const { queryByText, getByRole } = render(
+            <EnhancedAddToCartForm quantityAvailable={3} constraintMax={2} />
+        );
+
+        fireEvent.mouseDown(getByRole("button", { name: "1" }));
+
+        expect(queryByText("3")).not.toBeInTheDocument();
```
grahamhoyes, line 103: This would still pass if the component returned nothing at all. Adding an extra check that "2" is in the document would fix that (I tested locally and that does pass).

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.js
```diff
@@ -18,16 +19,34 @@ import {
...
 const Dashboard = () => {
     const reportIncident = (id) => {
         alert("Reports incident for component of id " + id);
     };
+
+    const addToCart = () => {
+        alert("add to cart");
+    };
+
+    const [mobileOpen, setMobileOpen] = React.useState(true);
```
grahamhoyes: why `mobileOpen`? Is this from copying from the header component?

grahamhoyes: If you want to keep this around until we can actually hook it up to redux and make the info buttons open it, make the default `false`.

hackathon_site/dashboard/frontend/src/components/general/SideSheetRight/SideSheetRight.js
```diff
@@ -8,19 +8,19 @@ import styles from "./SideSheetRight.module.scss";
 const SideSheetRight = ({ children, title, isVisible, handleClose }) => {
     return (
         <Drawer anchor="right" open={isVisible}>
```
grahamhoyes: Can you add `onClose={handleClose}` as a prop? That way clicking outside of the drawer will also cause it to close.

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.js
```diff
@@ -38,40 +38,54 @@ const createQuantityList = (number) => {
...
+                <div className={styles.formButton}>
+                    <Button
+                        variant="contained"
+                        color="primary"
+                        fullWidth={true}
+                        size="large"
+                        type="submit"
+                        onClick={handleSubmit}
+                        disabled={dropdownNum === 0}
+                        data-testid="qtySelect"
```
grahamhoyes: Should this test id be on the `<Select>` above?

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.js
```diff
@@ -38,40 +38,54 @@ const createQuantityList = (number) => {
...
         <>
             {requestFailure && (
-                <Alert className={styles.alert} variant="filled" severity="error">
+                <Alert variant="filled" severity="error">
                     {requestFailure.message}
                 </Alert>
             )}
             <form className={styles.form} onSubmit={handleSubmit}>
                 <FormControl variant="outlined" className={styles.formControl}>
                     <InputLabel>Qty</InputLabel>
```
grahamhoyes: In order to be used properly with `getByLabelText` (and for accessibility reasons), this should have an `id`

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.js
```diff
@@ -38,40 +38,54 @@ const createQuantityList = (number) => {
...
                     <Select
```
grahamhoyes, line 65: To correspond with the id of the `InputLabel` above, this should have a `labelId` prop

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.js
```diff
@@ -65,15 +69,37 @@ describe("<EnhancedAddToCartForm />", () => {
...
         expect(getByText(requestFailure.message)).toBeInTheDocument();
     });
+
+    test("button and select are disabled if quantityAvailable is 0", () => {
+        const { getByText, getByTestId } = render(
+            <EnhancedAddToCartForm quantityAvailable={0} />
+        );
+
+        const button = getByText("Add to cart");
+        const select = getByTestId("qtySelect");
```
grahamhoyes: Because the test id is on the wrong element, this is actually giving you the button.

Once you add a label to the select and input label, you can use `getByLabelText("Qty")`. However since this isn't an actual html input (under the hood, it isn't actually a `<select>`), you won't be able to use `.toBeDisabled()`. You could check for the `aria-disabled="true"` tag, or the `.Mui-disabled` class.

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.js
```diff
@@ -115,61 +131,44 @@ const DetailInfoSection = ({
...
+                rel="noopener"
+                target="_blank"
+                startIcon={<LaunchIcon />}
+            >
+                Link
+            </Button>
+            <Typography variant="body2" className={styles.heading}>
+                Notes
+            </Typography>
+            <Typography>{notes}</Typography>
```
grahamhoyes: Since the notes can have new lines in them, you could make this 

```javascript
{notes.split("\n").map((note, i) => (
     <Typography key={i}>{note}</Typography>
))}
```

otherwise it won't listen to the `\n` characters.

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.js
```diff
@@ -74,6 +93,10 @@ const Dashboard = () => {
                     reportIncident={reportIncident}
                 />
                 <ReturnedTable items={itemsReturned} />
+
+                <Button variant="contained" onClick={toggleMenu}>
```
grahamhoyes, line 97: Make sure we remove this eventually, but it's fine to keep in here for now until we have it hooked up to redux

grahamhoyes (approved):

Issue #98, opened 2020-07-31
Create landing page template and static files dir for Event app


grahamhoyes:
This is where the landing page of the site will live.

While the forms for the applications will be in the Application app, the actual templates for them will likely live within the event app, so that they can easily reuse static files (though it is possible to use cross-app static files).

grahamhoyes:
Resolved by #103

PR #103, opened 2020-08-02
Add Jinja2 and Create Landing Page

### LINKED ISSUE #97: Switch default templating engine to Jinja2
Set Jinja2 as the default templating engine: https://docs.djangoproject.com/en/3.0/topics/templates/.

Jinja2 is in general more powerful than Django's templating system, and we could reuse some of the macros we wrote when working in flask last year. 

We'll need to make sure that the Admin site, Django Debug Toolbar, and the DRF browsable API still use Django's templating system (as well as any relevant libraries we add in the future).

grahamhoyes:
Resolves #97 and #98 

- Sets Jinja2 as the default templating backend, while leaving existing apps that rely on Django's templating engine unaffected. Jinja2 templates should go within `<app_name>/jinja2/<app_name>/`.
- Creates a base template, `event/jinja2/event/base.html`.
- Adds basic static files for the event app.

Further points of investigation:
- What can we use to lint and autoformat the jinja2 templates, if anything? Something that could function as a check stage would be preferred.


hackathon_site/event/views.py
```diff
@@ -1,9 +1,14 @@
+from django.views.generic import TemplateView
 from rest_framework import generics, mixins
 
 from event.models import User
 from event.serializers import UserSerializer
 
 
+class IndexView(TemplateView):
```
albogdan, line 8: Don't you need a get or post method here

hackathon_site/event/tests.py
```diff
@@ -22,3 +24,10 @@ def test_creates_team_if_not_provided(self):
         profile = Profile.objects.create(user=self.user, status="Accepted")
         self.assertEqual(Team.objects.count(), 1)
         self.assertEqual(Team.objects.first(), profile.team)
+
+
+class IndexViewTestCase(TestCase):
```
albogdan, line 29: I think it'd be good to have test cases for HTTP_200 regardless of if you're logged in or not (aka add a test case for logged in please)

hackathon_site/event/urls.py
```diff
@@ -3,4 +3,4 @@
 
 app_name = "event"
 
-urlpatterns = [path("", IndexView.as_view())]
+urlpatterns = [path("", IndexView.as_view(), name="index")]
```
albogdan, line 6: shouldn't it be 

`urlpatterns = [path("/", IndexView.as_view(), name="index")]`

hackathon_site/event/views.py
grahamhoyes: Nope - that's what `TemplateView` does: https://github.com/django/django/blob/master/django/views/generic/base.py#L157

hackathon_site/event/tests.py
grahamhoyes: As we discussed externally, there's no way that being logged in could effect this view. So while technically for completeness it wouldn't hurt to have such a test, that opens us up to over-testing for things that can never happen.

hackathon_site/event/urls.py
grahamhoyes: Django treats that as a leading slash, and will raise a warning. Even if you navigate to `/` in the browser, this route will capture it.

PR #115, opened 2020-08-11
Added sass, materialize CDN, and prettier to event template directory

lisa-sa-li:


hackathon_site/event/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "event",
+  "version": "1.0.0",
+  "license": "MIT",
+  "dependencies": {
+    "node-sass": "^4.14.1",
+    "prettier": "^2.0.5"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "scss": "node-sass --watch static/event/styles/scss -o static/event/styles/css"
```
grahamhoyes, line 11: Some problems:
- This won't compile scss that's already in the directory, unless you trigger the file watcher by changing something (at least it looks that way for me).
- This leaves the watcher process running all the time. We'll also need a way to compile scss one-time, as part of the eventual release process.

I would recommend making this command (`scss`) just run node-sass one time. Then add another command, `scss-watch`, that does exactly this and watches for changes.

PR #127, opened 2020-08-18
Adds Formik to inventory filter

lisa-sa-li:
Adds Formik to inventory filter


grahamhoyes (commented):
Not sure how hard this is to change, but part of the scroll bar is hidden by the top bar
image

Given that we're using a drawer, not sure what the fix is if any

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
```diff
@@ -1,100 +1,183 @@
...
+                    label={item.name}
+                    checked={field.value.includes(item.name)}
+                />
+                <Chip label={item.qty} className={styles.filterCategoryChip} />
+            </div>
+        ))}
+    </FormGroup>
+);
+
+const CheckboxAvailability = ({ field, name, ...props }) => (
```
grahamhoyes: No need for the name prop

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
```diff
@@ -1,100 +1,183 @@
...
+            name={field.name}
+            value="In stock"
+            control={<Checkbox color="primary" />}
+            checked={field.value}
+        />
+    </FormGroup>
+);
+
+export const orderByOptions = [
+    { value: "Default", label: "Default" },
```
grahamhoyes: To make this easier for whoever plugs this into the API, can you make the values something different than the labels? I'm not sure what they'll end up being in the API, but I think something like `name`, `-name`, `stock`, `-stock` and empty string for default if I recall correctly.

Along with that, make sure that `handleSubmit` is being called with the values, not the labels (I think it is, just want to make sure).

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.js
```diff
@@ -1,60 +1,134 @@
...
-import { fireEvent } from "@testing-library/react";
+import EnhancedInventoryFilter, { orderByOptions } from "./InventoryFilter";
+import { render, fireEvent, waitFor } from "@testing-library/react";
 import { inventoryCategories } from "testing/mockData";
 
-describe("<InventoryFilter />", () => {
-    const applyFilter = jest.fn();
-    const removeFilter = jest.fn();
+describe("<EnhancedInventoryFilter />", () => {
+    it("Calls handleSubmitSpy when the 'Apply' button is clicked", async () => {
```
grahamhoyes: Minor, should just call this "Calls handleSubmit ...", rather than "handleSubmitSpy" (same for the ones below)

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.js
```diff
@@ -1,60 +1,134 @@
...
-            expect(queryByText(o)).toBeTruthy();
+        for (let o of orderByOptions) {
+            expect(queryByText(o.label)).toBeTruthy();
         }
         expect(getByText("Order by")).toBeInTheDocument();
         expect(getByText("Availability")).toBeInTheDocument();
         expect(getByText("Categories")).toBeInTheDocument();
     });
+
+    it("Submits the form, the clears it, and recieves the expected values", async () => {
```
grahamhoyes: A few typos in this description

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.js
```diff
@@ -1,60 +1,134 @@
...
+        await waitFor(() => {
+            expect(handleResetSpy).toHaveBeenCalledWith({
+                orderBy,
+                inStock,
+                inventoryCategories,
+            });
+        });
+    });
+
+    it("Looks for loader on Apply button", () => {
```
grahamhoyes: Maybe `it("Displays a loading wheel on Apply button when loading", ...)`? Generally you should describe what the component being tested is doing, not what the test itself is looking for if that makes sense. Same for below

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.js
```diff
@@ -1,60 +1,134 @@
...
+    });
+
+    it("Looks for loader on Clear all button", () => {
+        const { getByTestId, queryByTestId } = render(
+            <EnhancedInventoryFilter isApplyLoading={false} isClearLoading={true} />
+        );
+
+        expect(queryByTestId("circularProgressApply")).not.toBeInTheDocument();
+        expect(getByTestId("circularProgressClear")).toBeInTheDocument();
+    });
```
grahamhoyes, line 143: Can you also test that the apply/clear buttons are disabled when either of them is loading? The easiest thing would be to just add checks for those buttons being disabled into the two loading tests above

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
lisa-sa-li: yes it's called with values

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
```diff
@@ -1,105 +1,188 @@
...
+);
 
-const InventoryFilter = ({ categories, applyFilter, removeFilter }) => (
+const CheckboxCategory = ({ field, options }) => (
+    <FormGroup {...field} name={field.name}>
+        {options.map((item, i) => (
+            <div className={styles.filterCategory} key={i}>
+                <FormControlLabel
+                    name={field.name}
+                    value={item.name}
```
grahamhoyes: We should be using the category ID here as the value, since that's what the API takes. I think this works:

```javascript
<FormControlLabel
    name={field.name}
    value={item.id}
    control={<Checkbox color="primary" />}
    label={item.name}
    checked={field.value.includes(item.id.toString())}
/>
```

Then in mockData.js:

```javascript
export const inventoryCategories = [
    { id: 1, name: "MCU", qty: 12 },
    { id: 2, name: "MCU_limit_3", qty: 6 },
    { id: 3, name: "FPGA", qty: 9 },
    { id: 4, name: "Sensors", qty: 21 },
    { id: 5, name: "Sensors_limit_2", qty: 3 },
    { id: 6, name: "Peripherals", qty: 389 },
    { id: 7, name: "1080p_cameras", qty: 4 },
    { id: 8, name: "Grove", qty: 22 },
    { id: 9, name: "Grove_style_boards", qty: 12 },
    { id: 10, name: "Grove_style_modules", qty: 10 },
];
```

Only problem with this is that the ID values become strings in the form, which will require massaging to ints for the API. Is there any way around that?

grahamhoyes (commented):
One last thing about using category IDs. Other than that, LGTM

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
lisa-sa-li: Didn't find out how to do it without parsing all the ids before submitting the form values

grahamhoyes (approved):

PR #131, opened 2020-08-21
created CurrentTeamAPIView and serializer with test

1999akinori:
- added new `teams/team` urlpath to `event/api_urls/`
- created `CurrentTeamAPIView` at `event/views.py`
- test for user not logged in
- test for user having no profile and trying to access their team
- test for user having a profile and trying to access their team
- test for TeamSerializer, whether the serializer is outputting what we expect.

CurrentUserAPIView test cases are failing since the profile serializer was changed to include team.

### DIFF
--- a/hackathon_site/event/test_api.py
+++ b/hackathon_site/event/test_api.py
@@ -40,37 +40,16 @@ class CurrentUserTestCase(SetupUserMixin, APITestCase):
     def test_user_has_profile(self):
         self._login()
         response = self.client.get(self.view)
-        user_expect = User.objects.get(pk=self.user.pk)
-        serializer = UserSerializer(user_expect)
+        user_expected = User.objects.get(pk=self.user.pk)
+        serializer = UserSerializer(user_expected)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.json(), serializer.data)
-    
-    def test_user_uses_profile_and_group_serializer(self):
-        self._login()
-        response = self.client.get(self.view)
-        res_json = response.json()
-        
-        profile_expect = Profile.objects.get(pk=self.profile.pk)
-        profile_serializer = ProfileSerializer(profile_expect)
-
-        group_expect = Group.objects.get(pk=self.group.pk)
-        group_serializer = GroupSerializer(group_expect)
-
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(res_json['profile'], profile_serializer.data)
-        self.assertEqual(res_json['groups'], [group_serializer.data]) # need to figure out why its different
-
-
-    
 
 
 
 
 
 class CurrentTeamTestCase(SetupUserMixin, APITestCase):
-    # When not logged in, the response is 401 unauthorized
-    # When the user has no profile and tries to access the team, the response should be 404 Not Found (it is not possible to have a profile without a team, so no need to test that case)
-    # When logged in with a profile and tries to access the team, the correct response is returned. As an example for doing that with the serializer test separate search for the test_get_valid_single puppy method on this page:
     def setUp(self):
         super().setUp()
         self.group = Group.objects.create(name="Test Users")
@@ -98,8 +77,8 @@ class CurrentTeamTestCase(SetupUserMixin, APITestCase):
         # When user has a profile and attempts to access the team, then the user should get the correct response.
         self._login()
         response = self.client.get(self.view)
-        team_expect = Team.objects.get(pk=self.team.pk)
-        serializer = TeamSerializer(team_expect)
+        team_expected = Team.objects.get(pk=self.team.pk)
+        serializer = TeamSerializer(team_expected)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.json(), serializer.data)
 
--- a/hackathon_site/event/test_api.py
+++ b/hackathon_site/event/test_api.py
@@ -5,7 +5,7 @@ from rest_framework.test import APITestCase
 from hackathon_site.tests import SetupUserMixin
 
 from event.models import Profile, User, Team
-from event.serializers import TeamSerializer, UserSerializer
+from event.serializers import TeamSerializer, UserSerializer, ProfileSerializer, GroupSerializer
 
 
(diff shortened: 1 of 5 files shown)


hackathon_site/event/tests.py
```diff
@@ -31,3 +37,72 @@ def test_index_view(self):
...
+        profile = Profile.objects.create(user=self.user, team=team)
+        profile_serialized = ProfileSerializer(profile).data
+        profile_expected = {
+            "id": profile.id,
+            "status": profile.status,
+            "id_provided": profile.id_provided,
+            "attended": profile.attended,
+            "acknowledge_rules": profile.acknowledge_rules,
+            "e_signature": profile.e_signature,
+            "team": {
```
grahamhoyes: If we ever change the team serializer, this test fill fail as a result. The best thing to do would be to generate the expected `team` field here from the Team Serializer, which we will know is correct from it's own tests

hackathon_site/event/test_api.py
```diff
@@ -37,26 +38,69 @@ def test_user_has_no_profile(self):
...
+
+
+    
+
+
+
+
+
+class CurrentTeamTestCase(SetupUserMixin, APITestCase):
+    # When not logged in, the response is 401 unauthorized
```
grahamhoyes: You can remove these comments that were from the ticket description

hackathon_site/event/test_api.py
```diff
@@ -37,26 +38,69 @@ def test_user_has_no_profile(self):
...
+
+        group_expect = Group.objects.get(pk=self.group.pk)
+        group_serializer = GroupSerializer(group_expect)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(res_json['profile'], profile_serializer.data)
+        self.assertEqual(res_json['groups'], [group_serializer.data]) # need to figure out why its different
+
+
+
```
grahamhoyes: Don't forget to run Black for formatting, which will get rid of blank space like this

hackathon_site/event/tests.py
```diff
@@ -31,3 +37,72 @@ def test_index_view(self):
...
+            "email": user.email,
+            "profile": None,
+            "groups": [],
+        }
+        self.assertEqual(user_expected, user_serialized)
+
+class GroupSerializerTestCase(TestCase):
+    def test_serializer(self):
+
+        group = Group.objects.create()
```
grahamhoyes: You should give the group a name

hackathon_site/event/tests.py
```diff
@@ -31,3 +37,72 @@ def test_index_view(self):
...
+            "created_at": team.created_at.astimezone(tz).isoformat(),
+            "updated_at": team.updated_at.astimezone(tz).isoformat(),
+        }
+        self.assertEqual(team_expected, team_serialized)
+
+class UserSerializerTestCase(TestCase):
+    def test_serializer(self):
+
+        user = User.objects.create()
+        user_serialized = UserSerializer(user).data
```
grahamhoyes, line 69: We want to test the full serializer, which includes when it has a profile, team, and user. So, you should also create each of those and link them as necessary.

hackathon_site/event/tests.py
```diff
@@ -31,3 +37,72 @@ def test_index_view(self):
...
+        user_expected = {
+            "id": user.id,
+            "first_name": user.first_name,
+            "last_name": user.last_name,
+            "email": user.email,
+            "profile": None,
```
grahamhoyes: Once the profile has been created, we can directly put the profile serializer here. This way we are testing that the user-specific fields are tested, but we can just say "and the profile part of the response should be made with the `ProfileSerializer`. `ProfileSerializer` itself is tested below, so that gives us full coverage!

That would look something like you putting `ProfileSerializer(user.profile).data` here, or something along those lines

hackathon_site/event/tests.py
```diff
@@ -31,3 +37,72 @@ def test_index_view(self):
...
+            "groups": [],
```
grahamhoyes: As with above, we want to test that the groups are actually given in a list that uses `GroupSerializer`, which itself is tested independently so we know everything is correct.

hackathon_site/event/tests.py
```diff
@@ -31,3 +37,72 @@ def test_index_view(self):
...
+            "attended": profile.attended,
+            "acknowledge_rules": profile.acknowledge_rules,
+            "e_signature": profile.e_signature,
+            "team": {
+                "id": team.id,
+                "team_code": team.team_code,
+            },
+        }
+        # currently failing test
+        # self.assertEqual(profile_expected, profile_serialized)
```
grahamhoyes: It's failing because your team serializer now has `created_at` and `updated_at`, whereas thet code above doesn't have those fields. As with the other comments, using `TeamSerializer` inline will get around breaking tests like this when we change a nested serializer.

hackathon_site/event/test_api.py
```diff
@@ -37,26 +38,69 @@ def test_user_has_no_profile(self):
...
         self._login()
         response = self.client.get(self.view)
+        user_expect = User.objects.get(pk=self.user.pk)
+        serializer = UserSerializer(user_expect)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        data = response.json()
-        self.assertEqual(data, expected_response)
+        self.assertEqual(response.json(), serializer.data)
+    
+    def test_user_uses_profile_and_group_serializer(self):
```
grahamhoyes: You don't need this test. We know that the `UserSerializer` includes a group and profile (or we should from it's test), so the above test is sufficient.

hackathon_site/event/test_api.py
```diff
@@ -37,26 +38,69 @@ def test_user_has_no_profile(self):
...
-                    "attended",
-                    "acknowledge_rules",
-                    "e_signature",
-                )
-            },
-            "groups": [{"id": self.group.id, "name": self.group.name}],
-        }
         self._login()
         response = self.client.get(self.view)
+        user_expect = User.objects.get(pk=self.user.pk)
```
grahamhoyes: Minor, `user_expected` to be consistent with using past tense elsewhere

grahamhoyes:
Closed by #280

PR #137, opened 2020-09-03
Acknowledgement form

BrennyChao:
I'm not too sure if I'm doing this correctly, but I tried to avoid using the form control for the checkbox


hackathon_site/dashboard/frontend/src/components/user/AcknowledgementForm/AcknowledgementForm.js
```diff
@@ -0,0 +1,180 @@
...
+                    />
+                    {props.value}
+                </label>
+            )}
+        </Field>
+    );
+};
+
+export const AcknowledgementForm = ({ handleContinue, openDoc }) => {
+    const checkBox1 =
```
grahamhoyes, line 45: Since all of these are just used inline, you can put the text directly in line (with some line breaks to make it easier to read, they shouldn't be rendered in the HTML).

hackathon_site/dashboard/frontend/src/components/user/AcknowledgementForm/AcknowledgementForm.test.js
```diff
@@ -0,0 +1,74 @@
...
+                requestFailure={requestFailure}
+                openDoc={openDoc}
+            />
+        );
+
+        expect(getByText(/Continue/i)).toBeInTheDocument();
+        expect(getByText(/e-signature/i)).toBeInTheDocument();
+        expect(
+            getByText(
+                /I understand that making a request does not guarantee hardware. Hardware is given on a first-come-first-serve basis./i
```
grahamhoyes, line 25: While this is certainly helpful (and from a completeness perspective totally okay), I wouldn't worry about making sure the full text is in there, rather just make sure a part of it is there to make sure the element was rendered properly. That way if we reword the text, we don't need to copy-paste it down to the component.

There's nothing wrong with doing it this way from a code testing perspective, it's just not entirely necessary and may make changing things take extra steps :) 

If you really want to test for everything, this would be an acceptable use of a snapshopt test.

hackathon_site/dashboard/frontend/src/slices/users/userSlice.js
```diff
@@ -92,11 +100,52 @@ export const logIn = createAsyncThunk(
...
+//     }
+// );
+
 // Slice
 const userSlice = createSlice({
     name: "user",
     initialState,
-    reducers: {},
+    reducers: {
+        // toggleCheckbox1: (state) => {
```
grahamhoyes, line 136: I assume you've caught that this is unnecessary because formik handles the form state, all we need a reducer/thunk for is to submit the post request (which we don't even have an API endpoint for yet, so you can skip that part if you want)

hackathon_site/dashboard/frontend/src/slices/users/userSlice.js
```diff
@@ -92,11 +100,52 @@ export const logIn = createAsyncThunk(
...
+// export const acknowledge = createAsyncThunk(
+//     `${userReducerName}/acknowledge`,
+//     async({signature}, {dispatch, rejectWithValue}) => {
+//         try {
+//             const response = await post("/api/acknowledge",{signature});
+//             dispatch(push("/"));
+//             return response.data;
+//         } catch (e) {
+//             if(e.response.status === 400) {
+//                 return rejectWithValue({ status: 400, message: "Invalid credentials" });
```
grahamhoyes, line 112: 400 doesn't necessarily mean invalid credentials, it means the user did something wrong. That could include submitting an empty e-signature. The API would tell you what you did wrong (once it's been coded).

hackathon_site/dashboard/frontend/src/pages/Acknowledgement/Acknowledgement.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import styles from "./Acknowledgement.module.scss";
+
+import Header from "components/general/Header/Header";
+import Typography from "@material-ui/core/Typography";
+import AcknowledgementForm from "../../components/user/AcknowledgementForm/AcknowledgementForm";
+import Container from "@material-ui/core/Container";
+import Paper from "@material-ui/core/Paper";
+
+const Reports = () => {
```
grahamhoyes, line 10: Why `Reports`?

hackathon_site/dashboard/frontend/src/components/user/AcknowledgementForm/AcknowledgementForm.js
```diff
@@ -0,0 +1,180 @@
...
+    const onSubmit = (formikValues) => {
+        handleSubmit({ signature: formikValues.signature });
+    };
+
+    return (
+        <Formik
+            initialValues={{ signature: "", agreement: [""] }}
+            onSubmit={onSubmit}
+            validateOnBlur={false}
+            validateOnChange={false}
```
grahamhoyes, line 157: You should add a `validationSchema` in here as well, which you build using `Yup`. Take a look at the one used by the login form: https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/dashboard/frontend/src/components/user/LoginForm/LoginForm.js#L21

For you, each check box would have a requirement like `Yup.boolean().required(errorMessage)`, and the e-signature box would be `Yup.string().required(errorMessage)`. That way they can't submit the form until all fields have been filled out (and you should have an acompanying test for that, of course).

hackathon_site/dashboard/frontend/src/components/user/AcknowledgementForm/AcknowledgementForm.js
```diff
@@ -0,0 +1,180 @@
...
+                className={styles.acknowledgeCheckbox}
+            />
+            <FormikCheckBox
+                name="agreement"
+                value={checkBox4}
+                className={styles.acknowledgeCheckbox}
+            />
+            <Typography variant="body1" className={styles.acknowledgeText}>
+                {firstHalf}
+                <Link href="#" onClick={openDoc}>
```
grahamhoyes, line 120: @lisa-sa-li can you confirm, but I think this can just be a plain old link to some document on our website? Or is this supposed to pop up a react component that contains the waiver (pdf on our site would probably be the easiest)

hackathon_site/dashboard/frontend/src/slices/users/userSlice.js
```diff
@@ -92,11 +100,52 @@ export const logIn = createAsyncThunk(
     }
 );
 
+// export const acknowledge = createAsyncThunk(
+//     `${userReducerName}/acknowledge`,
+//     async({signature}, {dispatch, rejectWithValue}) => {
+//         try {
+//             const response = await post("/api/acknowledge",{signature});
```
grahamhoyes, line 107: This will eventually correspond to our Profile model: https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/event/models.py#L27

Sending an API request will set two values in the model: `acknowledge_rules` and `e_signature`. The easiest way for us to do this is just to make a PATCH (i.e. change) endpoint for the profile model, which takes in the fields you want to change. So this would probably be something like:

```javascript
const response = await patch("/api/event/profile", {e_signature: signature, acknowledge_rules: true})
```

`patch()` currently doesn't exist yet either, so that would need to be written too as a separate PR.

grahamhoyes (commented):
This is a really great start, nice work! I haven't checked out the branch locally yet, my comments are just from looking at the code.

hackathon_site/dashboard/frontend/src/components/user/AcknowledgementForm/AcknowledgementForm.js
lisa-sa-li: Yup, just the `<a>` tag is enough

PR #140, opened 2020-09-05
Formats the cart page

lisa-sa-li:


hackathon_site/dashboard/frontend/src/pages/Cart/Cart.module.scss
```diff
@@ -0,0 +1,9 @@
+.cart {
+    width: 100%;
+    max-width: 800px;
```
grahamhoyes, line 3: On my screen (regular 1080p 24" at the moment), this looks unusually narrow. 1280px is the lg breakpoint in mui, I think that would be more suitable. If that makes the components look too wide, some space between them wouldn't hurt.

I leave it up to your artistic expertise.

grahamhoyes (approved):

PR #141, opened 2020-09-06
User Registration

### LINKED ISSUE #117: Two-step user registration system
Like for previous events we've ran, we want a two-step user registration system:
1. The user signs up and has an account created, in a non-active state that doesn't let them log in. They are sent an activation email. 
2. The user follows a link in the activation email to activate their account. They can then login and use the site.

This seems to be a django app that accomplishes what we want: https://github.com/ubernostrum/django-registration. Some overriding of the views will be required to use our jinja templates instead of the provided templates.

grahamhoyes:
Resolves #117 

Adds a two-step user registration setup. The flow for QA:
- Go to /registration/signup/, and sign up.
- Watch the django console. In debug mode, it will print out the emails it's told to send rather than actually sending them.
- Go to the activation URL provided in the email
- You should be redirected to the dashboard, which itself should redirect you to the login page. After logging in, you should be on the dashboard.

Email settings still need to be specified in production. We also need to confirm that the email templates will build URLs properly when behind the reverse proxy.


hackathon_site/hackathon_site/settings/__init__.py
```diff
@@ -172,7 +174,16 @@
 STATIC_ROOT = os.path.join(BASE_DIR, "static")
 
 # Event specific settings
-REGISTRATION_OPEN_DATE = datetime(2020, 9, 1, tzinfo=pytz.timezone(TIME_ZONE))
-REGISTRATION_CLOSE_DATE = datetime(2020, 9, 30, tzinfo=pytz.timezone(TIME_ZONE))
-EVENT_START_DATE = datetime(2020, 10, 10, 10, 0, 0, tzinfo=pytz.timezone(TIME_ZONE))
-EVENT_END_DATE = datetime(2020, 10, 11, 17, 0, 0, tzinfo=pytz.timezone(TIME_ZONE))
+HACKATHON_NAME = "CoolHacks"
```
lisa-sa-li, line 177: SwagHacks would be a cooler hackathon

lisa-sa-li: SwiggitySwaggityHacks

hackathon_site/registration/jinja2/registration/signup.html
```diff
@@ -0,0 +1,34 @@
...
+    {% for field in form %}
+        {{ render_field(field) }}
+    {% endfor %}
+
+    {% if form.non_field_errors %}
+        {% for error in form.non_field_errors() %}
+            {{ error }}
+        {% endfor %}
+    {% endif %}
+{% endblock %}
```
lisa-sa-li: Is it possible to add a helper text below a field? I'd like "Must contain at least 8 characters" below the first password input

lisa-sa-li: Must contain letters blah blah

hackathon_site/registration/jinja2/registration/pending_activation.html
```diff
@@ -0,0 +1,14 @@
+{% extends "event/base.html" %}
+
+{% block body %}
+<div class="container">
+    <div class="section">
+        <div class="loginDiv z-depth-3 row">
+            <div class="row col s12 flexColCenter">
+                <h1 class="formH1">Activate your account</h1>
+                <p>In order to continue with your application, we first need to verify your email. Please check your email (maybe in the junk folder) for a welcome message from <a href="mailto:{{ from_email }}">{{ from_email }}</a>, with more instructions to activate your account. Once you've completed those steps, return to the site to continue with your application!</p>
```
lisa-sa-li: Just shortening it. It was way too long and I wouldn't have read it.

You must **verify your email** before you log in. Look for an email (maybe in the junk folder) from webmaster@localhost to activate your account in order to continue with your application!

hackathon_site/registration/jinja2/registration/test_emails/test_activation_email_subject.txt
```diff
@@ -0,0 +1,3 @@
+Hello there,
+
+This template is used as part of the unit tests. It can be ignored.
```
lisa-sa-li, line 3: Should the subject be 3 lines?

lisa-sa-li: Not that it matters tho since you join it into a single line later

lisa-sa-li (changes requested):
After logging in, it should redirect me to the non-existent page http://127.0.0.1:8000/accounts/profile/ right?

hackathon_site/registration/jinja2/registration/test_emails/test_activation_email_subject.txt
grahamhoyes: That's exactly why this one is 3 lines, so that the test can make sure it becomes 1

grahamhoyes:
> After logging in, it should redirect me to the non-existent page http://127.0.0.1:8000/accounts/profile/ right?

Yes, that's defined by the `LOGIN_REDIRECT_URL` setting: https://docs.djangoproject.com/en/3.1/ref/settings/#login-redirect-url. That's outside the scope of this PR though, which doesn't deal with login behaviour

lisa-sa-li (approved):

PR #144, opened 2020-09-07
Configure for User Uploaded Content

### LINKED ISSUE #58: Configure file uploads for Hardware images in debug mode
The admin site needs the ability to upload images for the `Hardware` model.

Locally, this involves setting a sensible location to serve user-uploaded content from. In production, this depends on how we configure the deployment and static file serving, so is outside of the scope of this issue. Throw an error in non-debug mode to remind us to set this up.

Some documentation links:
- https://docs.djangoproject.com/en/3.0/topics/http/file-uploads/
- https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-MEDIA_ROOT
- https://docs.djangoproject.com/en/3.0/howto/static-files/#serving-uploaded-files-in-development
- https://docs.djangoproject.com/en/3.0/howto/static-files/deployment/
- https://docs.djangoproject.com/en/3.0/topics/security/#user-uploaded-content-security

grahamhoyes:
Resolves #58.

Configures Django to accept user-uploaded content.

In debug mode, user-uploaded content will go in the `media` folder in the Django project root, and the url config will serve it under /media. You can test this by going to the admin page, and creating a new Hardware object and uploading an image. You should be able to edit the object afterwards, click on the image, and have it appear.

The media folder will need to be manually created on the server.


hackathon_site/hackathon_site/urls.py
```diff
@@ -45,5 +45,14 @@
 
 if settings.DEBUG:
     import debug_toolbar
+    from django.urls import re_path
+    from django.views.static import serve
 
     urlpatterns = [path("__debug__/", include(debug_toolbar.urls))] + urlpatterns
```
albogdan: Here you use `x = [] + x` but two lines down you use `x+= []` can you please pick one

hackathon_site/hackathon_site/settings/__init__.py
```diff
@@ -30,6 +30,9 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = bool(int(os.environ.get("DEBUG", 0)))
 
+# Overridden in hackathon_site.settings.ci
+IN_TESTING = False
```
albogdan: Isn't this the equivalent of `not DEBUG`

albogdan: Scratch that I see you removed it

hackathon_site/hackathon_site/urls.py
```diff
@@ -45,14 +45,21 @@
 
 if settings.DEBUG:
     import debug_toolbar
+    from django.core.exceptions import ImproperlyConfigured
     from django.urls import re_path
     from django.views.static import serve
 
     urlpatterns = [path("__debug__/", include(debug_toolbar.urls))] + urlpatterns
 
+    if settings.MEDIA_URL.startswith("http"):
```
albogdan, line 54: I'm confused why it matters if the `MEDIA_URL` starts with http

hackathon_site/hackathon_site/settings/__init__.py
grahamhoyes: It's not exactly the same as `not DEBUG`, the idea being you can have things that only happen in testing mode but not in production where `DEBUG` is `0` in both cases (I was trying to use it to change the media url). However they way I wanted to use it in the settings file didn't work so I removed it, but there may be uses to add it back in the future.

hackathon_site/hackathon_site/urls.py
grahamhoyes: You can set `MEDIA_URL` to something off-site if you serve your user-uploaded files on a different domain (eg because they go to an s3 bucket). In that case you won't be able to serve them in debug mode which is what the snippet here does, so I raise an `ImproperlyConfigured` to warn the user about that so they remove that part of the URL conf

hackathon_site/hardware/models.py
```diff
@@ -27,7 +27,7 @@ class Meta:
     quantity_available = models.IntegerField(null=False)
     notes = models.TextField(null=True)
     max_per_team = models.IntegerField(null=True)
-    picture = models.FileField(upload_to="hardware/pictures/", null=False)
+    picture = models.ImageField(upload_to="uploads/hardware/pictures/", null=False)
```
albogdan, line 30: Note that this will upload to media/uploads/hardware/pictures. I also don't know why we would be calling it uploads since we're not uploading, did you mean to call the resumes one uploads?

hackathon_site/requirements.txt
```diff
@@ -22,6 +22,7 @@ Jinja2==2.11.2
 MarkupSafe==1.1.1
 packaging==20.4
 pathspec==0.8.0
+Pillow==7.2.0
```
albogdan, line 25: What's this for?

hackathon_site/hardware/models.py
grahamhoyes: This is because of what I talked about in the readme and offline, we need a way to distinguish which uploaded files can be served by nginx and which can't (because they require some permissions). In the readme I said that anything under `/media/uploads/` can be served by nginx as public content, anything elsewhere under `/media/` (like `/media/resumes/`) needs to be served custom by a django view (that checks permissions, for example).

> since we're not uploading,

I assumed this actually was going to be done by an upload through the admin site. It doesn't make sense to have it as static files since then we can't add them easily, so even if we have a migration that creates them initially having them under `/media/uploads/` means we can use nginx to serve them per above.

hackathon_site/requirements.txt
grahamhoyes: It's used by `ImageField` to figure out if the file is an image or not. It's not part of Django's dependencies I assume since most people don't need it.

albogdan (approved):

PR #146, opened 2020-09-07
Add User Logout Ability and Fix Navbar

### LINKED ISSUE #57: Add default logout view / redirect
Add a page for handling logouts, that serves up https://docs.djangoproject.com/en/3.0/topics/auth/default/#django.contrib.auth.views.LogoutView. 

We can either make a "thanks for logging out" type page, or redirect them somewhere with https://docs.djangoproject.com/en/3.0/ref/settings/#logout-redirect-url. 

`drf-yasg` also menions a `django.conf.settings.LOGOUT_URL` setting, which it uses for it's logout button: https://drf-yasg.readthedocs.io/en/stable/settings.html#swagger-settings. Django's default settings docs don't seem to reference this, but we can set it anyway once it's available.

albogdan:
Resolves #57 

* Adds ability for user to logout of their session + unit tests
* Modifies base template to include navbar component, templates that inherit from it will all have the navbar except the ones that explicitly define the `nav` block as blank. 
    * Note that the navbar now by default includes the logo and a Login/Logout button
    * More buttons are added to it using the `nav_links` block
* Adds test to the landing page for rendering of the Login/Logout button depending on the session
* Updates the landing page with the new navbar

### DIFF
--- a/hackathon_site/event/jinja2/event/base.html
+++ b/hackathon_site/event/jinja2/event/base.html
@@ -18,13 +18,38 @@
         {% endblock %}
     </head>
     <body>
+        {% block nav %}
         <nav class="navbar primaryColor" role="navigation">
             <div class="nav-wrapper container">
                 <img src="{{ static('event/images/logo.svg') }}" class="logoNav" alt="logo" />
-                {% block nav %}
-                {% endblock %}
+
+                <ul class="right hide-on-med-and-down">
+                    {% block nav_links %}{% endblock %}
+
+                    {% block nav_logout_login_links %}
+                        {% if request.user.is_authenticated %}
+                            <li>
+                                <a href={{ url("event:logout") }} class="btn-small waves-effect waves-light colorBtn">Logout</a>
+                            </li>
+                        {% else %}
+                            <li>
+                                <a href={{ url("event:login") }} class="btn-small waves-effect waves-light colorBtn">Login</a>
+                            </li>
+                        {% endif %}
+                    {% endblock %}
+                </ul>
+
+                <ul id="nav-mobile" class="sidenav">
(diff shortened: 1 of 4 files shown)


hackathon_site/event/jinja2/event/base.html
```diff
@@ -18,13 +18,38 @@
         {% endblock %}
     </head>
     <body>
+        {% block nav %}
```
grahamhoyes: I would move this down to right below the image on line 24. There are some pages in figma that only have that and no nav content including login/logout buttons except for the logo, so having the block start below that would be an easy way to hide everything

hackathon_site/event/tests.py
```diff
@@ -96,3 +112,21 @@ def test_renders_when_logged_in(self):
         response = self.client.get(self.view)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertContains(response, "Dashboard")
+
+
+class LogOutViewTestCase(SetupUserMixin, TestCase):
+    """
+    Tests for the logout template. We simply redirect to the homepage.
+
+    This view uses django.contrib.auth.forms.AuthenticationForm, so no direct
```
grahamhoyes: It doesn't use `AuthenticationForm`, or any forms. You can say it uses `django.contrib.auth.views.LogoutView`, so no view logic needs testing

hackathon_site/event/urls.py
```diff
@@ -12,4 +12,7 @@
         name="login",
     ),
     path("dashboard/", DashboardView.as_view(), name="dashboard"),
+    path(
```
grahamhoyes, line 14: Move this to under the login path just to keep them together

grahamhoyes (commented):
Functionality looks good, just a few minor recommended tweaks

grahamhoyes (approved):

PR #148, opened 2020-09-08
143 - Password Reset Functionality

### LINKED ISSUE #143: Add password reset flow
- Implement templates for [`PasswordResetView`](https://docs.djangoproject.com/en/3.1/topics/auth/default/#django.contrib.auth.views.PasswordResetView), [`PasswordResetDoneView`](https://docs.djangoproject.com/en/3.1/topics/auth/default/#django.contrib.auth.views.PasswordResetDoneView), [`PasswordResetConfirmView`](https://docs.djangoproject.com/en/3.1/topics/auth/default/#django.contrib.auth.views.PasswordResetConfirmView) and [`PasswordResetCompleteView`](https://docs.djangoproject.com/en/3.1/topics/auth/default/#django.contrib.auth.views.PasswordResetCompleteView)
- Add those to the url conf, likely in `event.urls` under `/accounts/` with suitable names and URL routes. Pass in necessary attributes to `.as_view()`.
- Tests, similar to other account views.

albogdan:
Resolves #143 

Implements functionality to reset a user's password using 4 builtin Django views.

## Testing
* To test start the server and navigate to the login page then click on `Reset Password`
* This should take you to a page where you enter your email. Once you do that click submit and check the Django console for the email and a link to the next page
* Click the link to the email and that should take you to the password reset page where you enter a new password.
* Once you entered a new password and pressed submit you should get to a `Password Reset Successful` page.

## Features Implemented
* Connects existing login template to password reset page
* Creates 4 templates for the 4 stages in the password reset process
* Creates a password reset email HTML template
* Creates a subject text file for the subject of the password reset email
* Adds hackathon name to the jinja template
* Adds unit tests that test the password reset process end to end in the `PasswordResetTestCase` class
   * First we test if the `password_reset` page renders
   * Then we test if submitting an email redirects the user to the `password_reset_done` page
   * We then test if when the user clicks on the link in their email it takes them to the `password_reset_confirm` page which prompts them to enter a new password
   * Finally we test that entering a new password works and redirects the user to the `password_reset_complete` page

### DIFF
--- a/hackathon_site/event/jinja2/event/form_base.html
+++ b/hackathon_site/event/jinja2/event/form_base.html
@@ -24,10 +24,13 @@
         <div class="loginDiv z-depth-3 row">
             <div class="row col s12 flexColCenter {% if self.form_img() %}l7 m8{% endif %}">
                 <h1 class="formH1">{% block form_title %}{% endblock %}</h1>
+                <p>{% block form_description %}{% endblock %}</p>
+                {% block form_entry %}
                 <form class="col s12"
                     action="{{ form_action }}"
                     method="{{ form_method if form_method else "post" }}"
                 >
+                {% endblock %}
                     {{ csrf_input }}
                     {% block form_inputs %}{% endblock %}
                     <br />
--- a/hackathon_site/event/jinja2/event/form_base.html
+++ b/hackathon_site/event/jinja2/event/form_base.html
@@ -24,13 +24,11 @@
         <div class="loginDiv z-depth-3 row">
             <div class="row col s12 flexColCenter {% if self.form_img() %}l7 m8{% endif %}">
                 <h1 class="formH1">{% block form_title %}{% endblock %}</h1>
-                <p>{% block form_description %}{% endblock %}</p>
-                {% block form_entry %}
+                {% block form_description %}{% endblock %}
                 <form class="col s12"
-                    action="{{ form_action }}"
+                    {% if form_action %} action="{{ form_action }}" {% endif %}
                     method="{{ form_method if form_method else "post" }}"
                 >
-                {% endblock %}
                     {{ csrf_input }}
                     {% block form_inputs %}{% endblock %}
                     <br />
--- a/hackathon_site/event/jinja2/event/reset_password/reset_password_confirm.html
+++ b/hackathon_site/event/jinja2/event/reset_password/reset_password_confirm.html
@@ -1,0 +1,33 @@
+{% extends "event/form_base.html" %}
+
+{% block form_title %}{{ "Invalid Link" if form is none else "Reset Password" }}{% endblock %}
+
+{% block form_entry %}
+    <form class="col s12" method="{{ form_method if form_method else "post" }}">
+{% endblock %}
+
+{% block form_inputs %}
+    {{ render_field(form.new_password1, show_help_text=True) }}
+    {{ render_field(form.new_password2) }}
+
+    {% if form.non_field_errors %}
+        {% for error in form.non_field_errors() %}
+            {{ error }}
+        {% endfor %}
+    {% endif %}
+{% endblock %}
+
+{% block form_button %}
+<button type="submit" class="btn-small waves-effect waves-light secondaryColor colorBtn center">Reset Password</button>
+{% endblock %}
+
+{% block form_bottom %}
+<div class="col s12 loginLink">
+    <i class="material-icons primaryText">exit_to_app</i>
+    <a href="{{ url("event:dashboard") }}" class="primaryText hoverLink s6">Back to Dashboard</a>
+</div>
+<br />
+<div class="row center">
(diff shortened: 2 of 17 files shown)


hackathon_site/event/jinja2/event/reset_password/reset_password_email.html
```diff
@@ -0,0 +1,17 @@
+<p>Hello {{ user.get_username() }},</p>
```
grahamhoyes: Their username will be their email. Would it be better to use `user.firstname`?

hackathon_site/event/jinja2/event/reset_password/reset_password_email.html
```diff
@@ -0,0 +1,17 @@
...
+
+<p>You're receiving this email because you requested a password reset for your {{ event_name }} user account.</p>
+
+<p>Please go to the following page and choose a new password:</p>
+
+<a href="{{ protocol }}://{{ domain }}{{ url('event:reset_password_confirm', args=[uid,token])}}">Reset Password</a>
```
grahamhoyes, line 7: Created an issue in the newhacks repo to remind us to test this: https://github.com/ieeeuoft/newhacks-site-2020/issues/22

hackathon_site/event/jinja2/event/reset_password/reset_password_email.html
```diff
@@ -0,0 +1,17 @@
...
+
+<p>If that link doesn't work, copy and paste this link into your browser: </p>
+<p>{{ protocol }}://{{ domain }}{{ url('event:reset_password_confirm', args=[uid,token])}}</p>
+
+<p>All the best,</p>
+
+<p>The {{ site_name }} team.</p>
```
grahamhoyes: Should be `{{ event_name }}` (and maybe don't have a fullstop at the end)

hackathon_site/event/tests.py
```diff
@@ -162,3 +164,79 @@ def test_valid_submit_redirect(self):
...
+    django.contrib.auth.views.PasswordResetDoneView,
+    django.contrib.auth.views.PasswordResetConfirmView, and
+    django.contrib.auth.views.PasswordResetCompleteView, so no logic testing
+    is required.
+    """
+
+    def setUp(self):
+        super().setUp()
+        self.view_request_reset = reverse("event:reset_password")
+        # self.view_reset = reverse("event:reset_password_confirm")
```
grahamhoyes: can remove?

hackathon_site/event/jinja2/event/reset_password/reset_password_confirm.html
```diff
@@ -0,0 +1,33 @@
+{% extends "event/form_base.html" %}
+
+{% block form_title %}{{ "Invalid Link" if form is none else "Reset Password" }}{% endblock %}
+
+{% block form_entry %}
```
grahamhoyes: This was already possible without this block using the `form_action` and `form_method` variables, imo stick with that for consistency instead of having a second way to do it

hackathon_site/event/jinja2/event/form_base.html
```diff
@@ -24,10 +24,13 @@
         <div class="loginDiv z-depth-3 row">
             <div class="row col s12 flexColCenter {% if self.form_img() %}l7 m8{% endif %}">
                 <h1 class="formH1">{% block form_title %}{% endblock %}</h1>
+                <p>{% block form_description %}{% endblock %}</p>
+                {% block form_entry %}
```
grahamhoyes: See other comment, I'm not sure this is necessary

hackathon_site/event/jinja2/event/form_base.html
```diff
@@ -24,10 +24,13 @@
         <div class="loginDiv z-depth-3 row">
             <div class="row col s12 flexColCenter {% if self.form_img() %}l7 m8{% endif %}">
                 <h1 class="formH1">{% block form_title %}{% endblock %}</h1>
+                <p>{% block form_description %}{% endblock %}</p>
```
grahamhoyes: Could we move the `<p>` tags inside the block when it's defined? It would be nice to be able to put a h2 or something in here too, or apply different classes (I could use this in my PR)

grahamhoyes (commented):
A few comments attached.

There's still the issue of getting an error 500 when either the token or user account in the URL are invalid. Digging into things, that's because the page get's `form: None` in it's context, so doing `form.new_password1` crashes it.

Where django bails from the password reset confirm view: https://github.com/django/django/blob/master/django/contrib/auth/views.py#L284

Where extra things get added to the context on failure: https://github.com/django/django/blob/master/django/contrib/auth/views.py#L312. So in the template, we'll have `validLink == False` or  `form is none` to check for a bad page, and display an appropriate error.

grahamhoyes (approved):

PR #149, opened 2020-09-10
Application Form

### LINKED ISSUE #145: Add application form and template
- The form should be a `ModelForm`, based off the Application model.
- Add validation for relevant fields
- The template doesn't need a URL route in the end, because it will be served by the dashboard. One can be added optionally now for testing.
- Use [Select2](https://select2.org/) for the school selection, pulling from MLH's list of schools: https://github.com/MLH/mlh-policies/blob/master/schools.csv
    - Allow [tagging](https://select2.org/tagging) for entering custom options if their school isn't found

grahamhoyes:
Resolves #145.

- Adds an application form and template.
    - Almost half of this PR is scss to make select2 work on the application template, adapted from https://github.com/kevzlou7979/select2-materialize
- Adds associated validators for some fields.
    - The `UploadedFileValidator` is tested implicitly through the application form tests, since testing it in isolation would require mocking `FileField`, or making fake database models.
- Adds admin pages for applications and registration teams. You can only submit an application once per user, so you'll probably want to use the admin site to delete applications during testing.

The page is currently not hooked up to the dashboard at all, that will be a separate PR. The form is at `/registration/application/`.


hackathon_site/registration/models.py
```diff
@@ -25,23 +25,65 @@ def __str__(self):
 
 
 class Application(models.Model):
+    GENDER_CHOICES = [
+        (None, ""),
+        ("male", "Male"),
+        ("female", "Female"),
+        ("non-binary", "Non-binary"),
+        ("other", "Other"),
```
albogdan, line 36: Are we giving people the opportunity of inputting their own gender when they select other or just leaving it as Other?

hackathon_site/registration/models.py
```diff
@@ -25,23 +25,65 @@ def __str__(self):
...
     phone_number = models.CharField(max_length=20, null=False)
+    school = models.CharField(max_length=255, null=False)
+    study_level = models.CharField(
+        max_length=50, choices=STUDY_LEVEL_CHOICES, null=False
+    )
+    graduation_year = models.IntegerField(null=False)
     resume = models.FileField(upload_to="applications/resumes/", null=False)
     q1 = models.TextField(null=False)
     q2 = models.TextField(null=False)
     q3 = models.TextField(null=False)
```
albogdan: Should we add a max length for each response

hackathon_site/registration/forms.py
```diff
@@ -109,7 +109,23 @@ class Meta:
         }
 
     def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user")
```
albogdan, line 126: What exactly does this do

hackathon_site/registration/views.py
```diff
@@ -93,6 +93,11 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class ApplicationView(LoginRequiredMixin, FormView):
+class ApplicationView(LoginRequiredMixin, ModelFormMixin, FormView):
     form_class = ApplicationForm
     template_name = "registration/application.html"
+
+    def get_form_kwargs(self):
```
albogdan, line 102: Also what does this do

hackathon_site/registration/jinja2/registration/application.html
```diff
@@ -97,7 +97,7 @@
 
     select2.data('select2').$selection.css('height', '2.4375rem');
 
-    {% if form.school.value %}
+    {% if form.school.value() and form.school.value() is not none %}
```
albogdan: Isn't this redundant cause if form.school.value() is something it won't be none

If anything you'd wanna be testing for the not none first then the form.school.value()

hackathon_site/registration/admin.py
```diff
@@ -1,6 +1,32 @@
...
+
+
+@admin.register(TeamApplied)
+class TeamAppliedAdmin(admin.ModelAdmin):
+    search_fields = ("id", "team_code")
+    inlines = (ApplicationInline,)
+
+
+@admin.register(Application)
+class ApplicationAdmin(admin.ModelAdmin):
```
albogdan, line 18: What's the purpose of these classes as opposed to just doing it with `admin.site.register`

hackathon_site/registration/models.py
```diff
@@ -25,23 +28,94 @@ def __str__(self):
...
+    birthday = models.DateField(null=False)
+    gender = models.CharField(max_length=50, choices=GENDER_CHOICES, null=False)
+    ethnicity = models.CharField(max_length=50, choices=ETHNICITY_CHOICES, null=False)
+    phone_number = models.CharField(
+        max_length=20,
+        null=False,
+        validators=[
+            validators.RegexValidator(
+                r"^(?:\+\d{1,2})?\s?\(?\d{3}\)?[\s-]?\d{3}[\s-]?\d{4}$",
+                message="Enter a valid phone number.",
```
albogdan, line 74: We should display on the form an example of a valid phone number so they know how we expect it

hackathon_site/event/models.py
```diff
@@ -28,7 +28,9 @@ class Profile(models.Model):
     STATUS_CHOICES = [("Accepted", "Accepted"), ("Waitlisted", "Waitlisted")]
 
     user = models.OneToOneField(User, on_delete=models.CASCADE)
-    team = models.ForeignKey(Team, on_delete=models.CASCADE, null=False)
+    team = models.ForeignKey(
```
albogdan, line 31: Shouldn't team also be 1:1

grahamhoyes: No - team to profile is one to many (many people on a team). `OneToOneField` is just a foreign key field with a unique constraint.

hackathon_site/registration/admin.py
grahamhoyes: They're used to customize how the model appears in the admin site: https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#modeladmin-objects.

Doing `admin.site.register(Application, ApplicationAdmin)` does the same thing as the `@admin.register` decorator.

hackathon_site/registration/models.py
grahamhoyes: Probably, but I was thinking of leaving that for instances of the template rather than enforcing it in the template repo. Thoughts?

hackathon_site/registration/models.py
grahamhoyes: I figured no, since it's not a statistic that we've used in the past and it was extra work to implement. If users of the template want, they can remove the choices and leave it as just a text field, and use something like select2 with [tagging](https://select2.org/tagging) to have options available to start (like we do with the school field).

hackathon_site/registration/forms.py
grahamhoyes: When the form is initalized, `user` is passed as a kwarg to it. This happens from the `get_form_kwargs` method of `ApplicationView` in registration/views.py. Since the actual base form class doesn't expect user as a kwargs, we pop it before initializing the superclass.

hackathon_site/registration/views.py
grahamhoyes: https://docs.djangoproject.com/en/3.1/ref/class-based-views/mixins-editing/#django.views.generic.edit.FormMixin.get_form_kwargs

hackathon_site/registration/jinja2/registration/application.html
grahamhoyes: From testing, it seems that in jinja that `None` comes out to be truthy. Without that extra bit, if you manage to submit the form with empty fields (which you have to remove all the `required` attributes to do), they will come back with the string "None" as the value.

hackathon_site/registration/models.py
albogdan: `If users of the template want, they can remove the choices and leave it as just a text field, and use something like select2 with tagging to have options available to start (like we do with the school field).`

Maybe we should add this to the README

hackathon_site/registration/jinja2/registration/application.html
albogdan: Got it, would still be best practice to test for none first

hackathon_site/event/models.py
albogdan: My bad I meant the team in registration/models.py

hackathon_site/registration/models.py
```diff
@@ -25,23 +28,94 @@ def __str__(self):
...
+    STUDY_LEVEL_CHOICES = [
+        (None, ""),
+        ("highschool", "High School"),
+        ("undergraduate", "Undergraduate"),
+        ("gradschool", "Graduate School"),
+        ("other", "Other"),
+    ]
+
+    user = models.OneToOneField(User, on_delete=models.CASCADE, null=False)
+    team = models.ForeignKey(
```
albogdan, line 60: Shouldn't this team be one to one is what I meant

hackathon_site/event/models.py
grahamhoyes: No, it's the same thing. Multiple users can end up on the same team for applications (everyone is given a deafult one, but they are free to leave it and join another).

hackathon_site/registration/models.py
grahamhoyes: No, see my reply above

hackathon_site/registration/models.py
grahamhoyes: Would be good to make a docs section with all of this: https://github.com/ieeeuoft/hackathon-template/issues/150

albogdan (approved):

PR #166, opened 2020-09-14
Add logout redirect url to settings

grahamhoyes:
This is not noticeable when running locally, because it defaults to `/`. However when running under a subfolder, logging out ends up booting you off the event site.


grahamhoyes:
Closing because this is not a correct solution (changes needed to the url routes as well), and it was resolved by #167

PR #186, opened 2020-09-19
Creating Review App and Models

### LINKED ISSUE #178: Create Review App and Models
- Create a new `review` app
- Add a new model for Reviews:
```
class Review(models.Model):
	reviewer = foreign key to user
	application = OneToOne to application
	interest = int from 0 to 10
	experience = int from 0 to 10
	quality = int from 0 to 10
	reviewer comments = TextField optional
	status = accepted or waitlisted or rejected
	decision_sent_date = DateField, nullable # Add a comment that changing status should set this to False
        created and updated dates
```
- Remove status field from event.Profile model
- Add nullable boolean `rsvp` field to registration.Application
- Add setting `RSVP_DAYS`

No tests required

albogdan:
### Overview
- Resolve #178 
- Adds a review app to the django site, with the model as seen in the issue
- Removes status field from `event.Profile` model
- Adds a nullable boolean `rsvp` field to the `registration.Application` model
- Adds a setting for `RSVP_DAYS` which is the number of days participants are allowed to validly RSVP after being accepted

### Testing
- None required since we're not using any validators. That being said we might want to validate that the `interest`, `experience`, and `quality` fields are between 0 and 10 however that should probably be done in the form before inserting into the database.


hackathon_site/event/models.py
```diff
@@ -25,13 +25,11 @@ def __str__(self):
...
 
 class Profile(models.Model):
-    STATUS_CHOICES = [("Accepted", "Accepted"), ("Waitlisted", "Waitlisted")]
-
     user = models.OneToOneField(User, on_delete=models.CASCADE)
     team = models.ForeignKey(
         Team, related_name="profiles", on_delete=models.CASCADE, null=False
     )
-    status = models.CharField(max_length=64, choices=STATUS_CHOICES, null=False)
+    status = models.CharField(max_length=64, null=False)
```
grahamhoyes: I believe we want to remove this entirely, not just remove the choices

hackathon_site/review/models.py
```diff
@@ -0,0 +1,26 @@
...
+        ("Accepted", "Accepted"),
+        ("Waitlisted", "Waitlisted"),
+        ("Rejected", "Rejected"),
+    ]
+
+    reviewer = models.ForeignKey(
+        User, related_name="reviews", on_delete=models.SET_NULL, null=True
+    )
+    application = models.OneToOneField(
+        Application, on_delete=models.SET_NULL, null=True
```
grahamhoyes: This is null=False, since you can't have a review without an application tied to it. I think when I said to make this nullable so we don't have to cascade, I was thinking the foreign key was from application to review, whereas it's actually the other way around.

Tldr; this can be `on_delete=models.CASCADE, null=False`

hackathon_site/review/models.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+    interest = models.IntegerField(default=False, null=False)
```
grahamhoyes: Add a `MinValueValidator` and `MaxValueValidator` to these 3 fields limiting options from 0 or 1 to 10 perhaps? Or should we leave it open

https://docs.djangoproject.com/en/3.1/ref/validators/#django.core.validators.MinValueValidator

grahamhoyes: Also `default` is the default value, not a boolean flag. Goes for the fields below as well.

albogdan: See overview at the top

grahamhoyes (approved):

PR #187, opened 2020-09-24
Review Admin Forms

grahamhoyes:
## Overview

- Adds a proxy model for `registration.Team`, so that it can be registered in the admin site again with different inlines. The migrations don't do anything to the database.
- Adds forms to the admin site for reviewing applications, by overriding the `ModelForm` that would normally be used in `ApplicationInline`.
- Adds a custom formset for handling the application form, just to pass in a request object.

## Unit Tests Created
Form tests:
- Initialization given an existing review, no initialization with no review
- Doesn't allow review changes after a decision has been sent
- Saving a review for the first time
- Overwriting an existing review (the reviewer will be changed)
- Saving based off the `commit` parameter

Admin list view tests:
- Is accessible or inaccessible depending on permissions
- Shows member count and whether the team has been fully reviewed or not
- Filters on team reviewed or not

Admin change view tests:
- Is accessible or inaccessible depending on permissions
- Successfully submits reviews
- Renders existing reviews

## Steps to QA
- Create one or more users, and submit applications for them.
- Have some of the users join a team
- Go to the "Teams" section of the Review app on the admin site.
- Teams should appear in the list view.
    - Number of members should be correct.
    - The team should only be marked as reviewed if all members have been reviewed.
- Open the change page for a particular team.
- Submit a review. Everything should be updated properly.
- Go to the review model in the admin site, and set a decision sent date. Try to submit a new review through the team page (change some fields). You should not be able to.


grahamhoyes:
Need to make it so that the review.Teams page is accessible to people with review add or change permissions. By default it's probably based on team permission.

grahamhoyes:
Needs validation so that status can't be changed after a decision has been sent

grahamhoyes:
Remove the ability to delete a team on the review page (even though users won't have delete permissions, so the button won't show up for non-superusers)

grahamhoyes:
In review-admin-forms, make sure that when status is set to accepted or rejected, decision_sent_date is set to None (so that previously waitlisted people can be sent a new email)

hackathon_site/review/admin.py
```diff
@@ -1,4 +1,199 @@
...
+    def get_resume_link(self, obj):
+        link = '<a href="{}" target="_blank">Open resume</a>'.format(obj.resume.url)
+        return mark_safe(link)
+
+    get_resume_link.short_description = "Resume"
+
+    def get_reviewer_name(self, obj):
+        if hasattr(obj, "review") and obj.review.reviewer:
+            return f"{obj.review.reviewer.first_name} {obj.review.reviewer.last_name}"
+
```
albogdan: nit: extra line

hackathon_site/review/admin.py
```diff
@@ -1,4 +1,199 @@
...
+        return ""
+
+    get_reviewer_name.short_description = "Reviewer name"
+
+    def get_decision_sent_date(self, obj):
+        if hasattr(obj, "review") and obj.review.decision_sent_date:
+            return obj.review.decision_sent_date
+
```
albogdan: nit: extra line

hackathon_site/review/admin.py
```diff
@@ -1,4 +1,199 @@
...
+    def has_change_permission(self, request, obj=None):
+        """
+        Let users with change_review permissions use this inline, rather than the default
+        change_application. Viewing still requires registration.view_application.
+        """
+        return request.user.has_perm("review.change_review")
+
+
+class TeamReviewedListFilter(admin.SimpleListFilter):
+    title = _("Team Reviewed")
```
albogdan, line 121: What's the point of using `gettext_lazy` for all of these?

hackathon_site/review/test_admin.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def test_list_page_permission_denied(self):
+        """
+        Without view permissions, the teams review page should not be accessible
+        """
+        self._login()
+        response = self.client.get(self.list_view)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_list_page_shows_team_member_count(self):
```
albogdan, line 67: Can you add a small test with a team that's not full since a full team is kind of a degenerate case

albogdan (changes requested):
Overall good stuff, a few minor nitpicks and adding an extra test function to be more general. QA was uneventful and everything behaved as expected.

hackathon_site/review/admin.py
grahamhoyes: Looks cleaner :eyes: (will remove)

hackathon_site/review/admin.py
grahamhoyes: In case anyone wants to support translation for this project in the future. Other than that, no point that I know of (the docs just do it so I do too)

hackathon_site/review/test_admin.py
grahamhoyes: How so? .count() behaves the same whether there's 1 or 4 team members, so there's no reason to expect any difference since 4 isn't hard coded anywhere for the admin pages

grahamhoyes: Changed this test to only have 3 team members to alleviate concerts about 4 being hard coded somewhere

grahamhoyes:
Amended QA steps for waitlisted logic:
- Create a review that has been accepted or rejected and set a decision_sent_date. No changes should be allowed through the review form.
- Change the status to waitlisted via the review model page. Changes should now be allowed, but the decision sent date not cleared.
- Change the status from waitlisted to accepted on the team review page. Change should be allowed, and the decision sent date cleared.

albogdan (dismissed):

albogdan (approved):

PR #188, opened 2020-09-25
RSVP View Backend Implementation

### LINKED ISSUE #180: Create RSVP View
- Create RSVP view in registration/views.py
- Add route that takes a `yes` or `no` kwarg to `registration/urls.py`: `re_path("rsvp/(?P<rsvp>yes|no)/$", views.RSVPView.as_view(), name="rsvp")`
- Look at `kwargs["rsvp"]` to figure out which option they picked
- Check if expired: `now() > review.decision_sent_date + settings.RSVP_DAYS` (need timedelta in there), redirect straight to dashboard.
- If yes, set rsvp=True on application and create a profile
- If false, set rsvp=False on application
- Redirect to dashboard on success

albogdan:
### Overview
- Resolves #180 
- Implements a view that takes in either a `rsvp=yes` or `rsvp=no` `HTTP GET` parameters
- In the case that the request goes through, the user either
   - Gets `application.rsvp=True` and a profile created, if they `rsvp=yes`
   - Gets `application.rsvp=False`, if they `rsvp=no`
- The user cannot RSVP if the date is later than `RSVP_DAYS` past the date that a decision was sent to them. Additionally, at any time in the `RSVP_DAYS` time period, the user can change their mind and RSVP no
- The user cannot RSVP if they:
   1. Don't have an application
   2. Haven't had their application reviewed
   3. Haven't been accepted

   These result in `400` responses.
- Note that if we put their application status on the dashboard, we can just redirect to the dashboard for case `3`
- Otherwise we redirect to the dashboard immediately.

### Testing
- Unit tests for `400` responses, yes RSVP, no RSVP, and RSVP deadline passed

### QA
1. Create a user and application, and a review for yourself in the admin site. Then login to that user.
2. Go to `/registration/rsvp/yes` to RSVP yes, `/registration/rsvp/no` to RSVP no
   1. After you RSVP yes, go to admin confirm that you have a profile
   2. After you RSVP yes, go to admin confirm that you DO NOT have a profile
3. Try giving yourself a different `status` other than `Accepted` and see what happens
4. Try going to `/registration/rsvp/X` where `X=anything other than yes or no` - this should `404`


hackathon_site/registration/views.py
```diff
@@ -18,7 +18,9 @@
 
 from hackathon_site.utils import is_registration_open
 from registration.forms import SignUpForm, ApplicationForm
-from registration.models import Team
+from registration.models import Team as RegistrationTeam
+from event.models import Team as EventTeam
+from event.models import Profile
```
grahamhoyes: Combine this into the same line as above

hackathon_site/registration/views.py
```diff
@@ -211,3 +213,71 @@ def get(self, request, *args, **kwargs):
             response["Content-Encoding"] = encoding
 
         return response
+
+
+class RSVPView(LoginRequiredMixin, View):
+    """
+        This page checks if the RSVP deadline has passed and then does one of
```
grahamhoyes: Minor, unindent this entire docstring

hackathon_site/registration/views.py
```diff
@@ -211,3 +213,71 @@ def get(self, request, *args, **kwargs):
...
+
+        Finally we redirect to the dashboard.
+    """
+
+    def get(self, request, *args, **kwargs):
+        decision = kwargs["rsvp"]
+
+        # Check for nulls
+        user = self.request.user
+        if not hasattr(request.user, "application"):
```
grahamhoyes: you set `user = self.request.user` above, but use `request.user` here. Both are identical, but would just use `user` for consistency since you have it

hackathon_site/registration/views.py
```diff
@@ -211,3 +213,71 @@ def get(self, request, *args, **kwargs):
...
+            return HttpResponseBadRequest(
+                "You have not submitted an application.".encode(encoding="utf-8")
```
grahamhoyes, line 237: This can take a regular string, which django will encode with `settings.DEFAULT_CHARSET` automatically.

For whatever reason I decided to do this above in `LeaveTeamView` too, so we can either keep it or remove the encoding in all `HttpResponseBadRequest`s in this file.

hackathon_site/hackathon_site/tests.py
```diff
@@ -69,6 +70,24 @@ def _make_full_registration_team(self, team=None):
 
         return team
 
+    def _review(
+        self,
+        status="Accepted",
+        decision_sent_date=datetime.now().replace(tzinfo=settings.TZ_INFO).date(),
```
grahamhoyes: It's best not to put function calls in function definitions. They only get called once on function creation (in this case, when the test runner first imports the SetupUserMixin), not for every test. This means that things like the `@override_settings` decorator and mocking of `datetime` for one test method will have no effect.

Normally, the way to get around that is to do something like:
```python
def foo(bar=None):
    if bar is None:
        bar = datetime.now()
    ...
```

However in this case we actually want the option for `decsion_sent_date` to be `None` (when no decision has been sent yet), so something like this would work:

```python
def _review(self, status="Accepted", **kwargs):
    decision_sent_date = kwargs.pop("decision_sent_date", datetime.now().replace(tzinfo=settings.TZ_INFO).date()
    ...
```

grahamhoyes (commented):
Tested all of the mentioned cases, and they look good. A few suggestions below.

RSVPing passed the deadline doesn't lead to a 400 page, I assume that will be mentioned on their dashboard when they are redirected there?

hackathon_site/registration/views.py
albogdan: I say keep it?

grahamhoyes: Works for me, though I would avoid it in new files so we can let django handle it.

grahamhoyes (approved):

PR #196, opened 2020-10-17
179 admin site plus

### LINKED ISSUE #179: Add AdminSitePlus
- Add https://github.com/jsocol/django-adminplus
- Just need to go as far as installing and adding it to the URL routes, don't need to register any views. No unit tests required.
- Doesn't exactly support Django 3, but it seems to work without crashing

1999akinori:
## Overview

- resolves #179 
- installed adminplus

You can check if adminplus is working by adding a placeholder custom view and seeing it in the admin page.

Place in event/admin.py
```
from django.http import HttpResponse

@admin.site.register_view("placeholder")
def placeholder_view(request):
    return HttpResponse("Hello")
```

### DIFF
--- a/hackathon_site/hackathon_site/settings/__init__.py
+++ b/hackathon_site/hackathon_site/settings/__init__.py
@@ -53,13 +53,14 @@ CORS_ALLOW_CREDENTIALS = True
 # Application definition
 
 INSTALLED_APPS = [
-    "django.contrib.admin",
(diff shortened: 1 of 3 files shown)


grahamhoyes (approved):
LGTM (the views look a bit broken in the admin list page, but that's expected and the fault of the package being out of date)

albogdan (approved):

PR #212, opened 2020-10-24
Send Decision Emails to Participants

### LINKED ISSUE #184: Create a page that sends emails for reviews
- Make this a custom page in the admin site with AdminSitePlus. Accessible to superusers only.
- Use the send mail form from #182 to get status and date range
- Filter users by reviews on the above and those who have review.decision_sent_date is None to get their email
- Send a django bulk email. Use render_to_string() and templates from #183.
- Make sure that emails don't time out if sending hundreds. If they do, use multiprocessing.
- Pretty much identical to: https://github.com/albogdan/makeuoft-site-2.0/blob/master/application/review/manager.py#L226
- Use a single mail connection to avoid the request timing out: https://docs.djangoproject.com/en/3.1/topics/email/#sending-multiple-emails
- Since emails still get sent synchronously and may time out, consider a threading approach: https://stackoverflow.com/questions/32979945/django-send-mail-function-taking-several-minutes/32980725
- As soon as the email gets sent to the user, set the decision sent date (in case it crashes)
- Limit by the `quantity` field on the form. When email sending finishes, render how many are left back to the template.

albogdan:
## Overview

- Resolves #184 
- Creates a template for sending emails using the django admin site
- Creates a link in the top bar of the admin site when in `Review --> Teams` that takes you to the page to send decisions. Decisions can only be sent by superusers
- Can see number of Accepted, Waitlisted, Rejected people left to send emails to

## Unit Tests Created

- Check permissions for only superusers allowed to access page
- Send emails to Accepted, Waitlisted, Rejected and check that the correct amount are sent
- Send emails to Accepted, Waitlisted, Rejected and check that the text is correct in the subject and body

## Steps to QA

- Create some users and reviews and send various emails to them using the system. Make sure you do this as a superuser. 
- You should be able to see the emails in your console. Make sure the text is correct based on the status of the users you are sending emails for
- Make sure the number of emails sent corresponds to the number of emails you entered to send on the form
- Try sending with a date constraint and make sure it only sends emails to those users that were reviewed within those days.

### DIFF
--- a/hackathon_site/review/jinja2/review/send-decisions.html
+++ b/hackathon_site/review/jinja2/review/send-decisions.html
@@ -2,7 +2,6 @@
 
 {% block head_extends %}
 <!-- Select2 CSS -->
-<link href="https://cdn.jsdelivr.net/npm/select2@4.1.0-beta.1/dist/css/select2.min.css" rel="stylesheet" />
 <link rel="stylesheet" type="text/css" href="{{ static('event/styles/css/select2.css') }}" />
 
 <style>
@@ -35,17 +34,7 @@
 
 {% block form_inputs %}
     {% for field in form %}
-        {% if field.field.widget.__class__.__name__ == "Textarea" %}
-            {{ render_field(field, size="s12", show_help_text_as_label=True) }}
-        {% elif field.field.widget.input_type == "select" %}
-            {{ render_field(field, size="s12 select2-wrapper") }}
-        {% elif field.field.widget.input_type == "file" %}
-            {{ render_field(field, size="s12", show_label=False) }}
-        {% elif field.field.widget.input_type == "checkbox" %}
-            {{ render_checkbox_field(field, size="s12") }}
-        {% else %}
-            {{ render_field(field, size="s12") }}
-        {% endif %}
+        {{ render_field(field, size="s12") }}
     {% endfor %}
 {% endblock %}
 
--- a/hackathon_site/review/jinja2/review/send-decisions.html
+++ b/hackathon_site/review/jinja2/review/send-decisions.html
@@ -1,15 +1,5 @@
 {% extends "event/form_base.html" %}
 
-{% block head_extends %}
-<!-- Select2 CSS -->
-<link href="https://cdn.jsdelivr.net/npm/select2@4.1.0-beta.1/dist/css/select2.min.css" rel="stylesheet" />
-<link rel="stylesheet" type="text/css" href="{{ static('event/styles/css/select2.css') }}" />
-
-<style>
-
-</style>
-{% endblock %}
-
 {% block form_title %}SEND DECISION EMAILS TO PARTICIPANTS{% endblock %}
 {% block form_attrs %}enctype="multipart/form-data"{% endblock %}
 
@@ -35,14 +25,16 @@
 
 {% block form_inputs %}
     {% for field in form %}
-        {% if field.field.widget.__class__.__name__ == "Textarea" %}
-            {{ render_field(field, size="s12", show_help_text_as_label=True) }}
-        {% elif field.field.widget.input_type == "select" %}
-            {{ render_field(field, size="s12 select2-wrapper") }}
-        {% elif field.field.widget.input_type == "file" %}
-            {{ render_field(field, size="s12", show_label=False) }}
-        {% elif field.field.widget.input_type == "checkbox" %}
-            {{ render_checkbox_field(field, size="s12") }}
+        {% if field.field.widget.input_type == "select" %}
+            <div class="input-field row col s12">
+                <select>
+                    <option value="" disabled selected>Choose your option</option>
+                    <option value="Accepted">Accepted</option>
+                    <option value="Waitlisted">Waitlisted</option>
+                    <option value="Rejected">Rejected</option>
+                </select>
+                <label>{{field.label}}</label>
+            </div>
         {% else %}
             {{ render_field(field, size="s12") }}
         {% endif %}
@@ -56,45 +48,9 @@
 {% endblock %}
 
 {% block scripts %}
-
-<!-- Select2 JS -->
-<script src="https://cdn.jsdelivr.net/npm/select2@4.1.0-beta.1/dist/js/select2.min.js"></script>
-
 <script>
-$(document).ready(function() {
-    const selects = $("select");
-    selects.not(".select2-school-select").select2();
-    $(".select2-school-select").select2({tags: true});  // Allow custom options for school
-
-    // Properly style any pre-filled selects
-    $.each(selects, (i, e) => {
-        if (e.value) {
-            $(e).siblings("label").addClass("label-active");
-        }
+    $(document).ready(function(){
+        $('select').formSelect();
     });
-
-    selects.on("select2:opening", function () {
-        $(this).siblings("label").addClass("label-active");
-    });
-
-    selects.on("select2:closing", function (e) {
-        if (!this.value) {
-            $(this).siblings("label").removeClass("label-active");
-        }
-
-        // steal focus during close - only capture once and stop propogation
-        // https://stackoverflow.com/a/49261426/3882202
-        $(e.target).data("select2").$selection.one("focus focusin", function (e) {
-            e.stopPropagation();
-        });
-    });
-
-    // on first focus (bubbles up to document), open the menu
-    // https://stackoverflow.com/a/49261426/3882202
-    $(document).on('focus', '.select2-selection.select2-selection--single', function () {
-      $(this).closest(".select2-container").siblings('select:enabled').select2('open');
-    });
-});
-
 </script>
 {% endblock %}
--- a/hackathon_site/review/jinja2/review/send-decisions.html
+++ b/hackathon_site/review/jinja2/review/send-decisions.html
@@ -3,9 +3,6 @@
 {% block form_title %}SEND DECISION EMAILS TO PARTICIPANTS{% endblock %}
 {% block form_attrs %}enctype="multipart/form-data"{% endblock %}
 
-
-
-
 {% set form_action=url("admin:send-decision-emails") %}
 
 {% block form_description %}
@@ -27,7 +24,7 @@
     {% for field in form %}
         {% if field.field.widget.input_type == "select" %}
             <div class="input-field row col s12">
-                <select>
+                <select id="status" name="status">
                     <option value="" disabled selected>Choose your option</option>
                     <option value="Accepted">Accepted</option>
                     <option value="Waitlisted">Waitlisted</option>
--- a/hackathon_site/review/jinja2/review/send-decisions.html
+++ b/hackathon_site/review/jinja2/review/send-decisions.html
@@ -6,16 +6,18 @@
 {% set form_action=url("admin:send-decision-emails") %}
 
 {% block form_description %}
-    <div class="row col s12">
-    {% for error in form.non_field_errors() %}
-        <span class="formError">{{ error }}</span>
-    {% endfor %}
-    </div>
+    {% if form.non_field_errors() %}
         <div class="row col s12">
-            Accepted: {{ accepted_count }}
-            Waitlisted: {{ waitlisted_count }}
-            Rejected: {{ rejected_count }}
+        {% for error in form.non_field_errors() %}
+            <span class="formError">{{ error }}</span>
+        {% endfor %}
         </div>
+    {% endif %}
+    <div class="row col s12">
+        Accepted: {{ accepted_count }}
+        Waitlisted: {{ waitlisted_count }}
+        Rejected: {{ rejected_count }}
+    </div>
 {% endblock %}
 
 {% block form_inputs %}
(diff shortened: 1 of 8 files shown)


hackathon_site/review/jinja2/review/send-decisions.html
```diff
@@ -0,0 +1,100 @@
+{% extends "event/form_base.html" %}
+
+{% block head_extends %}
+<!-- Select2 CSS -->
```
grahamhoyes: We don't need to use select2 for this, since there's nothing that requires searching an offsite list. Materialize has a regular select input (which we've just styled select2 to match) that would be simpler to use: https://materializecss.com/select.html

hackathon_site/review/jinja2/review/send-decisions.html
```diff
@@ -0,0 +1,100 @@
...
+        <div class="row col s12">
+            Accepted: {{ accepted_count }}
+            Waitlisted: {{ waitlisted_count }}
+            Rejected: {{ rejected_count }}
+        </div>
+    {% endif %}
+{% endblock %}
+
+{% block form_inputs %}
+    {% for field in form %}
```
grahamhoyes, line 16: This is overcomplicated for a form that only has 4 fields, that are never textareas, files, or checkboxes. Rendering the fields individually with the default `render_field` macro would be simpler

hackathon_site/review/templates/review/change_list.html
```diff
@@ -9,6 +9,7 @@
 &rsaquo; {{ cl.opts.verbose_name_plural|capfirst }}
     {% if has_change_permission %}
     <a class="button" style="float: right; margin: -4px 15px -4px -5px;" href="{% url 'admin:assign-reviewer-to-team' %}">Review Next Team</a>
+        <a class="button" style="float: right; margin: -4px 15px -4px -5px;" href="{% url 'admin:send-decision-emails' %}">Send Decisions</a>
```
grahamhoyes: Minor, extra indent

hackathon_site/review/views.py
```diff
@@ -1,3 +1,93 @@
...
+
+
+class MailerView(UserPassesTestMixin, FormView):
+    template_name = "review/send-decisions.html"
+    success_url = reverse_lazy("admin:send-decision-emails")
+
+    def test_func(self):
+        return self.request.user.is_superuser
+
+    def get_form(self, form_class=None):
```
grahamhoyes: can just specify `form_class = MailerForm` on the view, don't need to redefine this method

hackathon_site/review/views.py
```diff
@@ -1,3 +1,93 @@
...
+        date_end = form.cleaned_data["date_end"]
+        status = form.cleaned_data["status"]
+        quantity = form.cleaned_data["quantity"]
+
+        queryset = Review.objects.filter(
+            status=status,
+            updated_at__gte=date_start,
+            updated_at__lte=date_end,
+            decision_sent_date__isnull=True,
+        ).all()[:quantity]
```
grahamhoyes: Minor, don't need the `.all()` (it doesn't do anything here)

hackathon_site/review/views.py
```diff
@@ -1,3 +1,93 @@
...
+
+        # Get a mail socket connection and manually open it
+        connection = mail.get_connection(fail_silently=False)
+        connection.open()
+
+        for review in queryset:
```
grahamhoyes: Can you wrap this in a try/except/finally, that closes the connection on finally to make sure it doesn't get left open? Something like:

```python
try:
    for review in queryset:
        ...
except Exception as e:
    raise e
finally:
    connection.close()
```

hackathon_site/review/views.py
```diff
@@ -1,3 +1,93 @@
...
+            decision_sent_date__isnull=True, status="Waitlisted"
+        ).count()
+
+        context["rejected_count"] = Review.objects.filter(
+            decision_sent_date__isnull=True, status="Rejected"
+        ).count()
+
+        return context
+
+    def post(self, request, *args, **kwargs):
```
grahamhoyes: Can be removed

hackathon_site/review/tests.py
```diff
@@ -1,3 +1,356 @@
...
+            Q(codename="view_application", content_type__app_label="registration")
+            | Q(codename="view_review", content_type__app_label="review"),
+        )
+        self._login(superuser=False)
+        for perm in self.view_permissions:
+            self.user.user_permissions.add(perm)
+
+        response = self.client.get(reverse("admin:review_teamreview_changelist"))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertNotContains(response, "Send Decisions")
```
grahamhoyes, line 147: I'm not sure why this is passing, because the there's nothing to hide the button from non-superusers at the moment

hackathon_site/review/templates/review/change_list.html
grahamhoyes: Also, should there be something here to hide the button from non-superusers? Something outside the `if has_change_permissions` block, since superuser => change permissions

hackathon_site/review/tests.py
```diff
@@ -1,3 +1,356 @@
...
+
+    def test_sends_quantity_number_of_accepted_emails(self):
+        self._login()
+        self._create_teams_and_reviews_for_mail_tests()
+
+        quantity_before = len(
```
grahamhoyes: `quantity_before = Review.objects.filter(decision_sent_date__isnull=True, status="Accepted").count()`, not that performance matters here too much

hackathon_site/review/tests.py
```diff
@@ -1,3 +1,356 @@
...
+            Review.objects.filter(
+                decision_sent_date__isnull=True, status="Accepted"
+            ).all()
+        )
+
+        self.form_data["quantity"] = 3  # Send 3 acceptance emails
+
+        response = self.client.post(self.view, data=self.form_data)
+
+        quantity_after = len(
```
grahamhoyes: Ditto about `.count()`, saves having to evaluate the full queryset. Same for below as well, won't add comments to all of them

hackathon_site/review/tests.py
```diff
@@ -1,3 +1,356 @@
...
+            Review.objects.filter(
+                decision_sent_date__isnull=True, status="Accepted"
+            ).all()
+        )
+
+        self.assertEqual(len(mail.outbox), 7)
+        self.assertRedirects(response, self.view)
+        self.assertEqual(quantity_before, quantity_after + 7)
+
+    def test_send_all_waitlisted_within_current_date_range(self):
```
grahamhoyes: Is there any reason we'd expect the behaviour for waitlisted status (and rejected below) to differ from the test for accepted above, since they end up just getting passed in to the queryset filter, warranting separate tests? 

If there's something different about these tests that makes them necessary, that would be helpful to call out in a docstring

hackathon_site/review/tests.py
```diff
@@ -1,3 +1,356 @@
...
+        self.client.post(self.view, data=self.form_data)
+
+        clean = re.compile("<.*?>")
+        clean_mail_body = re.sub(clean, "", mail.outbox[0].body)
+
+        self.assertIn(
+            f"Congratulations, you’ve been accepted to { settings.HACKATHON_NAME }",
+            mail.outbox[0].subject,
+        )
+        self.assertIn(
```
grahamhoyes, line 217: I would also test the things that get passed in as context (user, request, and rsvp deadline). We can be pretty confident that the default jinja2 variables get passed in. Especially we'd like to make sure that the dashboard_url is built properly, which should be `http://testserver/dashboard/` in this case I think.

This isn't something that we did for the sign up emails, which I think would be warranted and I've created #213 for.

hackathon_site/review/tests.py
```diff
@@ -1,3 +1,356 @@
...
+            f"The {settings.HACKATHON_NAME} team has reviewed your application, and we’re excited to welcome you to CoolHacks!",
+            clean_mail_body,
+        )
+
+    def test_correct_text_in_waitlisted_email(self):
```
grahamhoyes, line 222: The waitlisted email currently renders the `final_review_response_date` wrong, it needs a `strptime` in there (it's just sticking the datetime in there in 2020-10-07 00:00:00-05:18 format right now).

grahamhoyes (commented):
I get some weird behaviour when submitting the form multiple times after a redirect (with a limit of 1 email) where it submits but doesn't send anything, I can't reproduce it consistently so may be user error.

It would be nice when the page is done, the form is re-rendered with the existing fields populated so the user doesn't have to enter the same date range multiple times. This would involve changing the form view to not redirect though and instead render and return a template. Since it's only a superuser thing, it doesn't have to be changed.

hackathon_site/review/jinja2/review/send-decisions.html
```diff
@@ -0,0 +1,53 @@
+{% extends "event/form_base.html" %}
+
+{% block form_title %}SEND DECISION EMAILS TO PARTICIPANTS{% endblock %}
+{% block form_attrs %}enctype="multipart/form-data"{% endblock %}
+
+{% set form_action=url("admin:send-decision-emails") %}
+
+{% block form_description %}
+    {% if form.non_field_errors %}
```
grahamhoyes: This should be `{% if form.non_field_errors() %}`, and is a mistake in all the other jinja templates as well. `form.non_field_errors` is a function, and `bool(<function>)` is always True. I'm guessing this was originally introduced when using a django template as an example, which doesn't need explicit function calls.

hackathon_site/review/jinja2/review/send-decisions.html
```diff
@@ -0,0 +1,53 @@
...
+        <div class="row col s12">
+        {% for error in form.non_field_errors() %}
+            <span class="formError">{{ error }}</span>
+        {% endfor %}
+        </div>
+        <div class="row col s12">
```
grahamhoyes: With the above fix, you'll want this outside the `form.non_field_errors()` check, since that should be false if there are no errors and hence this should be hidden.

hackathon_site/review/jinja2/review/send-decisions.html
```diff
@@ -0,0 +1,53 @@
...
+            Accepted: {{ accepted_count }}
+            Waitlisted: {{ waitlisted_count }}
+            Rejected: {{ rejected_count }}
+        </div>
+    {% endif %}
+{% endblock %}
+
+{% block form_inputs %}
+    {% for field in form %}
+        {% if field.field.widget.input_type == "select" %}
```
grahamhoyes: This doesn't seem to be necessary. `render_field(field, size="s12")` works fine for the select field on my end (the added js at the bottom was the secret sauce)

hackathon_site/review/tests.py
```diff
@@ -1,3 +1,267 @@
...
+        self.assertIn(rsvp_deadline, mail.outbox[0].body)
+        self.assertIn(settings.HACKATHON_NAME, mail.outbox[0].body)
+        self.assertIn(settings.PARTICIPANT_PACKAGE_LINK, mail.outbox[0].body)
+        self.assertIn(settings.CHAT_ROOM_LINK, mail.outbox[0].body)
+        self.assertIn(
+            f"Congratulations, you’ve been accepted to { settings.HACKATHON_NAME }",
+            mail.outbox[0].subject,
+        )
+        self.assertIn(
+            f"The {settings.HACKATHON_NAME} team has reviewed your application, and we’re excited to welcome you to CoolHacks!",
```
grahamhoyes: `CoolHacks` hard coded at the end of the sentence, can cut that out or replace it with another `{settings.HACKATHON_NAME}`

hackathon_site/review/views.py
```diff
@@ -1,3 +1,99 @@
...
+
+    def form_valid(self, form, **kwargs):
+        """
+        On form submission valid get the data and send emails depending on the status.
+
+        Accepted / Waitlisted / Rejected emails will be sent for applications reviewed
+        (with last_updated_date) between date_start and date_end. The number of emails
+        for that status that will be sent is dictated by the quantity.
+        """
+        date_start = form.cleaned_data["date_start"]
```
grahamhoyes:
```python
date_start = datetime.combine(
    form.cleaned_data["date_start"], datetime.min.time()
).replace(tzinfo=settings.TZ_INFO)
date_end = datetime.combine(
    form.cleaned_data["date_end"], datetime.max.time()
).replace(tzinfo=settings.TZ_INFO)
```

Since the queryset expects a timezone-aware datetime object for `updated_at`, it's throwing warnings because it's getting a plain date object. This fixes that

hackathon_site/review/jinja2/review/send-decisions.html
```diff
@@ -0,0 +1,41 @@
+{% extends "event/form_base.html" %}
+
+{% block form_title %}SEND DECISION EMAILS TO PARTICIPANTS{% endblock %}
+{% block form_attrs %}enctype="multipart/form-data"{% endblock %}
```
grahamhoyes: dkm for being blind, but this is unnecessary

hackathon_site/review/jinja2/review/send-decisions.html
```diff
@@ -0,0 +1,41 @@
...
+
+{% set form_action=url("admin:send-decision-emails") %}
+
+{% block form_description %}
+    {% if form.non_field_errors() %}
```
grahamhoyes: Might as well just remove this I guess, there are no non-field errors anyway

grahamhoyes (approved):

PR #220, opened 2020-11-03
Sprinkled in some brackets for form.non_field_errors

### LINKED ISSUE #215: Fix checks of form.non_field_errors in jinja templates
In many of our jinja2 templates, we have `{% if form.non_field_errors %}`. `form.non_field_errors` is a function, and in jinja2 templates they need to be explicitly called. `bool(<function>)` is always True, so currently we're unnecessarily rendering some elements for errors. Should be replaced with `{% if form.non_field_errors() %}`. Do a grep to find all jinja2 templates that need fixing.

This likely got introduced when taking inspiration from a django template, which does not require (or support) explicit function calls. Shouldn't require any additional tests and no current tests should fail.

albogdan:
## Overview

- Resolves #215 
- Sprinkled in some brackets in 6 files found by `grep -r "form.non_field_errors" .` 
- Read the issue for more information

## Unit Tests Created

- Ran previously created unit tests

## Steps to QA

- Look, see if makes sense

### DIFF
--- a/hackathon_site/event/jinja2/event/change_password.html
+++ b/hackathon_site/event/jinja2/event/change_password.html
@@ -16,7 +16,7 @@
     {{ render_field(form.new_password1, show_help_text=True) }}
     {{ render_field(form.new_password2) }}
 
-    {% if form.non_field_errors %}
+    {% if form.non_field_errors() %}
         {% for error in form.non_field_errors() %}
             {{ error }}
         {% endfor %}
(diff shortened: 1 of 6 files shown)


arminale:
Looks good

arminale (approved):

grahamhoyes (approved):
LGTM. No more instances of `form.non_field_errors` without the function call, and nothing that breaks page layouts.

PR #226, opened 2020-12-02
Fixed admin squared count for team member count

### LINKED ISSUE #209: Filtered Review Teams square members count
A team can have a maximim of 4 users, but when viewing teams under Review > Teams, the team member count will be squared.
image

This is because the count operations that happen in the filter end up doing another join, which squares the number of rows. Solution should be to do a count distinct.

lzm9800:
Made changes to `review/admin.py` and added a test to `review/test_admin.py`

## Overview

- Resolves #209 
- Added `distinct=True` in `Count()` in the class `TeamReviewAdmin` in `review/admin.py`
- Added a test case for team member count with filter in `review/test_admin.py`

## Unit Tests Created

- `test_list_page_shows_team_member_count_with_filters`
    - tests team member count with filter: `reviewed=false`

## Steps to QA

- Run Unit Test: 
    - `python manage.py test --settings=hackathon_site.settings.ci review.test_admin`
- Check on Admin site: Create a team for them under Registration > Teams (not event > teams, that's something else). Then create an application for each of them from Registration > Applications. Once that's all done, under Review > Teams you can enable the "Not reviewed" filter to make sure the members count is correct instead of squared

### DIFF
--- a/hackathon_site/review/admin.py
+++ b/hackathon_site/review/admin.py
@@ -203,7 +203,7 @@ class TeamReviewAdmin(admin.ModelAdmin):
             .prefetch_related(
                 "applications", "applications__review", "applications__user"
             )
-            .annotate(members_count=Count("applications"))
+            .annotate(members_count=Count("applications", distinct=True))
             .annotate(most_recent_submission=Max("applications__updated_at"))
         )
 
--- a/hackathon_site/review/test_admin.py
+++ b/hackathon_site/review/test_admin.py
@@ -79,6 +79,20 @@ class TeamReviewListAdminTestCase(SetupUserMixin, TestCase):
         # More precise to look for this than just the number "3"
         self.assertContains(response, '<td class="field-get_members_count">3</td>')
 
+    def test_list_page_shows_team_member_count_with_filters(self):
+        """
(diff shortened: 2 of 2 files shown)


grahamhoyes:
@lzm9800 For now, you'll need to set the `DJANGO_SECRET_KEY` in your forked repository so that actions can run correctly. We'll need to come up with a workaround for this eventually

grahamhoyes:
@lzm9800 We've fixed the issue with actions, you'll need to update your branch in order for them to run. You can do that with the "update branch" button below

albogdan (commented):
Code looks good to me. A couple of things with regards to PR description:

1. You can use backticks to make your code more readable. For example instead of
 "distinct=True in Count() in admin.py"
to make it cleaner you can write
   `distinct=True` in `Count()` in `admin.py`
by writing the following in Markdown:
```
`distinct=True` in `Count()` in `admin.py`
```
You can format any code, commands and files using that format to make your writing cleaner.

2. We have multiple `admin.py` files. You should clarify which one the changes were made in, inside the PR description, for example `review/admin.py`. 

Thanks!

lzm9800:
I have edited the PR description @albogdan.

albogdan (approved):

Issue #225, opened 2020-12-01
Update CONTRIBUTING.md with our latest practices


grahamhoyes:
As more people are working on this project, we should write down the implicit practices we've been using. These include:
- Naming branches `<issue number>-<description>`
- Formatting with black for python, prettier for javascript and css (separate for event app and react)
- Writing tests
- Creating and claiming issues
- Creating a fork and submitting a PR if you're external to ieee
- Things to keep in mind, such as changing the registration close date in the settings file so you can get to certain pages. Don't commit those changes!
- Links to relevant getting started documentation

grahamhoyes:
Because of the issues in #227, also have people set the `DJANGO_SECRET_KEY` in their forks so they can run tests. This is a medium-term workaround, in the long run we should get the `pull_request_target` trigger working.

grahamhoyes:
Don't need to do that anymore, we got things working in #227

grahamhoyes:
Add instructions for maintainers on how to check out pull requests from forks locally. Eg for pr 111:

```
git fetch origin pull/111/head:pr/111
git checkout 111
```

grahamhoyes:
Update the PR template too so that it says `resolves #<number>`, since some people replace the # sign

grahamhoyes:
Perhaps add a "first time section" that links to all the places in the readme to look at, including docker, running `yarn run scss`, etc.

PR #227, opened 2020-12-02
Change trigger in CI/CD workflow

grahamhoyes:
## Overview

- Changes the trigger of the main workflow from `pull_request` to `pull_request_target`. This is largely the same, except actions get run on the base repository instead of the head - this means that secrets (`DJANGO_SECRET_KEY` in our case) can be used from the base repository, so all contributors don't need to set them up.
- This is motivated by #226, which is failing checks because the secret isn't set in their forked repo.
- Read more about the `pull_request_target` event [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,7 @@
 name: CI/CD
 
 on:
-  pull_request:
+  pull_request_target:
     branches: '**'
 
 jobs:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request_target.head.ref}}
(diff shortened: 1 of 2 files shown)


albogdan (dismissed):

grahamhoyes:
Actions aren't running for some reason, going to make another commit to try and nudge it

grahamhoyes:
There's something weird going on here with running actions on this pr, it may require doing a force merge which we don't want to do right now.

Since this isn't critical at the moment, going to close and mention in #225 that contributions from forks have to set the django secret key for now (there aren't any security implications on this template repo of that). In the long run if we want things like automated comments, we'll have to figure out the `pull_request_target` event (maybe once GitHub has better docs on it).

grahamhoyes:
Reopening this, since setting secrets in the forked repo does not appear to work. There are some things to try here still

albogdan (approved):

PR #229, opened 2020-12-02
Create a test that fails

grahamhoyes:
Just to make sure actions still run from the correct commit after switching to `pull_request_target` in #227


grahamhoyes:
Well that's a problem

grahamhoyes:
It worked!

PR #233, opened 2020-12-02
fix that opens the offending select2 element

### LINKED ISSUE #175: Select2 Fields aren't focusable when invalid
The application form can't be submitted when any of the select2 dropdowns are empty (since they're required), but since the actual select element is hidden it can't be focused on and doesn't alert the user. A not focusable error gets thrown in the console:

image

We'd like some way of getting the HTML5 validation popup back, which means making the hidden select element focusable, though anything that alerts the user of the missing field would be acceptable.

Hangukkid:
## Overview

- Resolves #175 
- Changes to the css that removes the console log errors
- When user attempts to submit an empty select2, it opens up the select2 element alerting the user 
- Trade off: Does not show the usual "required" popups as other fields.

## Unit Tests Created

- None, css changes.

## Steps to QA

- Login into template site and create application at /registration/application/
- Fill any arrangement of inputs but leave at least one select2 empty
- Completed if empty select2 box always opens

### DIFF
--- a/hackathon_site/event/static/event/styles/scss/select2.scss
+++ b/hackathon_site/event/static/event/styles/scss/select2.scss
@@ -180,11 +180,13 @@ THE SOFTWARE.
     border: 0 !important;
     clip: rect(0 0 0 0) !important;
     height: 1px !important;
-    margin: -1px !important;
+    margin: 0 auto;
     overflow: hidden !important;
     padding: 0 !important;
     position: absolute !important;
     width: 1px !important;
+    display: block;
+    opacity: 0;
 }
 .select2-container--default {
     .select2-selection--single {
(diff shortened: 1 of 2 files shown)


grahamhoyes:
@lisa-sa-li Can you take a look at this? Getting the same "required" validation popup to appear as the other fields seems difficult, but this still highlights to the user what they are missing.

There's some strangness with scroll where it's partially underneath the navbar too

lisa-sa-li:
I don't think much can be done to improve the behaviour right now. This achieves the basic goal of notifying to the user that they didn't fill a select2 form. @grahamhoyes Create a new issue to fix the scroll-top issue? I'm fine with approving this once the change below is fixed:

- In `styles.scss,` set `z-index: 1100` for the class `.navbar` on line 55. This fixes an issue where the dropdown is above the navbar. (the dropdown z-indez is 1051)

lisa-sa-li (approved):

PR #257, opened 2021-05-03
Move Event App Yarn Stuff

### LINKED ISSUE #241: Move the non-React yarn commands up from the event directory
We have a package.json file in the event directory at the moment, used for scss compiling and prettier checks on everything other than the react app: https://github.com/ieeeuoft/makeuoft-site-2021/blob/develop/hackathon_site/event/package.json

Can this be moved up a directory so that it's in the main `hackathon_site` folder? That way we wouldn't have to cd into event to run `yarn scss`, and could prettier check other directories as well. The commands (for scss compiling and prettier checks, in package.json and the CI workflows) would need to be updated as well for the new paths.

grahamhoyes:
## Overview

- Resolves #241 
- Moves the package.json for Django templates and such up from the event directory, to the hackathon_site directory
- Fixes the paths in the commands as necessary
- Replaces `node-sass` with regular `sass`, because the former is apparently [deprecated](https://www.npmjs.com/package/node-sass) in favor of the new Dart version
- Updates the readme and contributing guidelines for new scss / prettier commands
- Renames the CI workflow from "event-checks" to "template-checks" (any better ideas?), and updates with the new package.json location

This is a **backwards incompatible change** with current deployments, since it will break their CI workflows

## Unit Tests Created

- Nada

## Steps to QA

- Delete the node_modules folder in your existing event folder
- Delete `hackathon_site/event/static/event/styles/css`
    - Run the local dev server, make sure styles are gone
- Run `yarn scss` or `yarn scss-watch` to make sure they pick up the files properly (check that the dev server has styles again)
- Mess up a js or scss file somewhere, and run `yarn prettier-check` to catch it, then `yarn prettier` to resolve it

As for the workflow, because we use `pull_request_target` triggers now, we just cross our fingers and hope it works once we merge.


grahamhoyes:
The event-checks CI target should fail, since there's no package.json in the event folder anymore. Again, since we're using `pull_request_target` triggers, the CI workflows are coming from the base branch rather than the PR. If everything else looks good, this will have to be a force merge PR.

lisa-sa-li (approved):

PR #260, opened 2021-05-05
Installed Flow and replaced node-sass with sass in Dashboard App

lisa-sa-li:
## Overview

- Adds [Flow](https://flow.org/en/docs/getting-started/) to Dashboard app so we can type API responses
- Removes `node-sass `and replaces with `sass` (Was having yarn issues with Flow before doing this)

## Unit Tests Created

N/A

## Steps to QA

- In the hackathon_site/dashboard/frontend directory, run:
`yarn install --frozen-lockfile`
`yarn flow`

This will start the Flow server and `No errors!` should return since none of our files have been set up with Flow yet.

### DIFF
--- a/hackathon_site/dashboard/frontend/.babelrc
+++ b/hackathon_site/dashboard/frontend/.babelrc
@@ -1,0 +1,3 @@
+{
+    "presets": ["@babel/preset-flow"]
+}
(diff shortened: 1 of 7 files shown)


grahamhoyes:
Closing since we used typescript, and have upgraded sass elsewhere

PR #263, opened 2021-05-05
Remove inventory filter from drawer on desktop

lisa-sa-li:
## Overview

- On the inventory page on desktop, the filter will no longer be on a drawer to match the redesign on Figma
- Clicking on an item will open up the Product Overview drawer (with a static layout/item information)
- An CloseIcon will appear on the search bar, once we hook up Formik to it, we should conditionally render the icon to appear only if the input has a value

## Unit Tests Created

N/A

## Steps to QA

- Go to the inventory page
- Confirm that the inventory filter looks the same as the [Figma](https://www.figma.com/file/1bYQy6ioaZwxPD9w6D47jW/Hardware-signout-site?node-id=2640%3A20442) mobile and desktop design
- Click on an item, the Product Overview drawer should popup
- An `x` icon should be on the search bar, but pressing it won't do anything

### DIFF
--- a/hackathon_site/dashboard/frontend/package.json
+++ b/hackathon_site/dashboard/frontend/package.json
@@ -36,7 +36,7 @@
     "eject": "react-scripts eject",
     "i": "yarn install --frozen-lockfile",
     "pretty": "yarn prettier --write 'src/**/*.(scss|js)'",
-    "precheck": "yarn i && yarn test && yarn pretty"
+    "precheck": "yarn pretty && yarn test -o -u --watchAll=false"
   },
   "eslintConfig": {
     "extends": "react-app"
--- a/hackathon_site/dashboard/frontend/package.json
+++ b/hackathon_site/dashboard/frontend/package.json
@@ -15,7 +15,6 @@
     "formik": "^2.1.4",
     "history": "^4.10.1",
     "lint-staged": "^10.2.2",
-    "node-sass": "^4.14.0",
     "notistack": "^0.9.17",
     "prettier": "^2.0.5",
     "react": "^16.13.1",
@@ -27,6 +26,7 @@
     "redux": "^4.0.5",
     "redux-mock-store": "^1.5.4",
     "redux-thunk": "^2.3.0",
+    "sass": "^1.32.12",
     "yup": "^0.29.1"
   },
   "scripts": {
--- a/hackathon_site/dashboard/frontend/package.json
+++ b/hackathon_site/dashboard/frontend/package.json
@@ -33,7 +33,10 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --env=jest-environment-jsdom-sixteen",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "i": "yarn install --frozen-lockfile",
+    "pretty": "yarn prettier --write 'src/**/*.(scss|js)'",
+    "precheck": "yarn i && yarn test && yarn pretty"
   },
   "eslintConfig": {
     "extends": "react-app"
--- a/hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
+++ b/hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
@@ -5,6 +5,7 @@ import { Formik, Field } from "formik";
 import Typography from "@material-ui/core/Typography";
 import Button from "@material-ui/core/Button";
 import Chip from "@material-ui/core/Chip";
+import Paper from "@material-ui/core/Paper";
 import Divider from "@material-ui/core/Divider";
 import Checkbox from "@material-ui/core/Checkbox";
 import Radio from "@material-ui/core/Radio";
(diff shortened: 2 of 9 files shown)


hackathon_site/dashboard/frontend/package.json
```diff
@@ -15,7 +15,6 @@
     "formik": "^2.1.4",
     "history": "^4.10.1",
     "lint-staged": "^10.2.2",
-    "node-sass": "^4.14.0",
```
grahamhoyes, line 18: Since this is being done in #260, can the change be left out of this PR? Revert the changes to yarn.lock as well if so.

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
```diff
@@ -78,36 +79,38 @@ export const InventoryFilter = ({
...
-                    <Typography variant="h2">Categories</Typography>
-                </legend>
-                <Field
-                    name="inventoryCategories"
-                    component={CheckboxCategory}
-                    options={categories}
-                />
-            </fieldset>
-        </form>
+        <Paper elevation={2} className={styles.filterPaper} square={true}>
```
grahamhoyes, line 82: `styles.filterPaper` doesn't exist

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.js
```diff
@@ -5,137 +5,177 @@ import Drawer from "@material-ui/core/Drawer";
...
+import { productInformation, inventoryItems } from "testing/mockData";
 
 const Inventory = () => {
     const [mobileOpen, setMobileOpen] = React.useState(false);
-
     const toggleFilter = () => {
         setMobileOpen(!mobileOpen);
     };
 
-    const mobileWidth = useMediaQuery(useTheme().breakpoints.up("md"));
```
grahamhoyes, line 28: `useTheme` import can be removed

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.js
```diff
@@ -5,137 +5,177 @@ import Drawer from "@material-ui/core/Drawer";
...
+        alert("Add to cart");
+        setItemOverview(null);
+    };
+
+    const [itemOverview, setItemOverview] = React.useState(null);
+    const toggleMenu = () => {
+        setItemOverview(null);
     };
 
+    // itemOverview will be sent to api
```
grahamhoyes: Shouldn't need to send anything to the API since we should already have all the information in Redux, but it will dispatch an action to show the overview drawer for the appropriate item id (which will then pull what's needed from Redux)

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.js
```diff
@@ -5,137 +5,177 @@ import Drawer from "@material-ui/core/Drawer";
...
-                                    Filter
-                                </Button>
-                            </Hidden>
-
-                            <div className={styles.inventoryBodyToolbarRefresh}>
-                                <Typography variant="body2">123 items</Typography>
+                    <div className={styles.inventoryBodyRight}>
+                        <div className={styles.inventoryBodyToolbar}>
+                            <div className={styles.inventoryBodyToolbarDiv}>
+                                <TextField
```
grahamhoyes, line 100: According to [this](https://stackoverflow.com/questions/58998746/material-ui-textfield-type-search-not-showing-clear-icon) Stack Overflow post, the clear icon only appears when the field is uncontrolled. Since eventually this will need to be hooked up to formik as well, making it a controlled component, that won't work for the clear button.

You could do something like:

```
<TextField
    className={styles.inventoryBodyToolbarSearch}
    id="search-input"
    label="Search items"
    variant="outlined"
    type="text"
    InputProps={{
        endAdornment: (
            <InputAdornment position="end">
                <IconButton>
                    <CloseIcon />
                </IconButton>
            </InputAdornment>
        ),
    }}
/>
```

And add stuff in so that the icon only shows when there's a value for the field (needs formik) and resets the field.

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
lisa-sa-li: it does, it's just in a media query. Line 46 in the scss

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.js
lisa-sa-li: Added the adornment. Will add Formik in a diff PR, this one is a style change

hackathon_site/dashboard/frontend/package.json
lisa-sa-li: we might not merge the Flow one, so I think it would be better to do it here. I only did it in the Flow one because the scss wouldn't compile with node-sass after I installed Flow

hackathon_site/dashboard/frontend/package.json
```diff
@@ -27,13 +26,17 @@
...
   },
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --env=jest-environment-jsdom-sixteen",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "i": "yarn install --frozen-lockfile",
+    "pretty": "yarn prettier --write 'src/**/*.(scss|js)'",
+    "precheck": "yarn pretty && yarn test -o -u --watchAll=false"
```
lisa-sa-li, line 39: once I add Flow to this, precheck will just be something you should run before you commit

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.js
grahamhoyes: Ah, guess that's a limitation of Pycharm

grahamhoyes (dismissed):

grahamhoyes (approved):

PR #265, opened 2021-05-06
Migrate Admin site CSV downloading from makeuoft-site-2020 PR#30

### LINKED ISSUE #264: Mirgrate CSV Downloading from makeuoft-site-2020
* We implemented CSV download functionality for the Applications page in the Admin site in makeuoft-site-2020 [PR#30](https://github.com/ieeeuoft/makeuoft-site-2021/pull/30/)
* This needs to be migrated to the hackathon template as well
* Should implement unit tests as well checking if the CSV is properly made
    * https://stackoverflow.com/questions/29026779/testing-custom-admin-actions-in-django
    * https://stackoverflow.com/questions/41121785/how-to-test-file-response-in-django/47460338#47460338

albogdan:
## Overview

- Resolves #264
- Migrated [PR#30](https://github.com/ieeeuoft/makeuoft-site-2021/pull/30/) from the makeuoft-site-2020
- Created unit tests for the CSVs to ensure only checked off items are included into the csv
- Note that the `resume_sharing` column was not included since that column is not in the `hackathon-template` models and is unique to the `makeuoft-site-2020` schema.

## Unit Tests Created

- Unit test that checks if a HTTP403 error occurs when necessary permissions aren't acquired.
- Unit test that checks includes 8 applications and requests 4 of them are included in the CSV. Checks that only those 4 are included.

## Steps to QA

- Run the unit test
- Create some applications in the admin site(at least 2-3). Go to the admin site `Registration -> Applications` and check off which applications you want in the CSV. Then click on the dropdown and export to csv. Check to make sure that only the checked off applications were included in the CSV and that the data is correct.

### DIFF
--- a/hackathon_site/registration/test_admin.py
+++ b/hackathon_site/registration/test_admin.py
@@ -1,0 +1,62 @@
+import csv
+import io
+
+from django.test import TestCase
+from django.urls import reverse
+from rest_framework import status
+
+from hackathon_site.tests import SetupUserMixin
+from registration.admin import ApplicationAdmin
+
+
+class ExportCsvAdminTestCase(SetupUserMixin, TestCase):
+    """
+    Test that CSV exportation in the Registration/Application tab works as expected.
+    """
+
+    def setUp(self):
+        super().setUp()
+        self.application_view = reverse("admin:registration_application_changelist")
+        self.export_fields = [field[0] for field in ApplicationAdmin.export_fields]
+
+        self.user.is_staff = True
+        self.user.save()
+
+        self.team1 = self._make_full_registration_team(self_users=False)
+        self.team2 = self._make_full_registration_team(self_users=False)
+
+        self.data = {
+            "action": "export_as_csv",
+            "_selected_action": [i.id for i in self.team1.applications.all()],
+        }
+
+    def test_permission_denied(self):
+        self._login()
+
+        response = self.client.post(self.application_view, self.data)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_returns_csv_successfully(self):
+        self.user.is_superuser = True
+        self.user.save()
+        self._login()
+
+        response = self.client.post(self.application_view, self.data)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        content = response.content.decode("utf-8")
+        cvs_reader = csv.reader(io.StringIO(content))
+
+        body = list(cvs_reader)
+        headers = body.pop(0)
+
+        self.assertEqual(headers, self.export_fields)
+        self.assertEqual(len(self.team1.applications.all()), len(body))
+
+        response_emails = [user[2] for user in body]  # Emails at index 2 of CSV
+        expected_emails = [app.user.email for app in self.team1.applications.all()]
+        non_expected_emails = [app.user.email for app in self.team2.applications.all()]
+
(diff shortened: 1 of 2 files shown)


hackathon_site/registration/test_admin.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_returns_csv_successfully(self):
+        self.user.is_superuser = True
+        self.user.save()
+        self._login()
+
+        response = self.client.post(self.application_view, self.data)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        content = response.content.decode("utf-8")
+        cvs_reader = csv.reader(io.StringIO(content))
```
grahamhoyes, line 48: `csv_reader`

hackathon_site/registration/test_admin.py
```diff
@@ -0,0 +1,62 @@
...
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_returns_csv_successfully(self):
+        self.user.is_superuser = True
+        self.user.save()
+        self._login()
+
+        response = self.client.post(self.application_view, self.data)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
grahamhoyes, line 46: Test for the content type as well?

hackathon_site/registration/test_admin.py
```diff
@@ -0,0 +1,62 @@
...
+            "_selected_action": [i.id for i in self.team1.applications.all()],
+        }
+
+    def test_permission_denied(self):
+        self._login()
+
+        response = self.client.post(self.application_view, self.data)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_returns_csv_successfully(self):
```
grahamhoyes, line 39: Could this test be made a bit more thorough? Currently it's testing just headers and emails. At the very least, it would be good to test the name and review status fields since they use the `rgetattr` function which is untested for nested attributes currently

hackathon_site/registration/test_admin.py
```diff
@@ -0,0 +1,62 @@
...
+
+class ExportCsvAdminTestCase(SetupUserMixin, TestCase):
+    """
+    Test that CSV exportation in the Registration/Application tab works as expected.
+    """
+
+    def setUp(self):
+        super().setUp()
+        self.application_view = reverse("admin:registration_application_changelist")
+        self.export_fields = [field[0] for field in ApplicationAdmin.export_fields]
```
grahamhoyes, line 20: Would it be better to hard code these in the test rather than importing them? There may be an edge case where someone typos the class attribute and this wouldn't catch it, but that may be far enough of an edge case that we don't care.

grahamhoyes:
Would it be better to put the mixin somewhere more central than in the review app? We may end up using it for other exports as well (for the tech team to export hardware for example), so maybe having the mixin in `hackathon_site/utils.py` would be good?

You could create a `test_utils.py` or something in there to test the mixin in isolation, you'd just need to create a class that subclasses it in the test file to provide `export_fields` and calls `test_export_csv_mixin.export_as_csv(None, queryset)` with a manually written queryset. You could reuse the fields from the current tests in there, then when people use the mixin they just have to test that the correct fields are there since they can assume the mixin itself is tested.

albogdan:
Issue solved with PR #272

PR #267, opened 2021-05-07
Refactor flexPosition mixin

lisa-sa-li:
## Overview

- Before `flexPosition`'s default values were `justify-content: space-between` and `align-items: center`, the reasoning being that this was the most common configuration of flexbox we used. However, I realized this isn't clear to someone reading the code so I thought it would be better to use `flexPosition` as a short-form way to call flexbox (thereby reducing up to 4 lines of code to just 1).
- Now `flexPosition` will only add properties that are not its default. Also flex-direction was added as a parameter

## Unit Tests Created

N/A

## Steps to QA

- Go through all the changes, view the component's style in console, and confirm it's the same as it was prior, or just trust me :)

### DIFF
--- a/hackathon_site/dashboard/frontend/src/App.scss
+++ b/hackathon_site/dashboard/frontend/src/App.scss
@@ -3,15 +3,11 @@
 
 .App {
     background-color: color(backgroundC);
-    display: flex;
-    flex-direction: column;
-    justify-content: space-between;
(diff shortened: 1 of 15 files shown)


grahamhoyes (approved):
This is like reading another language (literally), but I think everything looks good to me. Page layouts are still behaving from what I can tell

PR #274, opened 2021-06-04
IEEE-9: Add a default "Application Reviewers" group

### LINKED ISSUE #243: Add a default "Application Reviewers" group
Since any instance of the platform using the admin site for reviews needs the appropriate user permissions set up, create a migration in the review app to add an "Application Reviewers" group with the following permissions:
- User view
- Application view
- Review add
- Review change
- Review view
- Review Team view

arminale:
## Overview

- Resolves #243 
- [IEEE-9](https://ieeeuoft.atlassian.net/browse/IEEE-9?atlOrigin=eyJpIjoiNjI2YmY0MjAwZDM5NDgwZDhmNWIwZWI1MDBjYjcxOWMiLCJwIjoiaiJ9)
- Added a revertible migration to create an Application Reviewers group
- The group has the following permissions:
   - User view
   - Application view
   - Review add
   - Review change
   - Review view
   - Review Team view
- Modified existing reviewer admin unit tests to use the group, instead of assigning change permissions. "View only" unit tests are unchanged.
    - I realize that these unit tests now check something slightly different from the old ones. In my opinion this is okay and good because going forward the reviewers will not be assigned individual permissions, only the group. Therefore, we need to test the forms with the group. While we could duplicate the tests, once for the group and once for the actual change permissions, that seems like overkill to me.
    - The "view only" tests are used to ensure that users without add/change permission aren't able to load the review forms and have a security function, so I kept them as is.

## Unit Tests Created

- Created a unit test to check if the Application Reviewers group has exactly the permissions specified in `review` app root

## Steps to QA

- Run `python manage.py migrate review` to apply. Login to the website and verify that the group has been created. Assign it to a new user.
    - Make sure the user can view applications, reviews, team reviews, and users. 
    - Make sure the user can add and change, but not delete reviews
    - Make sure the user has no other access
- Run `python manage.py migrate review 0002` to revert the migration. Login and make sure the group has been deleted.


albogdan (changes requested):
Overall LGTM - please change the name of the migration though. It should be `0003_reviewergroup.py` instead of `0003_reviewergoup.py`

grahamhoyes:
My only concern about using the new group in the tests instead of permissions is that it hides the fact that we should be explicitly testing the permissions - in theory, someone could have written a custom `@user_passes_test` permission decorator that looks for the group and the tests would pass, rather than the permissions. Anyone with those permissions should be able to do reviews even without being on the group - admins, for example.

Strictly, I think the test that the group has the correct permissons is the only test that needed changing.

I'm fine either way though, since it's a pedantic detail - @albogdan, do you have any concerns with this?

arminale:
I wasn't aware of `@user_passes_test`. I'll revert the tests to check for permissions and then write the `@user_passes_test` for the group. Thanks letting me know!

Also thanks for catching the typo @albogdan

grahamhoyes:
> I wasn't aware of `@user_passes_test`. I'll revert the tests to check for permissions and then write the `@user_passes_test` for the group. Thanks letting me know!
> 
> Also thanks for catching the typo @albogdan

@arminale [`@user_passes_test`](https://docs.djangoproject.com/en/3.2/topics/auth/default/#django.contrib.auth.decorators.user_passes_test) was just an example I was giving of where the conditions this test checks for may be subverted, it's a decorator for view permissions. I think the only thing you need to do is revert the changes to the other tests so that they have the permissions hard coded into the tests, and keep the one you added making sure the added group has the correct permissions.

arminale:
Better late than never I guess...
I reverted the tests to use directly hardcoded permissions instead of the group

hackathon_site/review/migrations/0003_reviewergoup.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 3.1.8 on 2021-06-04 02:42
+
+from django.db import migrations
+from django.core.management.sql import emit_post_migrate_signal
+from review import REVIEWER_PERMISSIONS
```
grahamhoyes: Since this import is here, the migration is not static - changing this variable will cause the migration to have different results for different people. Hard-code the permissions into this file

hackathon_site/review/test_admin.py
```diff
@@ -267,6 +270,23 @@ def test_permission_denied(self):
         response = self.client.get(self.change_view)
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
+    def test_reviewer_group_correct_permissions(self):
+        """
+        Ensure the reviewer group has all the permissions it is supposed to and nothing else.
```
grahamhoyes, line 274: You should actually hard-code the expected permissions into this test. This test currently checks that the things in the variable are set, but not that they are the ones that we want.

This also goes along with my other comment. Migrations are run during tests, so if someone were to remove something from that array, the migration would update accordingly (bad) and this test would still pass (also bad).

grahamhoyes (commented):
Sorry to welcome you back by changing my mind on something I didn't notice a month ago!

Let me know if my comments don't make sense.

hackathon_site/review/test_admin.py
grahamhoyes: Since the permissions should be hard-coded into the migrations and test, you don't really need the `REVIEWER_PERMISSIONS` variable at all.

albogdan (approved):

PR #280, opened 2021-06-08
IEEE-21 /api/event/teams/team

arshinmar:
## Overview

- Resolves Issue 21
- Piggybacks from #131
- Added API functionality for getting team info.
- Created a team serializer.
- Checks in tests.py ensure that profile, team and user are effectively linked and serializer info matches expected model.

## Unit Tests Created

- Added test additions to test_API file.
- Added serializer & linking tests to tests.py. 

## Steps to QA

- Simply run regular tests.


.gitignore
```diff
@@ -154,3 +155,6 @@ hackathon_site/**/*.css.map
 
 # node_modules, found outside the react app
 node_modules/
+
+# OS Stuff
+.DS_Store
```
grahamhoyes: Newline at end of file

hackathon_site/event/models.py
```diff
@@ -27,7 +27,7 @@ def __str__(self):
 class Profile(models.Model):
     user = models.OneToOneField(User, on_delete=models.CASCADE)
     team = models.ForeignKey(
-        Team, related_name="profiles", on_delete=models.CASCADE, null=False
+        Team, related_name="profile", on_delete=models.CASCADE, null=False
```
grahamhoyes: Two things:
- This should stay as profiles. Since this is a foreign key, we know that the relationship between Profiles and Teams is many to one, therefore team to profile is one to many. The `related_name` sets up the reverse relationship property on teams - which we want to be `some_team.profiles`
- In the future, any time you modify a model file, you have to generate a migration with `python manage.py makemigrations` (not in this case, since this change will be reverted). That will provide the database with instructions to upgrade the schemas, and in this case things like `related_name` are also stored in migrations to build a history of the model.

hackathon_site/event/serializers.py
```diff
@@ -10,7 +10,20 @@ class Meta:
         fields = ("id", "name")
 
 
+class TeamSerializer(serializers.ModelSerializer):
```
grahamhoyes, line 13: It wasn't in the ticket, but I think we should add a profiles serializer into here. That way, on the frontend, we can just get `/api/event/teams/team/`, and we'll have all the information we need about the team (aka, the members on it), rather than having to make a separate request.

When we eventually have a list view of `/api/event/teams/` for the admin side of the site, that will also come in handy so we don't have to fetch the list of team members separately.

To do that, you won't want to use the `ProfileSerializer` below, since that includes the team - so you'll end up with a circular reference. I would create a new `ProfileInTeamSerializer`, that includes the profile as well as important related user fields - first name, last name, email. I think you can just put `user__first_name` etc in the fields list, otherwise you'll need a user serializer too.

hackathon_site/event/test_api.py
```diff
@@ -37,25 +43,60 @@ def test_user_has_no_profile(self):
...
-                    "id",
-                    "id_provided",
-                    "attended",
-                    "acknowledge_rules",
-                    "e_signature",
-                )
-            },
-            "groups": [{"id": self.group.id, "name": self.group.name}],
-        }
+        # expected_response = {
```
grahamhoyes: Fix this up

hackathon_site/event/tests.py
```diff
@@ -648,3 +658,74 @@ def test_reset_password_confirm_valid_submit_redirect(self):
...
+
+        user_expected = {
+            "id": user.id,
+            "first_name": user.first_name,
+            "last_name": user.last_name,
+            "email": user.email,
+            "profile": profile_serialized,
+            "groups": group_serialized,
+        }
+        if user_expected != user_serialized:
```
grahamhoyes: Can remove this

hackathon_site/event/views.py
```diff
@@ -12,6 +12,12 @@
 from registration.forms import JoinTeamForm
 from registration.models import Team
 
+from django.views.generic import TemplateView
```
grahamhoyes: Unused import

hackathon_site/event/views.py
```diff
@@ -12,6 +12,12 @@
...
+from rest_framework import generics, mixins
+
+from event.models import User, Team
```
grahamhoyes: User unused

hackathon_site/event/views.py
```diff
@@ -12,6 +12,12 @@
...
+from event.serializers import UserSerializer, TeamSerializer
```
grahamhoyes: `UserSerializer` unused

hackathon_site/event/serializers.py
grahamhoyes: Actually, disregard this for now. I'll create a separate ticket, so that we don't get in the habit of scope creep (which I am bad for).

grahamhoyes (commented):
See comments.

I also think your version of Black might be wrong, since some of the formatting doesn't agree. Run `black --version`, and make sure you're on version `19.10b0`. The later versions change formatting habits a bit it seems.

hackathon_site/event/serializers.py
```diff
@@ -10,7 +10,20 @@ class Meta:
...
+        fields = (
+            "id",
+            "team_code",
+            "created_at",
+            "updated_at",
+        )
+
+
 class ProfileSerializer(serializers.ModelSerializer):
+    team = TeamSerializer(read_only=True)
```
grahamhoyes: I would actually recommend removing this, but leaving `team` in the fields list so it sticks around as an ID. I can't think of any case where we would need the full team to be serialized inside a profile endpoint.

hackathon_site/event/serializers.py
arshinmar: cool, is it ok if I resolve this?

grahamhoyes: Let's keep it unresolved just so it's visible for future reference

hackathon_site/event/models.py
arshinmar: Should be updated now

hackathon_site/event/serializers.py
arshinmar: donezo

arshinmar:
> See comments.
> 
> I also think your version of Black might be wrong, since some of the formatting doesn't agree. Run `black --version`, and make sure you're on version `19.10b0`. The later versions change formatting habits a bit it seems.

I'm having issues installing Black with that version. For example, "npm install black==19.10b0" or "yarn install black==19.10.0". I also tried "i" instead of install and also a decimal instead of b. Nothing seems to work.

Other than that, everything seems ok

arshinmar:
hey, so I still don’t fully understand the error

the test function that is bringing up the error is 

```
def test_join_team(self):
    """
    Once the user has applied and before they have been reviewed, they can join
    a different team
    """
    self._login()
    self._apply()

    new_team = RegistrationTeam.objects.create()
    data = {"team_code": new_team.team_code}
    response = self.client.post(self.view, data=data)
    self.assertRedirects(response, self.view)
    redirected_response = self.client.get(response.url)
    self.assertContains(redirected_response, new_team.team_code)
```

The error it brings up is the same as the one on GitHub Actions.

My issue is the error still persists after adding the import statements shown below and changing the usages of the word “Team” to “EventTeam” in event/tests.py.

```
from event.models import Profile, User
from event.models import Team as EventTeam
from hackathon_site.tests import SetupUserMixin
from registration.models import Team as RegistrationTeam
```

I didn't see any of the team imports (other than the word TeamSerializer, which we aren't using in the function above) be greyed out.

Furthermore, it seems like when I use ```new_team = RegistrationTeam.(...)``` it fails to post. But when I use ```new_team = EventTeam.(...)```, it seems like it posts, but the response doesn't have an URL.

I tried changing the event/views.py code to include a similar input ```from event.models import Team as EventTeam```, but that doesn't change anything either.

I've committed my code, if you could let me know what am I missing in my analysis, it would be much appreciated.

grahamhoyes (approved):

PR #286, opened 2021-06-17
IEEE-18 POST: /api/hardware/orders/

EdwardJTL:
## Overview

- Resolves IEEE-18
- Update order status choices; remove `cart` and add `cancelled`
- Create permission class for user with event profile
- Add logger config for Hardware
- Add util to create event profile in SetupUserMixin
- Update OrderListView
  - Use get_permissions to enforce different access constraints for LIST and POST
  - Use get_serializer_class to parse data differently for LIST and POST
- Implement post for creating order
  - Add post action that uses two serializers to perform POST
    - the action is atomic
  - Create serializer for parsing response and creating order object
    - Serializer expects hardware ids and quantities in array, and merges requests for the same hardware
    - Serializer verifies that including the team's past orders that are not returned or cancelled, the new requested quantities will not violate any hardware limit or category limit; else, it throws a 400 bad request
    - Serializer verifies that the requested order is not empty - empty array or 0 total requested quantities; else it throws a 400 bad request
    - Serializer creates new order, with as many order items as it can satisfy with the hardwares' remaining quantities.
  - Create serializer for forming the response
    - Order id is only none when no new order items are created
    - Response has "hardware" entry with array of hardware ids and quantities created
    - Response has "errors" entry with array of hardware ids and the error messages if the order could not be fulfilled in its entirety
- Reordered some import statements 

## Unit Tests Created

- Testing the post action for the aforementioned behaviours, mainly to verify access control, limits, and remaining quantities.

## Steps to QA

- Run tests
- Create entries in dev data bases however you like, and run the request through swagger


hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +50,91 @@ def get_unique_hardware_count(obj: Category):
...
+    class OrderCreateHardwareSerializer(serializers.Serializer):
+        id = serializers.PrimaryKeyRelatedField(
+            queryset=Hardware.objects.all(), many=False, required=True
+        )
+        quantity = serializers.IntegerField(required=True)
+
+    team_id = serializers.PrimaryKeyRelatedField(
+        queryset=TeamEvent.objects.all(), many=False, required=True
+    )
+    hardware = OrderCreateHardwareSerializer(many=True, required=True)
```
EdwardJTL, line 57: we could change this to a dictionary instead of a nested serializer
nested serializer:
- 🟢dont need to validate hardware pk ourselves
- 🔴there can be multiple entries for the same hardware. additional logic is needed to merge them; or we can not merge them and process them regardless if you are okay with that)
dictionary:
- 🟢all entries have unique keys, wont have multiple entires for the same hardware
- 🔴need to validate the pk's ourselves in the serializer

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +50,91 @@ def get_unique_hardware_count(obj: Category):
...
+                OrderItem.objects.create(
+                    order=new_order, hardware=hardware,
+                )
+
+        return new_order, unfulfilled_hardware_requests
+
+
+class OrderCreateResponseSerializer(serializers.Serializer):
+    def to_representation(self, instance):
+        response = {"order_id": instance.id}
```
EdwardJTL: i think this might be a good thing to return? idk

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +50,91 @@ def get_unique_hardware_count(obj: Category):
...
+        fulfilled_order_items = instance.items
+        fulfilled_count = Counter(
+            map(lambda item: item.hardware_id, fulfilled_order_items.all())
+        )
+        response["hardware"] = dict(fulfilled_count)
+        unfulfilled = self.context["unfulfilled"]
+        response["errors"] = {
+            hardware.id: count for (hardware, count) in unfulfilled.items()
+        }
```
EdwardJTL: i thought it is enough for the server to return the pk and unfulfilled amount for each hardware. The front end can figure out how to formate the response to present to the user (unless it is difficult in react to do that...?)

hackathon_site/hardware/views.py
```diff
@@ -31,9 +35,38 @@ def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
 
 
-class OrderListView(mixins.ListModelMixin, generics.GenericAPIView):
+class OrderViewSet(
```
EdwardJTL: had to change this to a viewset to specify different serializer for each method

hackathon_site/hardware/views.py
```diff
@@ -31,9 +35,38 @@ def get(self, request, *args, **kwargs):
...
+    @transaction.atomic
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        new_order, unfulfilled_hardware_requests = serializer.save()
+        if new_order is None:
+            return Response(
+                {"details": "Cannot be fulfilled"},
+                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            )
```
EdwardJTL: this wasnt in the ticket
I'm not creating an order object when the entire order cannot be fulfilled, in which case i think we shouldn't return 201

hackathon_site/hardware/api_urls.py
```diff
@@ -1,16 +1,20 @@
-from django.urls import path, re_path
+from django.urls import path, re_path, include
 from hardware import views
+from rest_framework import routers
 
 
 app_name = "hardware"
 
+router = routers.DefaultRouter()
+router.register(r"orders", views.OrderViewSet, basename="order")
```
grahamhoyes: Is this absolutely necessary because of the view set? If so, might as well convert the rest of the url patterns here into the router so we stay consistent.

hackathon_site/hardware/serializers.py
```diff
@@ -1,5 +1,9 @@
+from collections import Counter
+import functools
+import itertools
 from rest_framework import serializers
```
grahamhoyes, line 5: Minor, put a blank line between between stdlib imports and module imports (above this line), and between module imports and local imports (below this line)

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +50,91 @@ def get_unique_hardware_count(obj: Category):
...
+
+
+class OrderCreateSerializer(serializers.Serializer):
+    class OrderCreateHardwareSerializer(serializers.Serializer):
+        id = serializers.PrimaryKeyRelatedField(
+            queryset=Hardware.objects.all(), many=False, required=True
+        )
+        quantity = serializers.IntegerField(required=True)
+
+    team_id = serializers.PrimaryKeyRelatedField(
```
grahamhoyes: This doesn't need to be on the serializer imo. Since they have to be logged in to access the API, you can use `self.request.user.profile.team.id` in the view.

Cc @albogdan - do we think the tech team will ever be using the API to create orders for other teams? If not, this field can go. If the field stays, then we'll want a check that a regular user can't create an order except for the team they are on.

grahamhoyes: Hmm now why did I put it in the description in the jira ticket.... What was past Graham thinking?

hackathon_site/hardware/serializers.py
grahamhoyes: - In the jira ticket, it is a dictionary mapping hardware id to quantity
- Even if there are multiple hardware entries with the current list of dicts, that shouldn't be a problem - since we have to add N hardware objects to the order anyway, having hardware appear twice with N and M quantities just means we add N + M. Though I'm guessing that makes validation more annoying, hence why you use the Counter
- All entries having unique keys - technically, json doesn't enforce this... I'm not sure what django would actually do if you posted duplicate keys, probably just keep the last one
- I like not having to validate things ourselves :)

Tldr - I defer to you. Either list of dicts, or a straight dict is fine with me.

grahamhoyes: @albogdan thoughts?

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +50,91 @@ def get_unique_hardware_count(obj: Category):
...
+    hardware = OrderCreateHardwareSerializer(many=True, required=True)
+
+    @staticmethod
+    def merge_requests(hardware_requests):
+        return functools.reduce(
+            lambda x, y: x + y,
+            [Counter({e["id"]: e["quantity"]}) for e in hardware_requests],
+        )
+
+    def create(self, validated_data):
```
grahamhoyes: I think you can also do a `validate` method, which can creates `validated_data`. Might be a bit cleaner to move some of this logic out of there.

Thinking a bit more about that, throwing a validation error in the validate method will return a 400 from the API. We decided if there isn't enough hardware remaining we want to go ahead and place the order anyway, but how about these cases:
- Constraints weren't met
- Hardware ID doesn't exist

Should we actually abort the order for those? If so, that can go in the `validate` method. If not, disregard me .

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +50,91 @@ def get_unique_hardware_count(obj: Category):
...
+            (
+                map(
+                    lambda item: item.hardware,
+                    relevant_past_order_items.filter(
+                        order__team=validated_data["team_id"],
+                    ).all(),
+                )
+            )
+        )
+        all_past_hardware_counts = Counter(
```
grahamhoyes: All of this is to get quantity remaining, right? @YuYing-Liang is working on a queryset annotation that can compute it in the database, which should be a bit cleaner to work with here.

@YuYing-Liang - For yours, instead of putting it on the view queryset, we can actually add it directly to the Hardware model with a custom manager. I'll send you those details. @EdwardJTL - hang tight until that's in place. Benefit to you is less stuff for you to test, and it should be faster.

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +50,91 @@ def get_unique_hardware_count(obj: Category):
...
+            for _ in itertools.repeat(None, min(allowed_quantity, requested_quantity)):
+                OrderItem.objects.create(
+                    order=new_order, hardware=hardware,
+                )
+
+        return new_order, unfulfilled_hardware_requests
+
+
+class OrderCreateResponseSerializer(serializers.Serializer):
+    def to_representation(self, instance):
```
grahamhoyes: The downside of using `to_representation` is that we don't get proper documentation of it on swagger and such.

How annoying do you think it would be for you to create a serializer that actually uses serializer fields? We can then annotate the view with a drf_yasg decorator so that swagger knows it uses a different serializer for the response

hackathon_site/hardware/serializers.py
grahamhoyes: I would do that on the backend instead of the frontend, since that's what most of our other APIs do. Having the severity key also lets us distinguish between warnings (eg ran out of some hardware) vs errors (invalid hardware id, constraints not met)

hackathon_site/hardware/views.py
grahamhoyes: Shouldn't need to, you can define `get_serializer_class` on a generic API view as well. You just key your dictionary on `self.request.method`, which will be `GET`, `POST`, etc. I think all the view set does here is let you call the method `create` instead of `post` which would call `self.create`, defined by CreateModelMixin.

I think its fine to convert this back to a generic view, and remove the router from the urls file.

hackathon_site/hardware/views.py
grahamhoyes: I agree we shouldn't return a 201, but it probably shouldn't be a 500 either. Can we get some more details? If it was something the user did wrong, it should be a 400.

grahamhoyes (commented):
A couple things to look into, I'll follow up where necessary. Great work though!

@albogdan you should take a look too about some of the design decisions

hackathon_site/hardware/views.py
```diff
@@ -31,9 +35,38 @@ def get(self, request, *args, **kwargs):
...
+    serializer_action_classes = {
+        "list": OrderListSerializer,
+        "create": OrderCreateSerializer,
+    }
+
+    def get_serializer_class(self):
+        try:
+            return self.serializer_action_classes[self.action]
+        except (KeyError, AttributeError):
+            return super(OrderViewSet, self).get_serializer_class()
```
grahamhoyes: Can just do `super().get_serializer_class()` now. We don't support old Python anyway.

hackathon_site/hardware/serializers.py
grahamhoyes: Good point actually - we should put that in the response, so it can be fetched from the frontend (though we could also just make the frontend refresh all orders)

hackathon_site/hardware/serializers.py
EdwardJTL: Yea I thought it would be in the user profile too loool

hackathon_site/hardware/serializers.py
EdwardJTL: I think we should def catch hardware doesn't exist and throw, cuz that's just a bad request from the client. Rn the serializer already does that for us

I'm not sure what "constraints" you are referring to? If it's the allowed quantities constraints, the design was to create order for what we can and tell the user whatever we can't fulfill. So I don't think we can just throw there. 

Another thing is idk if using validate can give me the counts for things that cannot be unfulfilled.

hackathon_site/hardware/serializers.py
EdwardJTL: I didn't know about that, I'll look into it

hackathon_site/hardware/views.py
EdwardJTL: Yea I wasn't sure about the code. What details do you want to include?

grahamhoyes: Any details you think should be passed to the frontend for the user to fix.

If it's not enough quantity available, that should just go in the error key of the response and keep the status as 201. Anything that's the user's fault that will stop us from fulfilling an order (constraints not me, hardware ID invalid) should throw a 400 (look up DRF `ValidationError`). It's fine if you validate those one at a time and just throw the first error you hit.

hackathon_site/hardware/serializers.py
grahamhoyes: We have two types of constraints:
- Each hardware has a `max_per_team` field. So check how many of that hardware the team has out (order items excluding orders that have been cancelled, and where `part_returned_health` is not null indicating it was returned). If how many they requested would put them over the limit, throw a 400 (`ValidationError`) with a suitable error message we can display
- Same with category, there's a `max_per_team` field. The query is probably a bit more annoying to craft, but same logic as above.

The only time we should fulfill a partial order is when some of the hardware got checked out and there isn't enough quantity remaining, because that isn't the user's fault. Them requesting more than they should be allowed is their fault (but could happen eg due to two team members submitted orders one after another without refreshing their pages).

hackathon_site/hardware/serializers.py
grahamhoyes: Here's the decorator I was talking about - [`@swagger_auto_schema`](https://drf-yasg.readthedocs.io/en/stable/custom_spec.html#the-swagger-auto-schema-decorator).

You would decorate the post method of the view something like

```
@swagger_auto_schema=(responses={201: OrderCreateResponseSerializer})
def post(...)
    ...
```

hackathon_site/hardware/api_urls.py
EdwardJTL: reverted back to APIView

hackathon_site/hardware/serializers.py
EdwardJTL: added

hackathon_site/hardware/serializers.py
EdwardJTL: Added object level validation in serializer

hackathon_site/hardware/api_urls.py
```diff
@@ -6,7 +6,7 @@
 
 urlpatterns = [
     path("hardware/", views.HardwareListView.as_view(), name="hardware-list",),
-    path("orders/", views.OrderListView.as_view(), name="order-list",),
```
grahamhoyes, line 9: Should still be called `OrderListView` to stick with convention. Contrast than with `OrderDetailView`, which doesn't exist yet, but would be the endpoint for a particural order (route `orders/<pk:int>/`). POST is executed on the list view, PUT/PATCH and DELETE are executed on the detail view.

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +54,128 @@ def get_unique_hardware_count(obj: Category):
...
+        )
+
+    # check that the requests are within per-team constraints
+    def validate(self, data):
+        requested_hardware = self.merge_requests(hardware_requests=data["hardware"])
+        hardware_query = (
+            Hardware.objects.filter(
+                id__in=[hardware.id for hardware in requested_hardware.keys()],
+            )
+            .all()
```
grahamhoyes, line 86: Minor, I would put a `.distinct()` after the `.all()`, just in case the same ID is submitted mulitple times

grahamhoyes: Actually nvm, looks like that gets fixed by your `merge_requests` function. Correct me if I'm wrong though.

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +54,128 @@ def get_unique_hardware_count(obj: Category):
...
+            .prefetch_related("categories", "order_items")
+        )
+        team_unreturned_orders = hardware_query.annotate(
+            past_order_count=Count(
```
grahamhoyes, line 90: add `, distinct=True` to the count, just in case. Sometimes when you have a bunch of filter queries the joins will end up duplicating things behind the scene.

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +54,128 @@ def get_unique_hardware_count(obj: Category):
...
+                & ~Q(order_items__order__status="Cancelled")
+                & Q(order_items__order__team=self.context["request"].user.profile.team),
+            )
+        )
+        category_counts = dict()
+        error_messages = []
+        for (hardware, requested_quantity) in requested_hardware.items():
+            try:
+                team_hardware = team_unreturned_orders.get(id=hardware.id)
+            except ObjectDoesNotExist:
```
grahamhoyes: The only time this would happens is if the id they submitted is invalid, right? I think in that case we should just throw a validation error since its a bad request, rather than swallowing it and moving on.

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +54,128 @@ def get_unique_hardware_count(obj: Category):
...
+            past_order_count=Count(
+                "order_items",
+                filter=Q(order_items__part_returned_health__isnull=True)
+                & ~Q(order_items__order__status="Cancelled")
+                & Q(order_items__order__team=self.context["request"].user.profile.team),
+            )
+        )
+        category_counts = dict()
+        error_messages = []
+        for (hardware, requested_quantity) in requested_hardware.items():
```
grahamhoyes, line 100: Unless I'm crazy, `requested_hardware` is effectively a dict mapping id: quantity. So `hardware` in here is just the id, not something you can do `hardware.id`, `hardware.name` etc on

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +54,128 @@ def get_unique_hardware_count(obj: Category):
...
+    @staticmethod
+    def merge_requests(hardware_requests):
+        return functools.reduce(
+            lambda x, y: x + y,
+            [Counter({e["id"]: e["quantity"]}) for e in hardware_requests],
+        )
+
+    # check that the requests are within per-team constraints
+    def validate(self, data):
+        requested_hardware = self.merge_requests(hardware_requests=data["hardware"])
```
grahamhoyes, line 79: This is just a counter (dict) mapping id -> quantity, right? If so, the rest of your logic assumes the keys are objects not ints

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +54,128 @@ def get_unique_hardware_count(obj: Category):
...
+    def create(self, validated_data):
+        requested_hardware = self.merge_requests(
+            hardware_requests=validated_data["hardware"]
+        )
+        relevant_past_order_items = (
+            OrderItem.objects.filter(hardware__in=requested_hardware.keys(),)
+            .exclude(order__status="Cancelled")
+            .select_related("hardware")
+        )
+        team_past_hardware_counts = Counter(
```
grahamhoyes: Can you convert this to pure database logic like you did above? My tiny python brain breaks as soon as map is involved (and doing it in the database is more efficient)

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +54,128 @@ def get_unique_hardware_count(obj: Category):
...
+            )
+        )
+        all_past_hardware_counts = Counter(
+            (map(lambda item: item.hardware, relevant_past_order_items.all(),))
+        )
+        new_order = None
+        unfulfilled_hardware_requests = dict()
+        for (hardware, requested_quantity) in requested_hardware.items():
+            allowed_quantity = min(
+                hardware.max_per_team - team_past_hardware_counts[hardware],
```
grahamhoyes: Don't need this part because of the validate function now

hackathon_site/hardware/serializers.py
```diff
@@ -46,9 +54,128 @@ def get_unique_hardware_count(obj: Category):
...
+                continue
+            if new_order is None:
+                new_order = Order.objects.create(
+                    team=validated_data["team_id"], status="Submitted"
+                )
+            if allowed_quantity < requested_quantity:
+                unfulfilled_hardware_requests[hardware] = (
+                    requested_quantity - allowed_quantity
+                )
+            for _ in itertools.repeat(None, min(allowed_quantity, requested_quantity)):
```
grahamhoyes: I'll let most of the other itertools/functools stuff slide, but this is Python.

```python
for _ in range(min(allowed_quantity, requested_quantity)):
    ...
```

Even better, for efficiency:

```python
order_items = [OrderItem(order=new_order, hardware=hardware) for _ in range(min(allowed_quantity, requested_quantity))]
OrderItem.objects.bulk_create(order_items)
```

If you wanted to be even more efficient, keep a list of order items to create outside of this loop, then insert them all in a single database query at the end with `OrderItem.objects.bulk_create()`.

grahamhoyes: On second thought, definitely do that last thing of creating them all with a single bulk create. This will significantly cut down on the number of database queries made when everyone submits orders at once.

grahamhoyes (commented):
I still have a bit more to look at in the response serializer and the views/tests, but here's some comments

hackathon_site/hardware/serializers.py
grahamhoyes: Because of DRF magic, the `id` key is actually a full hardware object

hackathon_site/hardware/serializers.py
EdwardJTL: it honestly shouldn't ever happen, since the hardware id is validated through the serializer field. but yea, not swallowing the error is a good idea.

hackathon_site/hardware/views.py
EdwardJTL: done

hackathon_site/hardware/serializers.py
EdwardJTL: using quantity remaining now

hackathon_site/hardware/serializers.py
EdwardJTL: done

hackathon_site/hardware/api_urls.py
EdwardJTL: yup reverted

hackathon_site/hardware/serializers.py
EdwardJTL: added

hackathon_site/hardware/serializers.py
EdwardJTL: updated

hackathon_site/hardware/serializers.py
EdwardJTL: updated

hackathon_site/hardware/serializers.py
EdwardJTL: yup done

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+            raise serializers.ValidationError(error_message)
+        return data
+
+    def create(self, validated_data):
+        # validated data should already satisfy all constraints
+        requested_hardware = self.merge_requests(
+            hardware_requests=validated_data["hardware"]
+        )
+        new_order = None
+        response_data = {"order_id": -1, "hardware": [], "errors": []}
```
EdwardJTL: so if no new order is created (when all requested quantities or remaining quantities are 0), it would return -1 as order id. idk if thats gonna bother anyone

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+        # "id": <Hardware Object>, and
+        # "quantity": <Int>
+        return functools.reduce(
+            lambda x, y: x + y,
+            [Counter({e["id"]: e["quantity"]}) for e in hardware_requests],
+        )
+
+    # check that the requests are within per-team constraints
+    def validate(self, data):
+        # requested_hardware is a Counter where the keys are <Hardware Object>'s and values are <Int>'s
```
grahamhoyes: Minor, wrap this comment so that line length is < 89 chars to match black

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+                & ~Q(order_items__order__status="Cancelled")
+                & Q(order_items__order__team=self.context["request"].user.profile.team),
+                distinct=True,
+            )
+        )
+        category_counts = dict()
+        error_messages = []
+        for (hardware, requested_quantity) in requested_hardware.items():
+            team_hardware = team_unreturned_orders.get(id=hardware.id)
+            team_hardware_count = getattr(team_hardware, "past_order_count", 0)
```
grahamhoyes, line 102: Is there any way this attribute will end up not being there?

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+            if hardware.max_per_team < (team_hardware_count + requested_quantity):
```
grahamhoyes: Super minor, I would flip the order of this comparison just to make it a bit easier to read

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+                error_messages.append("Hardware {} limit reached".format(hardware.name))
+            for category in hardware.categories.all():
+                category_counts[category] = (
+                    category_counts.get(category, 0)
+                    + team_hardware_count
+                    + requested_quantity
+                )
+        for (category, count) in category_counts.items():
+            if category.max_per_team < count:
```
grahamhoyes: Ditto about flipping the order of comparison just for readability

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+            if new_order is None:
+                new_order = Order.objects.create(
+                    team=self.context["request"].user.profile.team, status="Submitted"
+                )
+                response_data["order_id"] = new_order.id
+            order_items = [
+                OrderItem(order=new_order, hardware=hardware)
+                for _ in range(num_order_items)
+            ]
+            OrderItem.objects.bulk_create(order_items)
```
grahamhoyes: Can pull this outside of the for loop, by having a single list defined outside the loop that all the order items get added to, then calling `bulk_create` afterwards. Don't think that will cause any logic issues, and will help with performance if they request a lot of hardware (given that this view locks the database)

hackathon_site/hardware/serializers.py
grahamhoyes: I would say that's not great, since we need to include catches for that on the frontend.

The only way this could happen is if `requested_hardware` is empty, right? I'd say we should either throw a 400 in that case (submitting an order with no items sounds like a bad request to me). 

If you don't think that's appropriate and still want to return 200, then make order_id default to null since that's a bit cleaner to check for on the frontend. But I think 400 (raise a validation error) is appropriate.

hackathon_site/hardware/views.py
```diff
@@ -31,9 +36,31 @@ def get(self, request, *args, **kwargs):
...
+    @transaction.atomic
+    @swagger_auto_schema(responses={201: OrderCreateResponseSerializer})
+    def post(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        create_response = serializer.save()
+        response_serializer = OrderCreateResponseSerializer(data=create_response)
+        response_serializer.is_valid()
+        response_data = response_serializer.data
+        headers = self.get_success_headers(response_data)
```
grahamhoyes: I don't think this is actually going to do anything, since the `OrderCreateResponseSerializer` doesn't have a URL field (unless it magically derives one from order_id?)

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+            response_data["hardware"].append({hardware.id: num_order_items})
```
grahamhoyes: If num_order_items is 0, we'll still return that as part of the hardware key. Do you think we should be doing that, or only returning things that were actually fulfilled? I could go either way on this one, cc @albogdan

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+            hardware_requests=validated_data["hardware"]
+        )
+        new_order = None
+        response_data = {"order_id": -1, "hardware": [], "errors": []}
+        for (hardware, requested_quantity) in requested_hardware.items():
+            num_order_items = min(hardware.quantity_remaining, requested_quantity)
+            if num_order_items <= 0:
+                response_data["hardware"].append({hardware.id: 0})
+                response_data["errors"].append(
+                    {hardware.id: "There are no {} available".format(hardware.name)}
```
grahamhoyes: Minor, can make this `There are no {}s available` for grammar in most cases

hackathon_site/hardware/serializers.py
grahamhoyes: Ah, it can also end up not making an order if `num_order_items <= 0` for all items in the request (ie, requesting only items that are now sold out). 

Could we do it so that:
- If no hardware was submitted, 400
- If everything they requested was checked out, the order id is None instead of -1. Just set `allow_null=True` on your response serializer I think

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+            )
+            .all()
+            .prefetch_related("categories", "order_items")
+        )
+        team_unreturned_orders = hardware_query.annotate(
+            past_order_count=Count(
+                "order_items",
+                filter=Q(order_items__part_returned_health__isnull=True)
+                & ~Q(order_items__order__status="Cancelled")
+                & Q(order_items__order__team=self.context["request"].user.profile.team),
```
grahamhoyes: Edge case: submitting a request without a profile will cause a 500. I would add a check in the serializer above here to throw a ValidationError on the off chance we use this serializer elsewhere, but I think you should also add a [`@user_passes_test`](https://docs.djangoproject.com/en/3.2/topics/auth/default/#django.contrib.auth.decorators.user_passes_test) decorator to the view that checks the user has a profile

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+
+    hardware = OrderCreateHardwareSerializer(many=True, required=True)
+
+    @staticmethod
+    def merge_requests(hardware_requests):
+        # hardware_requests is a list where each element is an OrderedDict
+        # with key value pairs
+        # "id": <Hardware Object>, and
+        # "quantity": <Int>
+        return functools.reduce(
```
grahamhoyes, line 65: This will actually throw an error if hardware_requests is empty:

```
{
  "hardware": []
}
```

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+                response_data["order_id"] = new_order.id
+            order_items = [
+                OrderItem(order=new_order, hardware=hardware)
+                for _ in range(num_order_items)
+            ]
+            OrderItem.objects.bulk_create(order_items)
+            response_data["hardware"].append({hardware.id: num_order_items})
+            if num_order_items != requested_quantity:
+                response_data["errors"].append(
+                    {
```
grahamhoyes, line 154: Courtesy of @albogdan, this format doesn't match your serializer definition. This is `id: message`, but it should be `{"hardware_id": id, "error": message}`. Alex and I both prefer the format you have defined in the serializer.

Also that means that your view isn't properly validating the response serializer, because this should be failing

hackathon_site/hardware/views.py
```diff
@@ -31,9 +36,31 @@ def get(self, request, *args, **kwargs):
...
         return self.list(request, *args, **kwargs)
+
+    @transaction.atomic
+    @swagger_auto_schema(responses={201: OrderCreateResponseSerializer})
+    def post(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        create_response = serializer.save()
+        response_serializer = OrderCreateResponseSerializer(data=create_response)
+        response_serializer.is_valid()
```
grahamhoyes: This should be failing methinks, because of the bad response format for errors currently used.

Also technically, this should raise a 500: Internal Server Error if it fails. Not exactly sure how to do that though.

hackathon_site/hardware/serializers.py
grahamhoyes: Same thing, format doesn't match the serializer

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +38,131 @@ def get_unique_hardware_count(obj: Category):
...
+        # validated data should already satisfy all constraints
+        requested_hardware = self.merge_requests(
+            hardware_requests=validated_data["hardware"]
+        )
+        new_order = None
+        response_data = {"order_id": -1, "hardware": [], "errors": []}
+        for (hardware, requested_quantity) in requested_hardware.items():
+            num_order_items = min(hardware.quantity_remaining, requested_quantity)
+            if num_order_items <= 0:
+                response_data["hardware"].append({hardware.id: 0})
```
grahamhoyes: Also doesn't match the serializer

hackathon_site/hardware/serializers.py
grahamhoyes: Also wrong format for serializer

hackathon_site/hardware/views.py
EdwardJTL: i didn't know what this does i just copied it from the generic implementation lool.
removed :)

hackathon_site/hardware/serializers.py
EdwardJTL: I put it in the serializer. 
I saw that there is a mixin for when using class based views, so i used that instead. it returns a 403 forbidden response tho

EdwardJTL: i couldnt get the function decorator to work on the action methods.
The test cases we made for list didn't have a profile for the user, so i made an override for get_test_func, and we just allow get request to work even when the user doenst have a profile. But i couldn't use the same dictionary approach we did for the serializer since self is not available

hackathon_site/hardware/serializers.py
EdwardJTL: was missing an initial value, fixed now

hackathon_site/hardware/serializers.py
EdwardJTL: done :)

hackathon_site/hardware/serializers.py
EdwardJTL: done :)

hackathon_site/hardware/serializers.py
EdwardJTL: done :)

hackathon_site/hardware/serializers.py
EdwardJTL: done :)

hackathon_site/hardware/serializers.py
EdwardJTL: done :)

hackathon_site/hardware/serializers.py
EdwardJTL: done :)

hackathon_site/hardware/serializers.py
EdwardJTL: no i don't think so? the attribute comes from the query annotate and count thing, and it seems to be always there.
 I think i did this because `team_hardware.past_order_count` doesnt get recognized by PyCharm and there was no auto-completion 🤷‍♂️

grahamhoyes (commented):
Forgot to hit submit on this whoops

EdwardJTL:
reopening this to clear out the comments lol

PR #287, opened 2021-06-19
IEEE-2: convert quantity remaining to queryset

YuYing-Liang:
## Overview
- added quantity remaining queryset to custom hardware manager


hackathon_site/hardware/models.py
```diff
@@ -29,6 +29,7 @@ class Meta:
     max_per_team = models.IntegerField(null=True)
     picture = models.ImageField(upload_to="uploads/hardware/pictures/", null=False)
     categories = models.ManyToManyField(Category)
+    objects = AnnotatedHardwareManager()
```
grahamhoyes: I would put this at the top, above the `class Meta` and with a blank line between them. From a code organization perspective, it at least shouldn't be mixed in between database fields.

hackathon_site/hardware/models.py
```diff
@@ -95,3 +96,22 @@ class Incident(models.Model):
 
     def __str__(self):
         return self.id
+
+
+class AnnotatedHardwareManager(models.Manager):
```
grahamhoyes: This needs to be defined above the Hardware model (I would put it right above it), python can't reference things defined later in the file like JavaScript can

hackathon_site/hardware/models.py
```diff
@@ -95,3 +96,22 @@ class Incident(models.Model):
...
+    def get_queryset(self):
+        return super()
```
grahamhoyes: This whole thing needs to be wrapped in brackets, not valid python syntax as it is

hackathon_site/hardware/models.py
```diff
@@ -95,3 +96,22 @@ class Incident(models.Model):
...
+            .get_queryset()
+            .annotate(
+                quantity_checked_out=Count(
```
grahamhoyes: `Count`, `Q`, `F` etc are not imported

hackathon_site/hardware/serializers.py
```diff
@@ -1,10 +1,10 @@
+from django.db.models import Count, F, Q
```
grahamhoyes: Move this import over to models.py

hackathon_site/hardware/serializers.py
```diff
@@ -19,6 +19,7 @@ class Meta:
             "picture",
             "categories",
             "quantity_remaining",
+            "order_items_count"
```
grahamhoyes: Was this just for testing?

hackathon_site/hardware/views.py
```diff
@@ -4,7 +4,7 @@
 
 
 class HardwareListView(mixins.ListModelMixin, generics.GenericAPIView):
-    queryset = Hardware.objects.all()
+    queryset = Hardware.objects.get_queryset()
```
grahamhoyes: This can stay as `Hardware.objects.all()` to stick with convention. If you check out the [source code](https://github.com/django/django/blob/main/django/db/models/manager.py#L149) for the `all` method, it just calls `get_queryset()`, which you extended in your custom manager.

hackathon_site/requirements.txt
```diff
@@ -10,6 +10,7 @@ dj-rest-auth==1.0.6
 Django==3.1.8
 django-cors-headers==3.3.0
 django-debug-toolbar==2.2.1
+django-filter==2.4.0
```
grahamhoyes: I think these changes were already merged into develop in a separate PR? 

Looking at the git log, it looks like you branched this branch off of your local ieee-2 branch. That's generally not a good idea, since it will end up making changes appear again in the PR (the technical reason being that we squash and merge PRs which creates a new commit, so the old one you branched off isn't recognized).

Technically the way to fix this is to rebase, but I think pulling and merging develop in will fix this (if not, we can delete the offending commit from this branch and force push). In the future, make sure you always branch off develop.

hackathon_site/hardware/models.py
```diff
@@ -95,3 +96,22 @@ class Incident(models.Model):
...
+                    "order_items",
+                    filter=(
+                        Q(order_items__part_returned_health__isnull=True)
+                        & ~Q(order_items__order__status="Cart")
```
grahamhoyes: Once you merge in the develop branch, the "Cart" status is now gone, but we added a new "Cancelled" status which should be excluded here. So this can just be swapped with "Cancelled" instead

hackathon_site/hardware/admin.py
```diff
@@ -1,8 +1,9 @@
 from django.contrib import admin
-from .models import Hardware, Category, Order, Incident
+from .models import Hardware, Category, Order, Incident, OrderItem
 
 # Register your models here.
 admin.site.register(Hardware)
 admin.site.register(Category)
 admin.site.register(Order)
 admin.site.register(Incident)
+admin.site.register(OrderItem)
```
grahamhoyes: Minor, add a newline at the end of the file

hackathon_site/hardware/serializers.py
```diff
@@ -19,6 +19,7 @@ class Meta:
...
         )
 
     @staticmethod
```
grahamhoyes: Make sure you remove the `get_quantity_remaining` method, since you've now replaced it

grahamhoyes (commented):
This is a good start! A few things:

First, make sure you're using a properly configured IDE, or at least make sure to run the code before you open a PR. That will catch the syntax errors and missing imports I commented below.

Second, lets work on adding some unit tests to this. We have an implementation of the annotation, but are we sure it works? In `hardware/tests.py`, I would recommend creating a new test class like `HardwareQuantityRemainingTestCase(TestCase)` that just tests the behaviour of this new annotation. You can take a look at the current hardware serializer tests in the same file for some insparation, but consider things like:
- What happens when there are no order items?
- What happens when some order items are returned (are they properly excluded from the count)? How about when none are returned (are they included)?
- What happens when orders have been cancelled (are they properly excluded from the quantity checked out)?

Let me know if you have any questions, some things here may need a bit of deeper knowledge of Django / DRF (which you will soon possess)!

hackathon_site/hardware/models.py
grahamhoyes: OK actually, there's a circular dependency between this and #286. Just leave as Cart for now and write tests accordingly, but perhaps leave a `# TODO:` in the code so we remember to fix it later.

hackathon_site/hardware/models.py
```diff
@@ -1,6 +1,6 @@
 from django.db import models
 from event.models import Team as TeamEvent
-
+from django.db.models import Count, F, Q
```
grahamhoyes, line 2: For organization purposes, we group imports together into 3 categories, each separated by a new line:
- Python stdlib imports
- 3rd party packages (django, etc)
- Local modules

Could you tweak this section to be:
```python
from django.db import models
from django.db.models import Count, F, Q

from event.models import Team as TeamEvent
```

hackathon_site/hardware/serializers.py
```diff
@@ -3,8 +3,7 @@
 
 
 class HardwareSerializer(serializers.ModelSerializer):
-    quantity_remaining = serializers.SerializerMethodField()
-
+    quantity_remaining = serializers.IntegerField()
```
grahamhoyes, line 6: Can't comment on the right line, but make sure you remove the `get_quantity_remaining` method in this serializer since it's no longer needed

hackathon_site/hardware/tests.py
```diff
@@ -124,6 +124,84 @@ def test_unique_hardware_exists(self):
         }
         self.assertEqual(expected_response, data)
 
+class HardwareQuantityRemainingTestCase(TestCase):
+    def setUp(self):
+        self.hardware = Hardware.objects.create(
```
grahamhoyes: So this is a little bit weird, but because we define the `quantity_remaining` field as an annotation on the queryset, when you create a single object (either by doing `Hardware.objects.create()` or `hardware = Hardware(...); hardware=save()`, it won't actually be set yet. We need to explicitly query using `Hardware.objects` to get it.

After all of this, you can just do:
```python
self.hardware = Hardware.objects.get(pk=self.hardware.pk)
```

(See below, you'll also have to do that in some of the test cases)

I think you'll have to do the same thing in the `setUp` method of `HardwareSerializerTestCase` as well. When you run the tests, look for an error like `AttributeError: 'Hardware' object has no attribute 'quantity_remaining'` to tell you where it needs to be.

I was looking into ways to do this with the `self.hardware.refresh_from_db()` function, but it only does concrete fields it looks like: https://github.com/django/django/blob/main/django/db/models/base.py#L643. Perhaps I'll open an issue about that with Django at some point.

hackathon_site/hardware/tests.py
```diff
@@ -124,6 +124,84 @@ def test_unique_hardware_exists(self):
         }
         self.assertEqual(expected_response, data)
 
+class HardwareQuantityRemainingTestCase(TestCase):
```
grahamhoyes: Minor, but move this so that it's below the other hardware serializer test case

hackathon_site/hardware/tests.py
```diff
@@ -124,6 +124,84 @@ def test_unique_hardware_exists(self):
...
+            "datasheet": "/datasheet/location/",
+            "quantity_available": 4,
+            "quantity_remaining": 4,
+            "notes": None,
+            "max_per_team": 1,
+            "picture": "/media/picture/location",
+        }
+
+        data = hardware_serializer.quantity_remaining
+        self.assertEqual(expected_response, data)
```
grahamhoyes: This is going to compare the `quantity_remaining` field of the serializer (you'd actually have to do `hardware_serializer.data["quantity_remaining"] for that)` with the entire `expected_response` object. Since we already test the rest of the serializer elsewhere, and this is only for testing the `quantity_remaining` field, you can just do:

```python
def test_base_case_no_order_items(self):
    hardware_serializer = HardwareSerializer(self.hardware)
    self.assertEqual(hardware_serializer.data["quantity_remaining"], 4)
```

Ditto for the rest of the tests, so I won't repeat this comment. In summary: no need for the full `expected_response` object.

hackathon_site/hardware/tests.py
```diff
@@ -124,6 +124,84 @@ def test_unique_hardware_exists(self):
...
+
+        expected_response = {
+            "id": 1,
+            "name": "name",
+            "categories": [],
+            "model_number": "model",
+            "manufacturer": "manufacturer",
+            "datasheet": "/datasheet/location/",
+            "quantity_available": 4,
+            "quantity_remaining": 4,
```
grahamhoyes: Shouldn't this be 2?

hackathon_site/hardware/tests.py
```diff
@@ -124,6 +124,84 @@ def test_unique_hardware_exists(self):
...
+            "notes": None,
+            "max_per_team": 1,
+            "picture": "/media/picture/location",
+        }
+
+        data = hardware_serializer.quantity_remaining
+        self.assertEqual(expected_response, data)
+
+    def test_some_items_none_returned(self):
+        hardware_serializer = HardwareSerializer(self.hardware)
```
grahamhoyes: This serializes the object right away, so the objects you create below won't be in the database yet. You should create them first, then you'll have to call `self.hardware = Hardware.objects.get(pk=self.hardware.pk)` again to re-run the annotation. After that, you can instantiate the serializer.

hackathon_site/hardware/tests.py
```diff
@@ -124,6 +124,84 @@ def test_unique_hardware_exists(self):
...
+            "manufacturer": "manufacturer",
+            "datasheet": "/datasheet/location/",
+            "quantity_available": 4,
+            "quantity_remaining": 3,
+            "notes": None,
+            "max_per_team": 1,
+            "picture": "/media/picture/location",
+        }
+        data = hardware_serializer.data
+        self.assertEqual(expected_response, data)
```
grahamhoyes: I think the test you deleted [here](https://github.com/ieeeuoft/hackathon-template/pull/287/commits/a639d2401e0ff25cd85d2c79cd20b459c2679029) is actually good to have. But since we don't actually have a cancelled status yet, you should:
- Update the queryset in the model manager to actually look filter on "Cart" status
- Write the test, with the order status being "Cart" (and no items returned yet)
- Name the test accordingly

The test should then pass with everything you have now. When @EdwardJTL swaps the Cart for the Cancelled status in his PR, the test will fail, and he can go update the test / queryset.

hackathon_site/hardware/models.py
```diff
@@ -15,8 +15,29 @@ class Meta:
     def __str__(self):
         return self.name
 
+class AnnotatedHardwareManager(models.Manager):
+    def get_queryset(self):
+        return (
+            super()
```
grahamhoyes, line 24: Don't forget to run black (`black .`) to fix up the formatting

hackathon_site/hardware/models.py
```diff
@@ -15,8 +15,29 @@ class Meta:
...
+                        ),
+                        distinct=True,
+                    )
+                )
+                .annotate(
+                    quantity_remaining=(F("quantity_available") - F("quantity_checked_out"))
+                )
+        )
 
 class Hardware(models.Model):
```
grahamhoyes, line 44: Ok after digging into the `refresh_from_db()` stuff, let's update the model to be this:

```python
class Hardware(models.Model):
    objects = AnnotatedHardwareManager()

    class Meta:
        verbose_name_plural = "hardware"

    class Config:
        annotated_fields = ("quantity_checked_out", "quantity_remaining")

    ...

    created_at = models.DateTimeField(auto_now_add=True, null=False)
    updated_at = models.DateTimeField(auto_now=True, null=False)

    def refresh_from_db(self, using=None, fields=None):
        super().refresh_from_db(using, fields)

        # Fetch the queryset again to populate the annotations. This is somewhat
        # inefficient since it performs the query twice, so use sparingly.
        db_instance_qs = self.__class__.objects.filter(pk=self.pk)
        db_instance = db_instance_qs.get()

        for field_name in self.Config.annotated_fields:
            setattr(self, field_name, getattr(db_instance, field_name))

    def __str__(self):
        return self.name
```

I added a `Config` class, and overwrote `refresh_from_db`. Since no matter what we'd have to do a separate query anyway to get the update fields, might as well make this method do it.

hackathon_site/hardware/tests.py
grahamhoyes: See my other comment with the updated model, you can just call `self.hardware.refresh_from_db()` with that change

hackathon_site/hardware/tests.py
grahamhoyes: Ditto about just calling `self.hardware.refresh_from_db()` with the model change after creating the order items, then instantiating the serializer

hackathon_site/hardware/tests.py
YuYing-Liang: Why do we need to pass in a pk?

hackathon_site/hardware/tests.py
YuYing-Liang: isn't it already below HardwareSerializerTestCase?

hackathon_site/hardware/tests.py
YuYing-Liang: Yes, whoops

hackathon_site/hardware/tests.py
grahamhoyes: Ignore the bit about using `Hardware.objects.get()`, just use `self.hardware.refresh_from_db()` with the model change in another one of my comments. 

However more generally when you see pk, that's an alias for whatever the primary key on the model is, which for all of our models is actually `id`.

hackathon_site/hardware/tests.py
grahamhoyes: Below as in directly below (so there's not a test for a different serializer between them). Just grouping similar pieces of code together to make finding them easier later on

grahamhoyes (approved):
Great work!

PR #288, opened 2021-06-26
IEEE-3 Hardware API Filtering

Leo6Leo:
## Overview

- Add order by fields for name and quantity remaining 

Filter for in stock (quantity remaining > 0)

Filter by category IDs (list of ID integers)

OR query for all IDs

Search filter on name: Filtering - Django REST framework 

## Unit Tests Created

- No unit test created

## Steps to QA

- I am not sure whether this can work or not, I think I need to ask during our meeting.


hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,24 @@
+from django_filters import rest_framework as filters
+from .models import Hardware
+from .serializers import HardwareSerializer
+from django_filters.fields import Lookup
```
grahamhoyes: Can you bring this up so it's below the other `django_filters` import above? The general structure of our imports is:
```
<imports from python stdlib>

<imports from 3rd party packages, like rest_framework>

<imports from local packages>
```

Note the new line between the import sections as well. I prefer imports to be alphebetized within each group as well, but I won't be that pedantic on PR reviews.

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,24 @@
+from django_filters import rest_framework as filters
+from .models import Hardware
```
grahamhoyes: Just to prevent things from breaking if we do refactors, make these absolute imports like `from hardware.models import Hardware`

hackathon_site/hardware/views.py
```diff
@@ -1,16 +1,24 @@
 from hardware.models import Hardware, Category, Order
```
grahamhoyes: Ditto about grouping imports by stdlib/3rd party/local (even though it was messed up before your changes), and absolute imports. Could you clean the import section up here?

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,24 @@
...
+
+class HardwareFilter(filters.FilterSet):
+    queryset = Hardware
+    serializer_class = HardwareSerializer
+
+    class Meta:
+        model = Hardware
+        fields = ["name", "quantity_available"]
+
+    def comma_separate_filter(self, queryset, value):
```
grahamhoyes: Minor: `comma_separated_filter`

hackathon_site/hardware/views.py
```diff
@@ -1,16 +1,24 @@
...
 )
 
+from .api_filters import HardwareFilter
+
 
 class HardwareListView(mixins.ListModelMixin, generics.GenericAPIView):
     queryset = Hardware.objects.all()
     serializer_class = HardwareSerializer
 
+    filter_backends = (filters.DjangoFilterBackend,)
```
grahamhoyes: Do get searching, we also need to add `SearchFilter` from DRF.

```python
from rest_framework.filters import SearchFilter
...
filter_backends = (filters.DjangoFilterBackend, SearchFilter)
```

hackathon_site/hardware/views.py
```diff
@@ -1,16 +1,24 @@
...
+    filterset_fields = ("name", "quantity_remaining")
+    filterset_class = HardwareFilter
+    search_fields = "name"
```
grahamhoyes: I think this has to be a list/tuple

grahamhoyes:
Looks we forgot to actually add `django_filters` to the list of installed apps. Can you add it to the `INSTALLED_APPS` array in `hackathon_site/hackathon_site/settings/__init__.py` under the rest of the 3rd party apps, so it looks like this?

```
INSTALLED_APPS = [
    ...
    "import_export",
    "django_filters",
    "dashboard",
    ...
]
```

That should get rid of the template not found error when visiting http://localhost:8000/api/hardware/hardware/

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,24 @@
...
+        value_list = value.split(",")
+        queryset = super().filter(queryset, Lookup(value_list, "in"))
```
grahamhoyes: hmm I'm not sure this is going to work, since `filter` is a method on a `Filter` subclass, not a `FilterSet`. You could create your own custom filter class, based on `CharFilter`:

```
class CommaSeparatedFilter(filters.CharFilter):
    lookup_expr = "in"
    
    def filter(self, qs, value):
        if value:
            value = [x.strip() for x in value.split(",")]
        
        return super().filter(qs, value)
```

And then use that, passing in `field_name=id` when instantiating it

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,24 @@
...
+
+        return queryset
+
+    in_stock = filters.BooleanFilter(
+        field_name="quantity_remaining", lookup_expr="gt", exclude=False
+    )
+    category_ID = filters.CharFilter(method="comma_separate_filter")
```
grahamhoyes: OK after more digging, it seems there is actually a CSV filter that's built in called `BaseInFilter` (I guess they expect you to extend it, but it works on its own).

You can change this line to:
`ids = filters.BaseInFilter(field_name="ids")`

If you ctrl+click your way into the source code, it basically just sets the lookup expr to `in`, and then passes in the comma separated list you get as-is, which is legal sql apparently (I would have thought it needs brackets, but this works fine in my testing)

hackathon_site/hardware/api_filters.py
grahamhoyes: See below, there's one built in

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,24 @@
...
+        model = Hardware
+        fields = ["name", "quantity_available"]
+
+    def comma_separate_filter(self, queryset, value):
+        value_list = value.split(",")
+        queryset = super().filter(queryset, Lookup(value_list, "in"))
+
+        return queryset
+
+    in_stock = filters.BooleanFilter(
```
grahamhoyes: On its own this doesn't quite work, since it's missing the "0" part of "> 0". This should work though:

```
    in_stock = filters.BooleanFilter(label="In stock", method="filter_in_stock")

    @staticmethod
    def filter_in_stock(queryset, _, value):
        if value is True:
            return queryset.filter(quantity_available__gt=0)
        else:
            return queryset.filter(quantity_available__lte=0)
```

You can read about filter methods [here](https://django-filter.readthedocs.io/en/stable/ref/filters.html#method). The methods take 3 parameters: `queryset`, `name` (the name of the field to filter on), and `value` (the value to use). 

For a `BooleanFilter`, the value under the hood will get converted to either `True` or `False`.  I'm ignoring the `name` parameter, since we can just hard code it to be `quantity_available`.

Slight problem for now though: we don't actually have `quantity_available` as a field yet, pending @YuYing-Liang's pr. For now, you can switch it to `quantity_available` for testing purposes.

grahamhoyes: For both the filters (`in_stock` and `ids`, can you add a `label` parameter to the filter, which will give it a name in the DRF browser? `"In Stock" and "IDs" will work. You can also add a `help_text` if you want, explaining how to use the field.

grahamhoyes (commented):
Comments left to get things up and running, other than `quantity_available` not being merged just yet

hackathon_site/hardware/views.py
Leo6Leo: Do you mean for filterset_fields has to be a list/tuple?

grahamhoyes: `search_fields = ("name",)`

hackathon_site/hardware/api_filters.py
Leo6Leo: Note: this is the build-in csv filter. 

image

hackathon_site/hardware/api_filters.py
Leo6Leo: Thank you so much Graham for your detailed explanation!!! I have learnt a lot from them! @grahamhoyes

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,25 @@
+from django_filters import rest_framework as filters
+from django_filters.fields import Lookup
```
grahamhoyes: Unused import now, can remove

hackathon_site/hackathon_site/settings/__init__.py
```diff
@@ -169,6 +170,7 @@
     "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.LimitOffsetPagination",
     "DEFAULT_PERMISSION_CLASSES": ["rest_framework.permissions.IsAuthenticated"],
     "PAGE_SIZE": 100,
+    "DEFAULT_FILTER_BACKENDS": ("django_filters.rest_framework.DjangoFilterBackend",),
```
grahamhoyes: Just for consistency with the other keys, make this a list instead of a tuple.

Normally I prefer to use tuples for stuff in views/serializers, since they are immutable. But for the settings file, in theory we might want to mutate things from the CI overrides settings. Highly unlikely we'll ever do that for this setting, so mostly just to be consistent.

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,25 @@
...
+    in_stock = filters.BooleanFilter(label="In stock?", method="filter_in_stock")
+
+    @staticmethod
+    def filter_in_stock(queryset, _, value):
+        if value is True:
+            return queryset.filter(quantity_available__gt=0)
+        else:
+            return queryset.filter(quantity_available__lte=0)
+
+    id = filters.BaseInFilter(field_name="id", label="Id(s) of the hardware item")
```
grahamhoyes: "Comma separated list of hardware IDs"

hackathon_site/hardware/views.py
```diff
@@ -1,16 +1,25 @@
...
+from hardware.api_filters import HardwareFilter
+from hardware.models import Hardware, Category, Order
+
 
 class HardwareListView(mixins.ListModelMixin, generics.GenericAPIView):
     queryset = Hardware.objects.all()
     serializer_class = HardwareSerializer
 
+    filter_backends = (filters.DjangoFilterBackend, SearchFilter)
+    filterset_fields = ("name", "quantity_remaining")
```
grahamhoyes: I don't think this is necessary

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,25 @@
...
+
+from hardware.models import Hardware
+from hardware.serializers import HardwareSerializer
+
+
+class HardwareFilter(filters.FilterSet):
+    queryset = Hardware
+    serializer_class = HardwareSerializer
+
+    class Meta:
```
grahamhoyes: We actually don't need this meta block at all.

If you look at swagger for this view (http://localhost:8000/swagger/#/hardware/hardware_hardware_list), this adds two extra fields for name and quantity available. We don't need a filter field for name (since it's covered by the search box), and we don't need a filter for quantity_available either (the in stock filter is all we need there). So this meta block can be removed methinks

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,25 @@
...
+        model = Hardware
+        fields = ["name", "quantity_available"]
+
+    in_stock = filters.BooleanFilter(label="In stock?", method="filter_in_stock")
```
grahamhoyes: Minor, can you add a `help_text` to these as well, that can just be the same as the label? Turns out it's using the label in the DRF browsable API, but is looking for the help text in swagger

grahamhoyes (commented):
Good stuff!

A few minor comments, then just needs some unit tests. The test cases you might need:
- Test that searching works (ie create two hardware with different names, test that searching returns the correct one)
- Test that the in_stock filter works (3 cases - not present, true, and false)
- Test that the IDs filter works (just one test, maybe have 3 hardware added and provide a list of 2 of them)

To test the views, you'll need to manually construct the URL. Test out the filters at http://localhost:8000/api/hardware/hardware/ in the browsable API ("filters" button) to see what the URL ends up as with different options.

Here's a test class you can add to the bottom of `test_api.py` to get you started: 

```
class HardwareListTestCase(SetupUserMixin, APITestCase):
    def setUp(self):
        super().setUp()
        # Create 3ish hardware objects
        self.hardware1 = ...

        self.view = reverse("api:hardware:hardware-list")

    def _build_filter_url(self, **kwargs):
        return (
            self.view + "?" + "&".join([f"{key}={val}" for key, val in kwargs.items()])
        )

    def test_user_not_logged_in(self):
        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)

    def test_search_by_name(self):
        self._login()

        url = self._build_filter_url(search="arduino")
        response = self.client.get(url)

        data = response.data
        expected_data = {
            # Your expected data
        }
        self.assertEqual(data, expected_data)
```

hackathon_site/hardware/api_filters.py
grahamhoyes: There's an edge case here, if you submit something that's a non-integer (eg `id=asdf`), the page will crash.

To fix that, we need to add a validator for the field, and throw a `400 Bad Request` response if the user gives bad IDs (crashes are always bad). Validators need to be set on the actual field inside the filter, but I can't quite figure out how to do that yet. Will report back

Leo6Leo: Hey Graham, I found one solution but this is not the most ideal one.

image

But I can input them in the address bar.
image

When I input characters other than numbers, the error will be thrown out.

image

@grahamhoyes

grahamhoyes:
I have figured out the integer CSV filter stuff!

```python
from django import forms
from django_filters import rest_framework as filters, widgets

class CSVInputIntegerField(forms.IntegerField):
    widget = widgets.CSVWidget


class IntegerCSVFilter(filters.BaseInFilter):
```

Then inside your `HardwareFilter`, you can do:

```python
    id = IntegerCSVFilter(
        field_name="id",
    )
```

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,244 @@ def test_hardware_get_success(self):
         data = response.json()
 
         self.assertEqual(expected_response, data["results"])
+
+
+class HardwareListTestCase(SetupUserMixin, APITestCase):
+    def setUp(self):
+        super().setUp()
+        # Create 3ish hardware objects
```
grahamhoyes: Can remove this comment, that was just me suggesting to you that you might need to create more than 3 objects (doesn't seem like you need to)

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,244 @@ def test_hardware_get_success(self):
...
+    def test_search_by_name(self):
+        self._login()
+
+        url = self._build_filter_url(search="arduino")
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        data = response.json()
+        expected_data = {
+            # Your expected data
```
grahamhoyes: Can remove this comment

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,244 @@ def test_hardware_get_success(self):
...
+
+    def test_search_by_name(self):
+        self._login()
+
+        url = self._build_filter_url(search="arduino")
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        data = response.json()
+        expected_data = {
```
grahamhoyes: Since the actual response format is tested elsewhere in the serializer tests, no need to include the full response body here (because if we change the serializer, we'd have to update this test too). 

I would change the rest of this test case to:

```python
self.assertEqual(len(data["results"]), 1)
self.assertEqual(data["results"][0]["id"], 2)
```

This lets us test the two important things: That only the relevant items are returned (which this test currently misses), and that it is returning the correct items

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,244 @@ def test_hardware_get_success(self):
...
+                    "quantity_remaining": 4,
+                },
+            ],
+        }
+        response_true = self.client.get(url_true)
+        self.assertEqual(response_true.status_code, status.HTTP_200_OK)
+        data_true = response_true.json()
+        self.assertEqual(data_true, expected_data_true)
+
+        # when in_stock is false
```
grahamhoyes: This should be a separate test case (`test_in_stock_false`). That way if one of them fails, we get a verbose error message about why

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,244 @@ def test_hardware_get_success(self):
...
+            "notes": "notes",
+            "max_per_team": 1,
+            "picture": "http://testserver/media/picture/location",
+            "categories": [],
+            "quantity_remaining": 4,
+        }
+
+        self.assertEqual(data["results"][0], expected_data)
+
+    def test_in_stock_filter(self):
```
grahamhoyes: So that you're actually sure this test verifies that `quantity_remaining` is 0 (not that `quantity_available` is 0), you should set the `quantity_available` on hardware1 to 1, then create an order item for it.

Also as in my other comment, no need to test the full response format, just what's been returned. This test could become:

```python
    def test_in_stock_true(self):
        self._login()

        OrderItem.objects.create(hardware=self.hardware1, order=self.order)

        url = self._build_filter_url(in_stock="true")
        response = self.client.get(url)

        data = response.json()
        results = data["results"]
        self.assertEqual(returned_ids, [2, 3])
```

You'll also need to add this to the bottom of `setUp()` for that to work:

```python
self.team = Team.objects.create()
self.order = Order.objects.create(status="Submitted", team=self.team)
```

Same thing for when `in_stock` is false, but in a separate test method.

If you do that, the test will actually fail! The `quantity_remaining` stuff has now been merged in to develop, so you can merge in the develop branch and update the filter method to use it instead of `quantity_available`.

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,244 @@ def test_hardware_get_success(self):
...
+            ],
+        }
+        response_false = self.client.get(url_false)
+        self.assertEqual(response_false.status_code, status.HTTP_200_OK)
+        data_false = response_false.json()
+        print(data_false)
+        self.assertEqual(data_false, expected_data_false)
+
+        # when in_stock is not present
+        url_not_present = self._build_filter_url(in_stock="")
```
grahamhoyes: Same, should be in its own test method as with the other two

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,244 @@ def test_hardware_get_success(self):
...
+        self.assertEqual(response_not_present.status_code, status.HTTP_200_OK)
+        data_not_present = response_not_present.json()
+        print(data_not_present)
+        self.assertEqual(data_not_present, expected_data_not_present)
+
+    def test_id_filter(self):
+        self._login()
+
+        url = self._build_filter_url(id="1,3")
+        expected_data = {
```
grahamhoyes: Same as my other comments. No need to test the entire response format (since that's tested by the serializer tests), just pull out the IDs and test an array of those.

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,244 @@ def test_hardware_get_success(self):
...
+                },
+            ],
+        }
+        response_not_present = self.client.get(url_not_present)
+        self.assertEqual(response_not_present.status_code, status.HTTP_200_OK)
+        data_not_present = response_not_present.json()
+        print(data_not_present)
+        self.assertEqual(data_not_present, expected_data_not_present)
+
+    def test_id_filter(self):
```
grahamhoyes, line 958: Now that we have protections against ids not being integers, you should add a test for that as well to make sure it doesn't get broken in the future!

hackathon_site/hardware/api_filters.py
```diff
@@ -0,0 +1,30 @@
...
+    in_stock = filters.BooleanFilter(
+        label="In stock?", method="filter_in_stock", help_text="In stock?"
+    )
+
+    @staticmethod
+    def filter_in_stock(queryset, _, value):
+        if value is True:
+            return queryset.filter(quantity_available__gt=0)
+        else:
+            return queryset.filter(quantity_available__lte=0)
```
grahamhoyes: You can merge in the develop branch and change these to use `quantity_remaining` instead of `quantity_available` now

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,152 @@ def test_hardware_get_success(self):
...
+    def setUp(self):
+        super().setUp()
+
+        self.hardware1 = Hardware.objects.create(
+            name="hardware1",
+            model_number="model",
+            manufacturer="manufacturer",
+            datasheet="/datasheet/location/",
+            notes="notes",
+            quantity_available=0,
```
grahamhoyes: Since you're creating `OrderItems` for this below, you should set `quantity_available=1`. 

Otherwise it's not legal to create an order item for this hardware, so there could be unintended side effects in the future.

hackathon_site/hardware/test_api.py
grahamhoyes: Might have resolved this one by accident, but this can be simplified to just tests for IDs in the response like all the ones above

hackathon_site/hardware/test_api.py
grahamhoyes: Still need a test for invalid ids, something like `test_invalid_ids` that does `url = self._build_filter_url(id="1,2,foo"). You can just check that `response.status_code`.

You can add this in:

```python
def test_id_invalid(self):
    self._login()

    url = self._build_filter_url(id="1,2,foo")
    response = self.client.get(url)
    data = response.json()

    self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
    self.assertEqual(data, {"id": ["Enter a whole number."]})
```

Leo6Leo:
Hey Graham! I saw that there are conflicts in this branch, and it cannot be merged until the conflicts have been resolved. Would you mind give me any hints on resolving the conflicts? What caused the conflicts?

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,164 @@ def test_hardware_get_success(self):
...
+        results = data["results"]
+        returned_ids = [res["id"] for res in results]
+
+        self.assertEqual(returned_ids, [1, 2, 3])
+
+    def test_id_filter(self):
+        self._login()
+
+        url = self._build_filter_url(id="1,3")
+        expected_data = {
```
grahamhoyes: Still need to delete this, now that you're no longer using it!

grahamhoyes:
> Hey Graham! I saw that there are conflicts in this branch, and it cannot be merged until the conflicts have been resolved. Would you mind give me any hints on resolving the conflicts? What caused the conflicts?

The conflicts are because of some changes that were made in a different PR ([this one](https://github.com/ieeeuoft/hackathon-template/pull/293)) that happen to be on the same part of the code.

To start resolving them locally, you'll need to update your develop branch locally and start a merge:

```bash
$ git checkout develop
$ git pull
$ git checkout IEEE-3
$ git merge develop
```

That will give you the warning that there are conflicts in `views.py` and `test_api.py` of the hardware app.

My favourite way to resolve these is using the visual merge confilct resolver in PyCharm. After doing the above, open up the project in PyCharm and from the top menu choose git > resolve conflicts. That will give you a popup with all the files that need resolving.

Start with `test_api.py` by clicking on it then clicking "Merge...". A new editor will pop up with 3 columns: the left column is your changes, the right column is the changes that have been made onto develop that conflict with yours, and the middle is the merged file.

Any lines highlighted in blue (might be different depending on your theme) are changes that aren't conflicting, and are able be merged. They could be coming from the left or right columns, depending where the change was made. Here, we see that one of the serializers was renamed on `develop`, but that isn't used by your code so it isn't actually a conflict. For now, ignore the blue rows.

image

Rows in red are ones that were changed by both branches, so git isn't sure what to accept. For us, it's easy because two test cases were added to the bottom of the file:

image

Clicking the `>>` or `<<` arrows will accept the change from that branch. Sometimes you have to carefully merge blocks of code together (modifying the result in the middle column if necessary), but in this case it's easy - since both were additions that don't touch each other, just click the `>>` on the left to accept your change, then click `<<` on the right to accept the change from develop (which should insert it after your test class).

At this point, in the top right of the window it should say "3 changes. No conflicts". You could go manually accept all the blue changes that are left, or in the top left you can click the "All" button to accept all non-conflicting changes from both sides (which in this case, are really only from develop): 

image

After that you'll get a nice "All changes have been processed" message, and you can hit "Apply" in the bottom right to complete the merge.

The changes in `views.py` are a bit trickier, but the only actual conflicts (in red) are in the import statements. The green lines are ones that were inserted and don't conflict, the blue lines were changed and don't conflict, and the grey ones are lines that were deleted and don't conflict.

Merging the imports will be a bit annoying, but not hard. I would recommend clicking the merge arrows from both red sections to accept both changes (which will add some duplicate lines), then clicking that "apply all non-conflicting changes" button to merge the rest, then going to fix up the imports manually in the middle column. Remember our import rules:
- Standard library modules are first, in alphabetical order, followed by a blank line. There are none of these here, so that's fine.
- External libraries are next (`django`, `rest_framework`, `drf_yasg`, etc), again in alphabetical order and with a blank line after the block.
- Local modules (from `event`, `hardware`, etc) are last, also alphabetical.

Rearrange some things, combine imports together until there's no grey unused imports (there will be duplicate imports for `rest_framework` if you followed the steps above), and once it looks good click "Apply" to finish the merge.

After that, the pycharm conflicts window should close, and you can type "git commit" (with no other arguments) into your terminal to finish the merge, then push.

hackathon_site/hardware/test_api.py
```diff
@@ -194,3 +194,164 @@ def test_hardware_get_success(self):
         data = response.json()
 
         self.assertEqual(expected_response, data["results"])
+
+
+class HardwareListTestCase(SetupUserMixin, APITestCase):
```
grahamhoyes: Also, just to be consistent with the test case that's getting merged in and the rest of them, rename this to `HardwareListViewTestCase`

hackathon_site/dashboard/frontend/yarn.lock
```diff
@@ -8120,6 +8120,18 @@ memory-fs@^0.5.0:
     errno "^0.1.3"
     readable-stream "^2.0.1"
 
+<<<<<<< HEAD
```
grahamhoyes: Uhh this isn't good, this is a merge conflict in yarn.lock. Were you doing something in the react site that generated this, or did this happen when resolving conflicts?

Since there aren't any react changes here, you should check this back out from develop:

```bash
git checkout develop hackathon_site/dashboard/frontend/yarn.lock
```

hackathon_site/hardware/views.py
```diff
@@ -17,11 +19,18 @@
 
 logger = logging.getLogger(__name__)
 
+from hardware.api_filters import HardwareFilter
+from hardware.models import Hardware, Category, Order
```
grahamhoyes: This is a duplicate import now, since the same thing was done when merging in develop (line 10). After doing a merge, make sure to open up the files and check for unused imports, which could mean something went wrong when resolving the conflicts

hackathon_site/hardware/views.py
```diff
@@ -17,11 +19,18 @@
 
 logger = logging.getLogger(__name__)
 
+from hardware.api_filters import HardwareFilter
```
grahamhoyes: Move this up to be near the rest of the imports (above the `from hardware.models ...` import)

grahamhoyes (commented):
Seems a few things went awry with resolving conflicts

hackathon_site/dashboard/frontend/yarn.lock
Leo6Leo: got it! When I was trying to merge develop to IEEE-3 branch, yarn.lock was not merged.

hackathon_site/hardware/views.py
Leo6Leo: got it!

grahamhoyes (approved):
Good job!

PR #290, opened 2021-07-17
IEEE-5 Category API Updates

Leo6Leo:
## Overview

- Add a few more unit tests around the unique object count
- Look at the debug toolbar stats, may need to select_related(hardware) in the view
- Type annotate the serializer methods to help out swagger

## Unit Tests Created

Serializer tests for the category serializer are here: https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/hardware/test_api.py#L112 

The unique_hardware_count field just returns how many types of hardware are in that category. Make a test like this perhaps:

Add a category with name “Microcontrollers”

Add a hardware with name “Arduino”, quantity available 2

Add another hardware with name “ESP32”, quantity available 3

Add both hardware to the microcontrollers category

Test that unique_hardware_count should return 2 (not 5)

### DIFF
--- a/hackathon_site/hardware/serializers.py
+++ b/hackathon_site/hardware/serializers.py
@@ -36,7 +36,6 @@ class CategorySerializer(serializers.ModelSerializer):
 
     @staticmethod
     def get_unique_hardware_count(obj: Category) -> int:
-        # return Hardware.objects.filter(categories__id=obj.id).count()
         return obj.hardware_set.annotate(Count("id", distinct=True)).count()
 
 
--- a/hackathon_site/hardware/serializers.py
+++ b/hackathon_site/hardware/serializers.py
@@ -35,7 +35,7 @@ class CategorySerializer(serializers.ModelSerializer):
         fields = ("id", "name", "max_per_team", "unique_hardware_count")
 
     @staticmethod
-    def get_unique_hardware_count(obj: Category):
+    def get_unique_hardware_count(obj: Category) -> int:
         #return Hardware.objects.filter(categories__id=obj.id).count()
         return obj.hardware_set.count()
 
--- a/hackathon_site/hardware/serializers.py
+++ b/hackathon_site/hardware/serializers.py
@@ -37,7 +37,7 @@ class CategorySerializer(serializers.ModelSerializer):
     @staticmethod
     def get_unique_hardware_count(obj: Category) -> int:
         # return Hardware.objects.filter(categories__id=obj.id).count()
-        return obj.hardware_set.annotate(Count("id",distinct=True)).count()
+        return obj.hardware_set.annotate(Count("id", distinct=True)).count()
 
 
 class OrderListSerializer(serializers.ModelSerializer):
--- a/hackathon_site/hardware/serializers.py
+++ b/hackathon_site/hardware/serializers.py
@@ -36,7 +36,7 @@ class CategorySerializer(serializers.ModelSerializer):
 
     @staticmethod
     def get_unique_hardware_count(obj: Category) -> int:
-        #return Hardware.objects.filter(categories__id=obj.id).count()
+        # return Hardware.objects.filter(categories__id=obj.id).count()
         return obj.hardware_set.count()
 
 
--- a/hackathon_site/hardware/serializers.py
+++ b/hackathon_site/hardware/serializers.py
@@ -36,7 +36,8 @@ class CategorySerializer(serializers.ModelSerializer):
 
     @staticmethod
     def get_unique_hardware_count(obj: Category):
-        return Hardware.objects.filter(categories__id=obj.id).count()
+        #return Hardware.objects.filter(categories__id=obj.id).count()
+        return obj.hardware_set.count()
 
 
 class OrderListSerializer(serializers.ModelSerializer):
--- a/hackathon_site/hardware/serializers.py
+++ b/hackathon_site/hardware/serializers.py
@@ -37,7 +37,7 @@ class CategorySerializer(serializers.ModelSerializer):
     @staticmethod
     def get_unique_hardware_count(obj: Category) -> int:
         # return Hardware.objects.filter(categories__id=obj.id).count()
-        return obj.hardware_set.count(distinct=True)
+        return obj.hardware_set.annotate(Count("id",distinct=True)).count()
 
 
 class OrderListSerializer(serializers.ModelSerializer):
--- a/hackathon_site/hardware/serializers.py
+++ b/hackathon_site/hardware/serializers.py
@@ -37,7 +37,7 @@ class CategorySerializer(serializers.ModelSerializer):
     @staticmethod
     def get_unique_hardware_count(obj: Category) -> int:
         # return Hardware.objects.filter(categories__id=obj.id).count()
-        return obj.hardware_set.count()
+        return obj.hardware_set.count(distinct=True)
 
 
 class OrderListSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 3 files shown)


grahamhoyes:
I have figured out the integer CSV filter stuff!

```python
from django import forms
from django_filters import rest_framework as filters, widgets

class CSVInputIntegerField(forms.IntegerField):
    widget = widgets.CSVWidget


class IntegerCSVFilter(filters.BaseInFilter):
```

Then inside your `HardwareFilter`, you can do:

```python
    id = IntegerCSVFilter(
        field_name="id",
    )
```

grahamhoyes:
I think you branched off your `IEEE-3` branch when creating this one, since those changes are showing up in the changed files. Make sure to always branch off `develop`, but here's the fix:

```bash
$ git checkout develop
$ git rebase --onto develop IEEE-3 IEEE-5
$ git push --force
```

That will pretty much replay all commits since you branched off of `IEEE-3` onto `develop`, and change the base of the `IEEE-5` branch to be develop. Since that changes the commit history for the `IEEE-5` branch, the force push is necessary to rewrite the commit history on the remote branch on github.

Leo6Leo:
Hey Graham, I found that If we use "Category.objects.all().selected_related(“hardware”)", the category api will not work. In the category model, there is no foreign key called "hardware", therefore it did not work. 
image

I haven't figured out any other methods to improve the SQL efficiency. (remove the similar queries)

hackathon_site/hardware/serializers.py
```diff
@@ -35,8 +35,9 @@ class Meta:
         fields = ("id", "name", "max_per_team", "unique_hardware_count")
 
     @staticmethod
-    def get_unique_hardware_count(obj: Category):
-        return Hardware.objects.filter(categories__id=obj.id).count()
+    def get_unique_hardware_count(obj: Category) -> int:
+        # return Hardware.objects.filter(categories__id=obj.id).count()
+        return obj.hardware_set.count()
```
grahamhoyes: Nice change! This should be identical to what was there before, but add `distinct=True` to the `count` just in case (something we should have done before)

hackathon_site/hardware/test_api.py
```diff
@@ -136,6 +136,41 @@ def test_category_get_success(self):
 
         self.assertEqual(expected_response, data["results"][0])
 
+    def test_hardware_quantity_success(self):
+        self._login()
+
+        self.category = Category.objects.create(name="Microcontrollers", max_per_team=4)
```
grahamhoyes: Note the best to overwrite the `self.category` that's already defined, could make this `self.category2`. Though since you're not using it outside of this test case, you don't actually need the `self.` (same for `hardware1` and `hardware2` below)

hackathon_site/hardware/test_api.py
```diff
@@ -136,6 +136,41 @@ def test_category_get_success(self):
...
+        self.hardware1 = Hardware.objects.create(name="Arduino", quantity_available=2)
+        self.hardware2 = Hardware.objects.create(name="ESP32", quantity_available=3)
+
+        self.hardware1.categories.add(self.category)
+        self.hardware2.categories.add(self.category)
+
+        expected_response = {
```
grahamhoyes: As with the hardware API pr, we shouldn't hard-code the full response, since that's tested in the serializer test cases (and changing the serializer would berak this view unnecessarily).

Can do like you did in the other PR:

```python
expected_unique_hardware_counts = [0, 2]
actual_unique_hardware_counts = [result["unique_hardware_count"] for result in data["results"]]

self.assertEqual(expected_unique_hardware_counts, actual_unique_hardware_counts)
```

hackathon_site/hardware/test_api.py
```diff
@@ -136,6 +136,41 @@ def test_category_get_success(self):
 
         self.assertEqual(expected_response, data["results"][0])
 
+    def test_hardware_quantity_success(self):
+        self._login()
+
+        self.category = Category.objects.create(name="Microcontrollers", max_per_team=4)
+        self.hardware1 = Hardware.objects.create(name="Arduino", quantity_available=2)
```
grahamhoyes: Perhaps add one of these to the `self.category` created in the setUp method as well, and expect the unique counts to be [1, 2]. That way we also cover behaviour when a hardware is in multiple categories, which is possible!

hackathon_site/hardware/test_api.py
```diff
@@ -49,10 +49,10 @@ def test_hardware_get_success(self):
         }
 
         response = self.client.get(self.view)
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        data = response.json()
-
-        self.assertEqual(expected_response, data["results"][0])
+        # self.assertEqual(response.status_code, status.HTTP_200_OK)
```
grahamhoyes: Why was this done? This test doesn't seem to be failing on my end

grahamhoyes:
> Hey Graham, I found that If we use "Category.objects.all().selected_related(“hardware”)", the category api will not work. In the category model, there is no foreign key called "hardware", therefore it did not work.
> image
> 
> I haven't figured out any other methods to improve the SQL efficiency. (remove the similar queries)

@Leo6Leo Try changing the queryset on `CategoryListView` to

```python
queryset = Category.objects.all().prefetch_related("hardware_set")
```

The difference between `select_related` and `prefetch_related` has always confused me, but use `prefetch_related` when dealing with a many-to-many or reverse foreign key field (`hardware_set` is a reverse foreign key). I tested, and that cuts out the duplicate queries.

hackathon_site/hardware/test_api.py
Leo6Leo: @grahamhoyes  So only when we want to use the variable in the other test within the same class, we need to use "self" right? Is that possible to use the variable in the other test in another class?

hackathon_site/hardware/test_api.py
Leo6Leo: my bad, just did it to test whether it works or not. Forgot to remove it.

hackathon_site/hardware/test_api.py
grahamhoyes: This is the basic process of what happens for every single test method:
- The database is reset
- The setUp() method on the test class is called, which sets up your default data
- The particular test method is called

Anything you create in the setUp() method that you want accessible in other test methods, you should define on `self`. Inside the test functions themselves, nothing else will need (or be able to) access them, so no need to use `self`.

> Is that possible to use the variable in the other test in another class?

In another class entirely, no. If you mean another test function, then also no. Only things defined in the `setUp()` method will be accessible to other functions.

There are a few caveats to all of this, but for the most part, if you're not accessing a variable from another method in the class, no need to use `self`.

hackathon_site/hardware/serializers.py
```diff
@@ -35,8 +35,9 @@ class Meta:
         fields = ("id", "name", "max_per_team", "unique_hardware_count")
 
     @staticmethod
-    def get_unique_hardware_count(obj: Category):
-        return Hardware.objects.filter(categories__id=obj.id).count()
+    def get_unique_hardware_count(obj: Category) -> int:
+        # return Hardware.objects.filter(categories__id=obj.id).count()
```
grahamhoyes: Remove the comment

hackathon_site/hardware/test_api.py
```diff
@@ -136,6 +136,29 @@ def test_category_get_success(self):
...
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = response.json()
+
+        expected_unique_hardware_counts = [1, 2]
+        actual_unique_hardware_counts = [
+            result["unique_hardware_count"] for result in data["results"]
+        ]
+
+        print(data)
```
grahamhoyes: Remove this print statement, otherwise it's going to appear in the logs in the test runner

grahamhoyes (commented):
Just two quick things to remove, then looks good!

grahamhoyes (approved):
Good job!

PR #292, opened 2021-07-24
IEEE-32: added minimum age to registration form

YuYing-Liang:
## Overview

- Added a minimum age to settings (set to 14 years old bc  I thought we should let the highschool geniuses in but if its better to be an adult, I'll change it to 18)
- added cleaned_birthday method in registration/forms.py that throws an exception if younger than 14

## Unit Tests Created

- added tests to test a valid & invalid birthday
- @grahamhoyes I didn't quite understand the @ override thing so pls review


hackathon_site/hackathon_site/settings/__init__.py
```diff
@@ -258,6 +258,7 @@
 REGISTRATION_CLOSE_DATE = datetime(2023, 9, 30, tzinfo=TZ_INFO)
 EVENT_START_DATE = datetime(2023, 10, 10, 10, 0, 0, tzinfo=TZ_INFO)
 EVENT_END_DATE = datetime(2023, 10, 11, 17, 0, 0, tzinfo=TZ_INFO)
+MINIMUM_AGE = 14
```
grahamhoyes, line 268: Minor, put a blank line above this just to separate it from the dates

hackathon_site/registration/forms.py
```diff
@@ -149,6 +152,18 @@ def clean(self):
...
         return cleaned_data
 
+    def clean_birthday(self):
+        latest_birthday = (
+            settings.EVENT_START_DATE - timedelta(days=settings.MINIMUM_AGE * 365)
+        ).date()
+        user_birthday = self.cleaned_data["birthday"]
+        if user_birthday > latest_birthday:
+            raise forms.ValidationError(
+                _("User is too young to participate."),
```
grahamhoyes: We give this error directly to the user, so I would change it to `f"You must be {settings.MINIMUM_AGE} to participate."`. The code is fine though

hackathon_site/registration/test_forms.py
```diff
@@ -242,6 +243,21 @@ def test_registration_has_closed(self, mock_is_registration_open):
         self.assertFalse(form.is_valid())
         self.assertIn("Registration has closed.", form.non_field_errors())
 
+    def test_invalid_birthday(self):
+        data = self.data.copy()
+        data["birthday"] = (
+            settings.EVENT_START_DATE - timedelta(days=(settings.MINIMUM_AGE * 365) - 1)
```
grahamhoyes: Ah, this works too! I was thinking you would hard code some dates here and use the override settings decorator to change the minimum age setting just for this test, but this is much smarter (and simpler)

hackathon_site/registration/forms.py
```diff
@@ -149,6 +152,18 @@ def clean(self):
             )
         return cleaned_data
 
+    def clean_birthday(self):
+        latest_birthday = (
+            settings.EVENT_START_DATE - timedelta(days=settings.MINIMUM_AGE * 365)
```
grahamhoyes: Ah joy, there's going to be an edge case with leap years. 

I think we're fine to install [dateutil](https://dateutil.readthedocs.io/en/stable/) and use `relativedelta`, maybe it will come in handy elsewhere as well. Other alternatives are to use 365.24 to get a bit more accurate, or construct a date object manually from parts which is ugly.

When you install it locally (`pip install python-dateutil)`, do `pip freeze | grep dateutil` and add that line to our requirements.txt where it should be alphabetically.

grahamhoyes (commented):
Other than minor changes, LGTM!

grahamhoyes (approved):
LGTM!

PR #293, opened 2021-07-25
IEEE-18 POST: /api/hardware/orders/

EdwardJTL:
## Overview

- Resolves IEEE-18
- Update order status choices; remove `cart` and add `cancelled`
- Create permission class for user with event profile
- Add logger config for Hardware
- Add util to create event profile in SetupUserMixin
- Update OrderListView
  - Use get_permissions to enforce different access constraints for LIST and POST
  - Use get_serializer_class to parse data differently for LIST and POST
- Implement post for creating order
  - Add post action that uses two serializers to perform POST
    - the action is atomic
  - Create serializer for parsing response and creating order object
    - Serializer expects hardware ids and quantities in array, and merges requests for the same hardware
    - Serializer verifies that including the team's past orders that are not returned or cancelled, the new requested quantities will not violate any hardware limit or category limit; else, it throws a 400 bad request
    - Serializer verifies that the requested order is not empty - empty array or 0 total requested quantities; else it throws a 400 bad request
    - Serializer creates new order, with as many order items as it can satisfy with the hardwares' remaining quantities.
  - Create serializer for forming the response
    - Order id is only none when no new order items are created
    - Response has "hardware" entry with array of hardware ids and quantities created
    - Response has "errors" entry with array of hardware ids and the error messages if the order could not be fulfilled in its entirety
- Reordered some import statements 

## Unit Tests Created

- Testing the post action for the aforementioned behaviours, mainly to verify access control, limits, and remaining quantities.

## Steps to QA

- Run tests
- Create entries in dev data bases however you like, and run the request through swagger

note: reopened from #286


hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+        response_hardware = response.json().get("hardware")
+        self.assertEqual(len(response_hardware), 1)
+        response_hardware_id = response_hardware[0].get("hardware_id")
+        self.assertIsNotNone(response_hardware_id, "No hardware id")
+        self.assertEqual(response_hardware_id, simple_hardware.id)
+        response_hardware_fulfilled = response_hardware[0].get("quantity_fulfilled")
+        self.assertIsNotNone(response_hardware_fulfilled, "No fulfilled quantity")
+        self.assertEqual(response_hardware_fulfilled, 1)
+
+        pass
```
grahamhoyes: Whats this lil guy doing here

hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        order_id = response.json().get("order_id")
+        self.assertIsNotNone(order_id, "No order id returned")
+
+        order = Order.objects.get(pk=order_id)
+        self.assertEqual(len(order.items.all()), 1, "More than 1 order item created")
+        self.assertCountEqual(order.hardware_set.all(), [simple_hardware])
+
+        response_hardware = response.json().get("hardware")
```
grahamhoyes: For this block, it might just be a bit cleaner to hard code your expected repsonse and do `self.assertEqual(response.json(), expected_response)`. Thats what we typically do for serializer tests to make them easy to read

hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+        self.assertEqual(len(response_hardware), 1)
+        response_hardware_id = response_hardware[0].get("hardware_id")
+        self.assertIsNotNone(response_hardware_id, "No hardware id")
```
grahamhoyes: Technically redundant, since the `assertEqual` below will fail if it's None

hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+            max_per_team=1,
+            picture="/picture/location",
+        )
+        hardware.categories.add(self.category_limit_10.pk)
+
+        request_data = {"hardware": [{"id": hardware.id, "quantity": 2}]}
+        response = self.client.post(self.view, request_data, format="json")
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        pass
```
grahamhoyes: Seems there are a few of these! I won't comment on the rest, just do a search for them

hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+            quantity_available=10,
+            max_per_team=1,
+            picture="/picture/location",
+        )
+        hardware.categories.add(self.category_limit_10.pk)
+
+        request_data = {"hardware": [{"id": hardware.id, "quantity": 2}]}
+        response = self.client.post(self.view, request_data, format="json")
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
grahamhoyes, line 273: Could you add a check for the response body too? `{'non_field_errors': ['Hardware <name> limit reached']}`.

Since we don't have swagger definitions for errors, it's sometimes nice to peak at the tests to see what they'll look like. These are also likely something we'll spit to the end user through a snackbar.

hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+            team=self.user.profile.team, status="Picked Up"
+        )
+        picked_up_order_item = OrderItem.objects.create(
+            order=picked_up_order, hardware=hardware
+        )
+
+        request_data = {"hardware": [{"id": hardware.id, "quantity": 2}]}
+        response = self.client.post(self.view, request_data, format="json")
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
grahamhoyes, line 319: Ditto about a test for the response body

grahamhoyes: Same for the tests below if it's not too annoying, I won't comment on all of them

hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+            notes="notes",
+            quantity_available=10,
+            max_per_team=4,
+            picture="/picture/location",
+        )
+        hardware.categories.add(self.category_limit_10.pk)
+
+        order = Order.objects.create(team=self.user.profile.team, status="Picked Up")
+        healthy_order_item = OrderItem.objects.create(order=order, hardware=hardware)
+        healthy_order_item.part_returned_health = "Healthy"
```
grahamhoyes: You can do this inline with the `OrderItem.objects.create(..., part_returned_health="Healthy")`

hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+            notes="notes",
+            quantity_available=10,
+            max_per_team=10,
+            picture="/picture/location",
+        )
+        hardware.categories.add(self.category_limit_4.pk)
+
+        order = Order.objects.create(team=self.user.profile.team, status="Picked Up")
+        healthy_order_item = OrderItem.objects.create(order=order, hardware=hardware)
+        healthy_order_item.part_returned_health = "Healthy"
```
grahamhoyes: Ditto, can inline into creating the order item

hackathon_site/hardware/test_api.py
```diff
@@ -188,9 +188,606 @@ def test_hardware_get_success(self):
...
+        order = Order.objects.get(pk=order_id)
+        self.assertCountEqual(order.hardware_set.all(), [hardware_1, hardware_2])
+        self.assertEqual(
+            order.items.filter(hardware=hardware_1).count(), num_hardware_1_requested
+        )
+        self.assertEqual(
+            order.items.filter(hardware=hardware_2).count(), num_hardware_2_requested
+        )
+
+        response_hardware = response.json().get("hardware")
```
grahamhoyes: Same as mentioned above, it would be clearer just to hard code the expected response json and compare the response with that

hackathon_site/event/permissions.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import permissions
+
+
+class UserHasProfile(permissions.IsAuthenticated):
+    """
+    Allows access only to authenticate users with profiles.
+    """
+
+    def has_permission(self, request, view):
+        return super().has_permission(request, view) and bool(
```
grahamhoyes: Minor, the `bool()` is unnecessary

grahamhoyes (commented):
Stellar stuff!

Functionality looks good. I didn't pick apart all of the tests, but from testing it out and reading through them everything looks correct.

Comments below are mostly just stylistic stuff to make the tests a bit easier to read, should people go through them in the future looking for response formats.

hackathon_site/hardware/test_api.py
EdwardJTL: oh for some reason i thought all tests need this "pass" thing LOL

hackathon_site/hardware/serializers.py
```diff
@@ -34,9 +39,149 @@ def get_unique_hardware_count(obj: Category):
...
+                    category_counts.get(category, 0)
+                    + team_hardware_count
+                    + requested_quantity
+                )
+        for (category, count) in category_counts.items():
+            if count > category.max_per_team:
+                error_messages.append("Category {} limit reached".format(category.name))
+        if error_messages:
+            error_message = "; ".join(error_messages)
+            raise serializers.ValidationError(error_message)
```
grahamhoyes: What happens if this just uses an array directly?

grahamhoyes (approved):
:rocket: :rocket: :rocket:

PR #294, opened 2021-07-25
IEEE-6: Create Redux Slice for hardware store

lisa-sa-li:
## Overview

- https://ieeeuoft.atlassian.net/browse/IEEE-6

## Unit Tests Created

- Checks hardwareSliceSelector returns the hardware store

## Steps to QA

- N/A


hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.module.scss
```diff
@@ -3,7 +3,7 @@
```
lisa-sa-li, line 3: oops I forgot that I commited this, but it fixes and issue with the inventory button filter buttons being sticky

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.test.js
```diff
@@ -0,0 +1,17 @@
+import {
```
grahamhoyes: Let's do the tests in typescript too

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export interface HardwareState {
+    isLoading: boolean;
+    error: string | null;
+    next: string | null;
+}
+
+export const hardwareReducerName = "hardware";
+const hardwareAdapter = createEntityAdapter();
```
grahamhoyes:
```suggestion
const hardwareAdapter = createEntityAdapter<Hardware>();
```

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+    name: hardwareReducerName,
+    initialState: hardwareAdapter.getInitialState(initialState),
+    reducers: {},
+});
+
+export const { actions, reducer } = hardwareSlice;
+export default reducer;
+
+// Selectors
+export const hardwareSliceSelector = (state) => state[hardwareReducerName];
```
grahamhoyes: `state` here needs to be typed with the type of our root state (I'm getting an error otherwise).

You'll have to convert `slices/store.js` to `slices/store.ts`. You should only get one error about `history` being untyped in `rootReducer`, just add `History` to the import at the top from `"history"` and do

```typescript
const rootReducer = (history: History) => ...
```

After that, you can export the type of the root state at the bottom:

```typescript
export type RootState = ReturnType<typeof store.getState>;
```

Then in this file:

```typescript
import { RootState } from "slices/store";
...

export const hardwareSliceSelector = (state: RootState) => state[hardwareReducerName];
```

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+    datasheet: string;
+    quantity_available: number;
+    notes: string;
+    max_per_team: number;
+    picture: string;
+    categories: number[];
+    quantity_remaining: string;
+}
+
+export interface HardwareState {
```
grahamhoyes: This isn't actually the type of the hardware state, since it doesn't include the keys added by `createEntityAdapter`.

You could do

```typescript
// Note: not exported
interface HardwareExtraState {
    isLoading: boolean;
    error: string | null;
    next: string | null;
}

const extraState: HardwareExtraState = {
    isLoading: false,
    error: null,
    next: null,
};

const hardwareAdapter = createEntityAdapter<Hardware>();

export const initialState = hardwareAdapter.getInitialState(extraState);
export type HardwareState = typeof initialState;
```

With all that, we can create properly typed selectors and stuff. For example:

```typescript
export const isLoadingSelector = createSelector(
    hardwareSliceSelector,
    (state: HardwareState) => state.isLoading
);
```

If you did that, your IDE should be able to properly autocomplete the fields that `state` has (the ones in `HardwareExtraState`, as well as `ids` and `entities`).

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+    name: string;
+    model_number: string;
+    manufacturer: string;
+    datasheet: string;
+    quantity_available: number;
+    notes: string;
+    max_per_team: number;
+    picture: string;
+    categories: number[];
+    quantity_remaining: string;
```
grahamhoyes: number

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface HardwareState {
+    isLoading: boolean;
+    error: string | null;
+    next: string | null;
+}
+
+export const hardwareReducerName = "hardware";
+const hardwareAdapter = createEntityAdapter();
+export const initialState: HardwareState = {
```
grahamhoyes: See my other comment, but this isn't actually the initial state

hackathon_site/dashboard/frontend/package.json
```diff
@@ -37,8 +37,8 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "i": "yarn install --frozen-lockfile",
-    "pretty": "yarn prettier --write 'src/**/*.(scss|js)'",
-    "precheck": "yarn pretty && yarn test -o -u --watchAll=false"
+    "pretty": "yarn prettier --write 'src/**/*.(scss|js|ts)'",
```
grahamhoyes: Gotta include `tsx` in here too. React typescript files have to be .tsx, can't use .ts like you can with .js

grahamhoyes (dismissed):
Just the tsx thing in the formatter (can leave that for the first PR to add .tsx files if you want), LGTM beyond that.

I think the failing backend test is because of an edge condition when they run around midnight that we really have to track down, will re-run the check and see if it passes.

grahamhoyes (approved):

PR #297, opened 2021-07-31
Connect Logout Button

EdwardJTL:
## Overview

- Resolves IEEE-40
- Adds async thunk for log out
  - post to `user/logout`
  - display snack bar with response.data.details on failure
  - push to "/" after the request succeeds
- Adds states and reducer for log out
  - when pending:
    - sets `logout.loading` to true when pending
  - when fulfilled:
    - clears `userData.user` to `null` to allow redirect to log in page
    - sets `isAuthenticated`to false
    - clears other logout states
  - when rejected:
    - clears `loading` and sets `failure`
- connects logout to NavBar
- fixes bug in extracting csrf token from cookie
- fixes bug where csrf token does not get updated on subsequent login calls
- exports `UnconnectedNavbar` component
  - `Navbar` is no longer a default export
- added tests

## Unit Tests Created

- tests for logout thunk and reducers to check the aforementioned behaviours
- tests that UnconnectedNavbar component calls the log out handler

## Steps to QA

- run react project, sign in (through either react project or django sites). 
- click logout in the narbar and check that:
  - the page is redirected to log in
  - you are signed out on all the django pages and cannot make further requests (/admin and /swagger)
- sign in again, then sign out (on react site). do the same checks

### DIFF
--- a/hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.js
+++ b/hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.js
@@ -1,6 +1,7 @@
 import React from "react";
 import styles from "./Navbar.module.scss";
 import { Link } from "react-router-dom";
+import { push } from "connected-react-router";
 import { connect } from "react-redux";
 // Images and logos
 import DashboardIcon from "@material-ui/icons/Dashboard";
@@ -12,7 +13,9 @@ import { ReactComponent as Inventory } from "assets/images/icons/Hardware.svg";
 // Components
 import Button from "@material-ui/core/Button";
 
-const UnconnectedNavbar = ({ cartQuantity, pathname }) => (
+import { logout, logoutSelector } from "slices/users/userSlice";
+
+const UnconnectedNavbar = ({ logout, cartQuantity, pathname }) => (
     <nav className={styles.nav}>
         <div className={styles.navFlexDiv}>
             <Link to={"/"}>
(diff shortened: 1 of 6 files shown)


hackathon_site/dashboard/frontend/src/components/general/Navbar/Navbar.js
```diff
@@ -89,6 +94,6 @@ const mapStateToProps = (state) => ({
     pathname: state.router.location.pathname,
 });
 
-const Navbar = connect(mapStateToProps)(UnconnectedNavbar);
+const Navbar = connect(mapStateToProps, { logout: logout })(UnconnectedNavbar);
```
grahamhoyes: Minor, can just do `{ logout }` for mapDispatchToProps

EdwardJTL: done

EdwardJTL:
lmk if we have a preference for naming. with the login stuff, we used both `login` and `logIn` in different places. I used all lowercase `logout` everywhere here.

grahamhoyes:
> lmk if we have a preference for naming. with the login stuff, we used both `login` and `logIn` in different places. I used all lowercase `logout` everywhere here.

I think this is just an inconsistency of whether we treat login and logout as one word or two. Doesn't matter from a variable naming perspective, all lowercase sounds good to me

grahamhoyes (approved):
:rocket:

PR #298, opened 2021-08-02
Post Delete Signals for Event/Profile and Registration/Application

EdwardJTL:
## Overview

- Resolves IEEE-29
- Adds post delete signals for event profile and registration application
  - deletes associated teams if deletion resulted in empty team

## Unit Tests Created

- Tests both signals
  - Tests simple case when deleting the sole application/profile on a team
  - Tests cascading case when the sole user on a team is deleted
  - Tests that deletion is not performed when the team is not empty

## Steps to QA

- On admin site, make a user
- QA Registration:
  - Make a team and an application with that user
  - Delete that application. Verify that the team was also deleted
  - Now make another team and application with that user
  - Delete that user. Verify that both the application and team are deleted
  - Now make two users. Make one team and two applications (one for each user).
  - Delete one of the applications. Verify that the team is not deleted.
- QA Event
  - Same steps as above. Swap application for profile.

Note: make sure you are creating/checking the right team model.

### DIFF
--- a/hackathon_site/event/__init__.py
+++ b/hackathon_site/event/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = "event.apps.EventConfig"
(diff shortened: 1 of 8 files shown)


grahamhoyes (approved):
Awesome!

PR #300, opened 2021-08-11
IEEE-12 Incident List API

### LINKED ISSUE #45: GET: /api/hardware/incidents/
List view of all hardware incidents, for use by the tech team admin panel.

- All fields
    - Also:
        - order_item
        - order__team
        - hardware__id
    - Filters:
        - By order_item__hardware__id
        - By order_item__order__team__id

Leo6Leo:
## Overview

- Resolves #45 
- List view of all hardware incidents, for use by the tech team admin panel.

All fields

Also:

order_item

order__team

hardware__id

Filters:

By order_item__hardware__id

By order_item__order__team__id

## Unit Tests Created

- Test for IncidentListView


hackathon_site/hardware/api_urls.py
```diff
@@ -8,6 +8,7 @@
     path("hardware/", views.HardwareListView.as_view(), name="hardware-list",),
     path("orders/", views.OrderListView.as_view(), name="order-list"),
     path("categories/", views.CategoryListView.as_view(), name="category-list"),
+    path("incidents/", views.IncidentsListView.as_view(), name="incidents-list"),
```
grahamhoyes: Minor, but the `name` param here is typically singular form, so `name="incident-list"`

hackathon_site/hardware/serializers.py
```diff
@@ -39,6 +39,25 @@ def get_unique_hardware_count(obj: Category) -> int:
         return obj.hardware_set.annotate(Count("id", distinct=True)).count()
 
 
+class IncidentsSerializer(serializers.ModelSerializer):
```
grahamhoyes: Minor, `IncidentSerializer` (singular form to match the rest)

hackathon_site/hardware/models.py
```diff
@@ -136,3 +136,11 @@ class Incident(models.Model):
 
     def __str__(self):
         return f"{self.id}"
+
+    @property
```
grahamhoyes: As we talked about today, this isn't ideal. We can accomplish this in the serializer instead,  details below!

hackathon_site/hardware/serializers.py
grahamhoyes: As we talked about:

```python
class OrderItemSerializer(serializers.ModelSerializer):
    class Meta:
        model = OrderItem
        fields = ("id", "hardware", "order", "part_returned_health")


class IncidentsSerializer(serializers.ModelSerializer):
    order_item = OrderItemSerializer()
    team_id = serializers.SerializerMethodField()

    class Meta:
        model = Incident
        fields = (
            "id",
            "state",
            "time_occurred",
            "description",
            "order_item",
            "team_id",
            "created_at",
            "updated_at",
        )

    @staticmethod
    def get_team_id(obj: Incident):
        return obj.order_item.order.team.id
```

grahamhoyes: Eventually, we'll need a custom field type for `team_id` that lets us use this serializer for POST requests as well, but that's a problem for another day.

hackathon_site/hardware/views.py
```diff
@@ -42,6 +43,17 @@ def get(self, request, *args, **kwargs):
         return self.retrieve(request, *args, **kwargs)
 
 
+class IncidentsListView(mixins.ListModelMixin, generics.GenericAPIView):
```
grahamhoyes: Minor, `IncidentListView` (singular again)

hackathon_site/hardware/views.py
```diff
@@ -42,6 +43,17 @@ def get(self, request, *args, **kwargs):
...
+    queryset = Incident.objects.all()
```
grahamhoyes: For efficiency:

```python
    queryset = Incident.objects.all().select_related(
        "order_item", "order_item__order__team"
    )
```

hackathon_site/hardware/api_filters.py
```diff
@@ -13,6 +13,16 @@ class IntegerCSVFilter(filters.BaseInFilter):
     field_class = CSVInputIntegerField
 
 
+class IncidentsFilter(filters.FilterSet):
```
grahamhoyes: `IncidentFilter` (singular)

hackathon_site/hardware/api_filters.py
```diff
@@ -13,6 +13,16 @@ class IntegerCSVFilter(filters.BaseInFilter):
...
+    queryset = Incident
+    serializer_class = IncidentsSerializer
+
+    order_item__hardware__id = filters.NumberFilter(label="order_item__hardware__id")
```
grahamhoyes: These work great! But just to keep the field names and labels cleaner, can you do:

```python
    hardware_id = filters.NumberFilter(field_name="order_item__hardware__id")
    team_id = filters.NumberFilter(field_name="order_item__order__team__id")
```

Still works the exact same, just makes the query parameter cleaner and builds a label automatically

hackathon_site/hardware/serializers.py
grahamhoyes: Will need a test for this serializer, in `hardware/tests.py`. Since there's nothing special going on in the serializer, you should be good with just one test method in the class where you hard-code the expected JSON response. Since time fields are annoying to fix in tests, take a look at [this](https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/hardware/tests.py#L204) order serializer test for how to do it.

hackathon_site/hardware/test_api.py
```diff
@@ -159,6 +160,58 @@ def test_hardware_quantity_success(self):
...
+        self.view = reverse("api:hardware:incidents-list")
+
+    def test_incident_get_success(self):
+        self._login()
+
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        queryset = Incident.objects.all()
+        # need to provide a request in the serializer context to produce absolute url for image field
```
grahamhoyes, line 385: Cool! We probably need to fix a bunch of other tests to do that

hackathon_site/hardware/test_api.py
```diff
@@ -159,6 +160,58 @@ def test_hardware_quantity_success(self):
         self.assertEqual(expected_unique_hardware_counts, actual_unique_hardware_counts)
 
 
+class IncidentsListViewTestCase(SetupUserMixin, APITestCase):
```
grahamhoyes: Also needs test cases for your two filters:
- One that filters by hardware ID. You already have two hardware instances, you'll just need to create a second order item and incident.
- One that filters by team ID. You'll have to create a second team and order to do that, which you can tie to the second order item you'll need for testing the hardware filter.

grahamhoyes (commented):
Fantastic start!

hackathon_site/hardware/serializers.py
Leo6Leo: How does "get_team_id" know the value should go to the "team_id" field? Will Django automatically recognize the "team_id" in the "get_team_id" method? @grahamhoyes

grahamhoyes: It's the other way around - by default, `SerializerMethodField` will look for a method called `get_<field name>`: https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield

Leo6Leo: got it! It makes a lot of sense rn. thanks Graham!

hackathon_site/hardware/views.py
```diff
@@ -43,6 +44,19 @@ def get(self, request, *args, **kwargs):
         return self.retrieve(request, *args, **kwargs)
 
 
+class IncidentListView(mixins.ListModelMixin, generics.GenericAPIView):
+    queryset = Incident.objects.all().select_related(
+        "order_item", "order_item__order__team"
+    )
+    serializer_class = IncidentSerializer
+
+    filter_backends = (filters.DjangoFilterBackend, SearchFilter)
```
grahamhoyes, line 60: `SearchFilter` is unused here, since you also have to specify what fields can be searched on.

What do you think about these fields?

```python
    search_fields = (
        "state",
        "order_item__order__team__team_code",
        "order_item__hardware__name",
        "order_item__hardware__manufacturer"
    )
```

With those, update the queryset to be:

```python
    queryset = Incident.objects.all().select_related(
        "order_item", "order_item__order__team", "order_item__hardware"
    )
```

Then just one simple test case where you do

```python
url = self._build_filter_url(search="name")
```

and this will be good to go!

grahamhoyes (commented):
Great work! Just need to make use of the `SearchField`

Sorry for taking so long to review this one 🙃

hackathon_site/hardware/test_api.py
```diff
@@ -400,6 +400,18 @@ def test_team_id_filter(self):
         returned_ids = [res["id"] for res in results]
         self.assertCountEqual(returned_ids, [1])
 
+    def test_name_search_filter(self):
+        self._login()
+
+        url = self._build_filter_url(name="name")
```
grahamhoyes: For the search filter, it's just `<url>?search=whatever`, and that will search all of the search fields you specify.

This test is actually a false-positive - since you have no filter arguments that look for `?name`, nothing happens to the results. But since there's only one `Incident` object created, this test doesn't capture that - and technically, none of the other ones do either.

Can you add a second `Incident` object to this test class? Your hardware filter test will be good to go then, and if you attach the incident to a different team (aka new order and order item as well), your team filter test will be good as well.

Leo6Leo:
image

These mailer tests seem failed. All my tests have passed. 

image
And all error msgs are the same except for those two with the exclamation mark.

This is the error msg of the one with the exclamation mark.
image

hackathon_site/hardware/test_api.py
```diff
@@ -318,6 +319,114 @@ def test_hardware_quantity_success(self):
...
+
+        self.order_item2 = OrderItem.objects.create(
+            order=self.order, hardware=self.other_hardware,
+        )
+
+        self.incident = Incident.objects.create(
+            state="Broken",
+            description="Description",
+            order_item=self.order_item,
+            time_occurred="2022-08-08 00:00:00",
```
grahamhoyes: It looks like Django is throwing a warning because it expects datetimes to be timezone aware. Easy fix:

```python
from datetime import datetime
...
from django.conf import settings
...

self.incident = Incident.objects.create(
    state="Broken",
    description="Description",
    order_item=self.order_item,
    time_occurred=datetime(2022, 8, 8, tzinfo=settings.TZ_INFO),
)
```

hackathon_site/hardware/test_api.py
```diff
@@ -318,6 +319,114 @@ def test_hardware_quantity_success(self):
...
+        )
+
+        self.incident2 = Incident.objects.create(
+            state="Missing",
+            description="Description",
+            order_item=self.order_item2,
+            time_occurred="2022-08-08 00:00:00",
```
grahamhoyes: Same as above

hackathon_site/hardware/test_api.py
```diff
@@ -318,6 +319,114 @@ def test_hardware_quantity_success(self):
...
+
+        url = self._build_filter_url(team_id="1")
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = response.json()
+        results = data["results"]
+
+        returned_ids = [res["id"] for res in results]
+        print(results)
```
grahamhoyes: An errant print!

hackathon_site/hardware/views.py
```diff
@@ -43,6 +44,26 @@ def get(self, request, *args, **kwargs):
         return self.retrieve(request, *args, **kwargs)
 
 
+class IncidentListView(mixins.ListModelMixin, generics.GenericAPIView):
```
grahamhoyes, line 47: Once #308 is merged, up to you if you want to add permissions into the view and tests now or if you want to do that as a separate PR

grahamhoyes:
The mail tests work now, I reran the github pipeline. We have a know issue where they fail around midnight Toronto time - we don't know why or exactly what the time range is, and we don't know if they'll also fail if we run them in real life around midnight!

hackathon_site/hardware/test_api.py
Leo6Leo: I have encountered this bug before, and when I set the year from 2021 to 2022, it just disappeared on my end.

hackathon_site/hardware/views.py
Leo6Leo: I will make a separate PR to add the permissions to these views and tests when IEEE-69 is merged.

grahamhoyes (approved):
Nice work! 

As long as you hit the "Update branch" button in github, this should stay approved and you can merge it in.

PR #302, opened 2021-08-14
Hardware API Filters and Orders

grahamhoyes:
## Overview

Adds the last of the changes to the hardware API necessary to connect the inventory page

- Filtering by category ID
- Renames the `id` filter to `ids`, just to make it clear it takes a list
- Ordering by name and quantity remaining

## Unit Tests Created

- Combines the hardware list view test classes that were previously duplicated (so one was ignored)
- Tests for category filter and ordering

## Steps to QA

- Play around with the filters and ordering in the browsable API at http://localhost:8000/api/hardware/hardware/ by clicking the "filters" button


hackathon_site/hardware/api_filters.py
```diff
@@ -28,8 +28,14 @@ def filter_in_stock(queryset, _, value):
...
         field_name="id",
         label="Comma separated list of hardware IDs",
         help_text="Comma separated list of hardware IDs",
     )
+
+    categories = IntegerCSVFilter(
+        field_name="categories",
+        label="Comma separated list of category IDs",
+        help_text="Comma separated list of category IDs",
+    )
```
albogdan: Maybe `hardware_ids` and `category_ids` would be better names for these two?

albogdan (commented):
LGTM besides that one naming convention

albogdan (approved):

PR #303, opened 2021-08-14
Connect hardware store

grahamhoyes:
## Overview

- Loosely resolves IEEE-23, IEEE-28, and potentially other stuff
- Begins the work of connecting the inventory page to the redux store. At this point, we have:
    - The filters form submits its values to the store for persistence
    - There is a thunk to get hardware with filters
    - The gotten hardware will be put into the inventory grid
    - There is a search box for combining a search string with the filters

There is a lot of new typescript and testing magic going on here, so please ask questions if you have any!

## Unit Tests Created
Unit tests get interesting in this one! Up until now, our thunks have been tested in isolation (typical unit test style), and we have tested the _unconnected_ versions of our components.

The latest [guiding principles from Redux on writing tests](https://redux.js.org/usage/writing-tests) are to write tests that are closer to integration tests, rather than testing redux and components in isolation. There are two key points I take away from this approach:
1. We should be testing thinks as the users actually see them (this is a guiding principle of React Testing Library)
2. Avoid testing "implementation details" (such as redux), and instead test actual results

In most cases, this means that we will now render all of our components in tests wrapped in a real redux store (see the new `render` helper function), and mock API responses in all tests. We can prepopulate the store with the `preloadedState` option, or for things that use `createEntityAdapter` I prefer just to dispatch the thunk that fetches the data as part of the test.

To test things that are tied to transient states, like loading bars or disabled buttons when a request is in progress, we can just put a delay (with `setTimout()` and some Promise magic) in the mocked request, and use `waitFor` in the tests to make sure that the appropriate components appear and disapper. This is in contrast to setting the `isLoading` variable in the state manually, since that is just an "implementation detail". 

The main tests that have been added:
- Tests for the inventory filters. In addition to being UI tests, these are effectively the tests for the `getHardwareWithFilters` thunk, and are the reason that the tests for the latter are so sparse.
- Tests for the inventory grid. These tests manually dispatch a `getHardwareWithFilters` thunk so that data is in the store for the component to render - we're testing the connected component, not the unconnected one!
  - There actually isn't even an unconnected version of this component anymore, since now nothing is stopping us from using redux hooks for simple stuff
- Tests for the inventory search box. Same as the filters test, this asserts all the way down on the level of the mocked API.
- Basic tests for the inventory page   

## Steps to QA

- Create some hardware and categories through the admin site
  - Categories aren't actually used to populate the options in the filters just yet, but it is prepopulated with ids 1-10
- Go to the inventory page, and test out the filters and search box

Note: The styling on the inventory page is really wonky. @lisa-sa-li will fix that :)


hackathon_site/dashboard/frontend/src/api/api.ts
```diff
@@ -1,8 +1,11 @@
-import axios from "axios";
+import axios, { AxiosResponse } from "axios";
+
+// Re-export the response type, so it's available without needing to import axios
+export type { AxiosResponse } from "axios";
 
 export const SERVER_URL =
     process.env.NODE_ENV === "development"
-        ? process.env.REACT_APP_DEV_SERVER_URL.replace(/\/$/, "")
+        ? process.env.REACT_APP_DEV_SERVER_URL?.replace(/\/$/, "")
```
EdwardJTL: would this set the variable to `undefined` if the env variable is undefined as well?

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -0,0 +1,44 @@
+/** Generics */
+export interface APIListResponse<T> {
+    count: number;
+    next: string | null;
```
EdwardJTL, line 4: is this like an optional string?

hackathon_site/dashboard/frontend/src/components/HOCs/withParticipantCheck/withParticipantCheck.test.js
```diff
@@ -60,9 +61,9 @@ describe("withAuthenticationCheck", () => {
 
         const store = mockStore(mockState);
 
-        const { queryByText } = render(
-            withStore(<WrappedComponent>{content}</WrappedComponent>, store)
-        );
+        const { queryByText } = render(<WrappedComponent>{content}</WrappedComponent>, {
+            store,
+        });
```
EdwardJTL, line 66: some of them use the format
```
render(
<Wrapped>...,
{ store }
);
```
and some use 
```
render(<Wrapped>..., {
store,
});
```
do we want to be consistent with these?

grahamhoyes: Do you just mean the formatting? That's left up to prettier based (mostly) on the line length, we have little control over it.

hackathon_site/dashboard/frontend/src/api/types.ts
grahamhoyes: This just says that `next` is either a string or null (which is what the API will return). You can make a [union type](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#union-types) by chaining types with `|`.

The normal way to mark something as optional is to do `next?: string`, which pretty much means `next: string | undefined`.

hackathon_site/dashboard/frontend/src/api/api.ts
grahamhoyes: Yes, good catch.

I want to avoid implicitly coding in a default `localhost:8000`, so I've updated it to check first and throw an error if it's not set.

hackathon_site/dashboard/frontend/src/components/inventory/InventoryGrid/InventoryGrid.tsx
```diff
@@ -0,0 +1,49 @@
...
+                    <Grid
+                        xs={6}
+                        sm={4}
+                        md={3}
+                        lg={2}
+                        xl={1}
+                        className={styles.Item}
+                        key={item.id}
+                        item
+                        // onClick={() => setItemOverviewId(item.id)}
```
EdwardJTL, line 29: should this comment be checked in?

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
```diff
@@ -1,53 +1,62 @@
...
-import Item from "components/inventory/Item/Item";
+import InventoryFilter from "components/inventory/InventoryFilter/InventoryFilter";
+import InventoryGrid from "components/inventory/InventoryGrid/InventoryGrid";
 import ProductOverview from "components/inventory/ProductOverview/ProductOverview";
-import { productInformation, inventoryItems } from "testing/mockData";
+import { productInformation } from "testing/mockData";
+import InventorySearch from "components/inventory/InventorySearch/InventorySearch";
+import { RootState } from "slices/store";
+import { connect, ConnectedProps } from "react-redux";
+import { clearFilters, getHardwareWithFilters } from "slices/hardware/hardwareSlice";
```
EdwardJTL: are we sorting imports by where they are imported from?

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
```diff
@@ -1,53 +1,62 @@
...
         setItemOverviewId(null);
     };
 
     // use itemOverviewId to fetch the info from the store
-    React.useEffect(() => console.log("itemOverviewId", itemOverviewId), [
-        itemOverviewId,
-    ]);
+    // React.useEffect(() => console.log("itemOverviewId", itemOverviewId), [
+    //     itemOverviewId,
+    // ]);
```
EdwardJTL: should this stay?

hackathon_site/dashboard/frontend/src/components/inventory/InventoryGrid/InventoryGrid.tsx
grahamhoyes: I'm going to leave this one in as a reminder for what should go there, but there is definitely lots of commented out code I need to remove

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
grahamhoyes: Yeah, I should clean that up. I'd like imports to be:

- React import first
- Imports from libraries
- Imports of local components

Within each of those i don't feel like alphabetizing, but I should clean it up a bit. Will do that shortly.

EdwardJTL (approved):

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -0,0 +1,44 @@
...
+export interface HardwareFilters {
+    in_stock?: boolean;
+    hardware_ids?: number[];
+    category_ids?: number[];
+    search?: string;
+    ordering?: HardwareOrdering;
+}
+
+/** Category API */
+export interface Category {
```
Leo6Leo, line 39: What are the purpose of these codes? Are they the "template" for receiving API data? Tell react what type of each field we are going to receive right?

hackathon_site/dashboard/frontend/src/testing/mockData.tsx
```diff
@@ -151,17 +152,22 @@ export const addCartTest = () => {
...
-    { id: 7, name: "1080p_cameras", qty: 4 },
-    { id: 8, name: "Grove", qty: 22 },
-    { id: 9, name: "Grove_style_boards", qty: 12 },
-    { id: 10, name: "Grove_style_modules", qty: 10 },
+export const inventoryCategories: Category[] = [
+    { id: 1, name: "MCU", max_per_team: 12, unique_hardware_count: 10 },
+    { id: 2, name: "MCU_limit_3", max_per_team: 3, unique_hardware_count: 6 },
+    { id: 3, name: "FPGA", max_per_team: 2, unique_hardware_count: 9 },
+    { id: 4, name: "Sensors", max_per_team: 12, unique_hardware_count: 21 },
+    { id: 5, name: "Sensors_limit_2", max_per_team: 2, unique_hardware_count: 3 },
```
Leo6Leo, line 160: How can we put json file here? Or get the data from API?

hackathon_site/dashboard/frontend/src/pages/Reports/Reports.test.js
```diff
@@ -1,9 +1,10 @@
...
-import { render } from "@testing-library/react";
+
+import { render } from "testing/utils";
+
 import Reports from "./Reports";
-import { withStoreAndRouter } from "testing/helpers";
 
 test("renders without crashing", () => {
-    const { getByText } = render(withStoreAndRouter(<Reports />));
+    const { getByText } = render(<Reports />);
```
Leo6Leo, line 8: Hey Graham, why are you trying to replacing all "withStoreAndRouter"?

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -0,0 +1,44 @@
+/** Generics */
+export interface APIListResponse<T> {
```
YuYing-Liang, line 2: What's the advantage of using Generic type over the any type?

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -0,0 +1,44 @@
...
+    manufacturer: string;
+    datasheet: string;
+    quantity_available: number;
+    max_per_team: number;
+    picture: string;
+    categories: number[];
+    quantity_remaining: number;
+}
+
+export type HardwareOrdering =
```
YuYing-Liang, line 23: this is cool

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -1,48 +1,150 @@
...
+    },
 });
 
 export const { actions, reducer } = hardwareSlice;
 export default reducer;
 
+export const { setFilters, clearFilters } = actions;
+
 // Selectors
 export const hardwareSliceSelector = (state: RootState) => state[hardwareReducerName];
```
YuYing-Liang, line 138: I don't know Redux very well but doesn't this return a type hardwareReducer (which takes in a previous state and returns the new state of the hardwareSlice right?), which is line 133? and a selector takes a redux state and returns data derived from that state ... they kind of sound like the same thing to me, why do we need both?

hackathon_site/dashboard/frontend/src/slices/ui/uiSlice.js
```diff
@@ -15,16 +15,16 @@ const uiSlice = createSlice({
     initialState: initialState,
     reducers: {
         toggleCheckedOutTable: (state) => {
```
YuYing-Liang, line 17: this is just a general comment but these reducers seem to be called only in one place so why do we have to create a whole slice for these states?

YuYing-Liang (commented):
Nice job Graham! Redux & Typescript are very cool :) 

I was wondering what's the convention for files having .ts vs. .tsx extensions. I thought maybe .tsx was for anything that renders to the DOM but unit tests are also .tsx so I'm not sure.

hackathon_site/dashboard/frontend/src/components/general/Header/Header.test.js
```diff
@@ -1,19 +1,20 @@
...
-        const { getByText, getByTestId } = render(withStoreAndRouter(<Header />));
+        const { getByText, getByTestId } = render(<Header />);
         expect(getByText(/Hackathon Name/i)).toBeInTheDocument();
         expect(getByTestId("headerLogo")).toBeInTheDocument();
         expect(getByText("Dashboard")).toBeInTheDocument(); // Checking if navbar is in header
     });
 
     test("renders header with no navbar", () => {
         const { getByText, getByTestId, queryByText } = render(
-            withStoreAndRouter(<Header showNavbar={false} />)
```
arshinmar, line 16: hey, what was the rationale behind removing the "withStoreAndRouter" wrapper from all of the HTML elements? It seems like there was a new store/router function added, so why choose to remove "withStoreAndRouter" across the test files?

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
```diff
@@ -232,9 +233,9 @@ describe("Connected tables", () => {
...
     it("CheckedOutTable dispatches an action to toggle visibility when button clicked", async () => {
-        const { getByText } = render(
-            withStore(<CheckedOutTable items={itemsCheckedOut} />, store)
-        );
+        const { getByText } = render(<CheckedOutTable items={itemsCheckedOut} />, {
+            store,
+        });
         const button = getByText(/hide all/i);
 
         await fireEvent.click(button);
```
arshinmar, line 241: not sure how an "expectation" is set via the fireEvent.click function. Could you elaborate further?

hackathon_site/dashboard/frontend/src/testing/utils.tsx
```diff
@@ -0,0 +1,79 @@
...
+): Promise<T> =>
+    new Promise((resolve) => {
+        setTimeout(() => resolve(value), ms);
+    });
+
+export const makeMockApiListResponse = <T extends unknown>(
+    data: T[],
+    next?: string,
+    previous?: string
+): AxiosResponse<APIListResponse<T>> =>
```
arshinmar, line 71: wondering why you utilised an AxiosResponse instead of regular Promises?

arshinmar (commented):
An incredibly complex codebase that looks real interesting lol

grahamhoyes:
> I was wondering what's the convention for files having .ts vs. .tsx extensions. I thought maybe .tsx was for anything that renders to the DOM but unit tests are also .tsx so I'm not sure.

Anything that uses react JSX has to be in a `.tsx` file. With regular javascript you can put JSX in either `.jsx` or `.js` files (the latter is what we were doing), but not with typescript. Since we use JSX when rendering components in tests, they have to be `.tsx` as well.

hackathon_site/dashboard/frontend/src/api/types.ts
grahamhoyes: What do you mean "over the any type"?

Tying something as `foo: any` in typescript is not very helpful, since it doesn't tell you anything.

These generics (or templates as they would be called in C++) let us write a generic interface that we can make more specific by passing in additional type annotations. So something typed as with `response: APIListResponse<Hardware>` will know that `response.results` will be a hardware array, and if you try to access say `response.results[0].in_stock`, that will cause a type error since `in_stock` isn't a field on `Hardware`.

For a better explanation, read the typescript docs on generics: https://www.typescriptlang.org/docs/handbook/2/generics.html

hackathon_site/dashboard/frontend/src/api/types.ts
grahamhoyes: Union types!

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
grahamhoyes: So first, a state is pretty much just a javascript object.

A reducer is a function that takes in a state and an action, and returns an updated version of the state (reducers are a generic concept, they aren't redux-specific). In redux toolkit land, "Reducers" are a bit more all-encompasing, since it does also contain definitions for its state and actions - but that is a very Redux Toolkit thing.

And like you said, a selector takes in a state and returns something in that state (could be a specific field, could be an object within that state that has more fields).

Since `state: RootState` is basically just an object, all this function is doing is returning the slice of that object at `hardwareReducerName`. 

How that state object is created is wrapped up in redux abstractions for us. In `store.ts`, we define a map of reducer names to reducers (Redux Toolkit reducers, so wrappers for the actual reducers and functions); `configureStore()` then takes that object, and builds the initial state from it. How it actually does that I'm not sure, but you could dig into the redux toolkit source code to figure it out!

If you mouse over `hardwareSelectorState` in a good IDE, you should see that it's return type is

```typescript
 EntityState<Hardware> & HardwareExtraState
```

`EntityState` is defined by redux toolkit, and looks like this:

```typescript
export type EntityId = number | string

export interface EntityState<T> {
    ids: EntityId[]
    entities: Dictionary<T>
}
```

So, putting that all together, the return type of this function is our hardware state, which looks like

```typescript
interface HardwareState {
    ids: (number | string)[];
    entities: { 
        [key: string]: Hardware
    };
    isLoading: boolean;
    error: string | null;
    next: string | null;
    filters: HardwareFilters
}

```

hackathon_site/dashboard/frontend/src/slices/ui/uiSlice.js
grahamhoyes: The alternative would be to use state local to a component (with the `useState()` hook) - which you're right, we could use.

In this case, we use it so that it's persistent when the component is unmounted because the user changed pages. So if they minimize this table on their dashboard, then go look at the inventory list, then come back, it will stay minimized! With state local just to the table, it would be returned to the default state when you came back to the dashboard page.

hackathon_site/dashboard/frontend/src/components/general/Header/Header.test.js
grahamhoyes: Though it doesn't show up in the diff of this line, the `render` function is actually a different one now - notice at the top of the diff for pretty much all the test files, I import render from `testing/utils` now (in that file, it's defined as `customRender` and exported as `render`).

Since we are moving towards the recommended approach of testing all components with an actual redux store (as opposed to testing unconnected components like we used to), pretty much every render we do will need a store (and the router, which makes pages work). Since we'll need them everywhere, I just baked them in to the new custom `render` function.

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.js
grahamhoyes: These tests pretty much read like english. Quick note first though: this is an older test, so it is using a fake store object and not a real store (which is why I can call `store.getActions()`).

In this test we:
- Render the `CheckedOutTable` with a set of mocked items, and pass the render function a fake store from `redux-mock-store`
- Get the hide button with the `getByText` function
- Click said button
- Expect that the actions dispatched to the [mocked] store (`expect(store.getActions())`) contains an object that has a `type` of `toggleCheckedOutTable.type`.

This particular expect step is pretty dense and I'm basically crafting what the actual redux action looks like (again, in the new tests we won't do anything like this). 

To relate it to python, think of `expect(thing).to<something>` as `self.assert<whatever>` in python. Some simple comparisons:

- `self.assertEqual(a, b)`: `expect(a).toEqual(b)`
- `self.assertContains(a, "hi there")`: `expect(a).toContain("hi there")`
- `some_mock.assert_called_with(a, b)`: `expect(someMock).toHaveBeenCalledWith(a, b)`

hackathon_site/dashboard/frontend/src/testing/utils.tsx
grahamhoyes: Tldr; I wrap what this helper returns in a promise eventually. `AxiosResponse` itself is not a promise, it's just another type definition.

The return type of `get()` in `api/api.ts` is `Promise<AxiosResponse<T>>` (because that's what axios returns).

This helper function is just used to construct a list response from a set of items, so for example 

```typescript
const hardware: Hardware[] = [/* Define some hardware objects here */]
const mockListResponse = makeMockApiListResponse(hardware)
```

will give `mockListResponse` a type of `AxiosResponse<APIListResponse<Hardware>>` (note that we don't actually need to type annotate it, typescript is really good at inferring types).

Given that, the question is why return `AxiosResponse<...>` instead of `Promise<AxiosResponse<...>>`?

Because where I actually use this function, I eventually do wrap the return value in a promise.

For example, in the inventory grid tests:

```typescript
jest.mock("api/api");
const mockedGet = get as jest.MockedFunction<typeof get>;
...

// Inside some test
const apiResponse = makeMockApiListResponse(mockHardware);
mockedGet.mockResolvedValue(apiResponse);
```

`mockResolvedValue` just means "when something calls the get function, have it return a resolved promise with this value` - so that's where the promise wrapper is coming from. 

I could have also done (and used to before I discovered `mockResolvedValue`):

```typescript
mockedGet.mockReturnValue(Promise.resolve(apiResponse));
```

which is the same thing. Note if I just tried to return `apiResponse`, typescript will actually complain since it knows that `get` should be returning `Promise<AxiosResponse<...>>`, not just `AxiosResponse<...>`.

If I really wanted to manually resolve the promise, I could do that too:

```typescript
mockedGet.mockReturnValue(new Promise((resolve) => {
    resolve(apiResponse)
}));
```

If you take a look at `promiseResolveWithDelay` in utils.tsx and where I use it, I pretty much do exactly that - resolve the promise after a delay set in a `setTimeout`.

GauravRanganath (approved):
Hey went through all the changes and they look good to me :) Nice work @grahamhoyes!

PR #304, opened 2021-08-22
Leave Team API View

EdwardJTL:
## Overview

- Resolves IEEE-37
- Add API endpoint for leaving event team
- The request is served in an atomic action to prevent the case where another user joins the team while the team is getting deleted, which would cascade the delete
- Endpoint can only be accessed if user is signed in and already on a team
- User can leave team if there is no not-cancelled orders on the team
- User will be put on a new team
- If old team is now empty, delete the old team

## Unit Tests Created

- Test access
- Test leaving and creating team
- Test can only leave when orders are cancelled
- Test old team deleted if empty

## Steps to QA

- Make a profile for your user, call the endpoint
  - Should return a serialized team object for a new team
  - db should not have any other team
- Add another user to your team, call the endpoint
  - Should return a serialized team object for a new team
  - db should have both teams
- Add some orders to your team
  - Try different values for the status of the order
  - Call the endpoint
    - The request should only succeed when the order is cancelled
      - Check that the old team is no longer in the db, and that the orders are deleted as well
    - Otherwise, the request should return a 400


hackathon_site/event/views.py
```diff
@@ -200,3 +207,38 @@ def post(self, request, *args, **kwargs):
...
+
+    @transaction.atomic
+    @swagger_auto_schema(responses={201: TeamSerializer})
+    def post(self, request, *args, **kwargs):
+        profile = request.user.profile
+        team = profile.team
+
+        # Raise 400 if team has active orders
+        active_orders = OrderItem.objects.filter(
+            Q(part_returned_health__isnull=True),
```
grahamhoyes: Hmm do we want these two conditions? I would originally have thought if they had an order at all, lock them to the team.

I don't want there to be the case where they somehow leave the team after the event (when everything is returned), screwing up our metrics. Maybe only allow them to leave if no parts have actually been checked out (ie only cancelled orders allowed)? That would just be removing this first clause.

hackathon_site/event/views.py
```diff
@@ -200,3 +207,38 @@ def post(self, request, *args, **kwargs):
...
+            ~Q(order__status="Cancelled"),
+            Q(order__team=team),
+        )
+        if active_orders.exists():
+            return Response(
```
grahamhoyes: Since this is still an API view, can you craft this so it's the same shape as a normal DRF 400 response? That will let us use some helpers on the frontend to display the error with less edge cases.

You may be able to throw a DRF validation error to accomplish that.

grahamhoyes (commented):
Just some minor stuff, open to hearing your thoughts

hackathon_site/event/test_api.py
```diff
@@ -87,3 +88,106 @@ def test_user_has_profile(self):
...
+            manufacturer="manufacturer",
+            datasheet="/datasheet/location/",
+            quantity_available=4,
+            max_per_team=1,
+            picture="/picture/location",
+        )
+        order = Order.objects.create(status="Cart", team=self.team)
+        OrderItem.objects.create(order=order, hardware=hardware)
+
+        for status_choice in Order.STATUS_CHOICES:
```
grahamhoyes: Did you mean to change the status of the order inside this loop?

hackathon_site/event/test_api.py
```diff
@@ -87,3 +88,106 @@ def test_user_has_profile(self):
...
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.data["id"], self.user.profile.team.pk)
+        self.assertNotEqual(old_team.pk, self.user.profile.team.pk)
+
+        self.assertEqual(old_team.pk, other_user.profile.team.pk)
+        self.assertEqual(
+            Team.objects.filter(team_code=old_team.team_code).exists(), True
+        )
+
+    def check_leave_and_delete(self):
```
grahamhoyes: Minor, but move these two helper functions above the actual test cases

hackathon_site/event/test_api.py
```diff
@@ -87,3 +88,106 @@ def test_user_has_profile(self):
...
+
+    def test_user_has_no_profile(self):
+        self.profile.delete()
+        self._login()
+        response = self.client.post(self.view)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_leave_and_delete(self):
+        """
+            when there is no orders or other members on the team,
```
grahamhoyes: Minor, don't indent the docstrings to be consistent with the ones elsewhere. Same for the ones below

hackathon_site/event/test_api.py
grahamhoyes: `status_choice` is also a tuple so you probably want to be doing:

```python
for _, status_choice in Order.STATUS_CHOICES:
    ...
```

hackathon_site/event/api_urls.py
```diff
@@ -7,4 +7,5 @@
 urlpatterns = [
     path("users/user/", api_views.CurrentUserAPIView.as_view(), name="current-user"),
     path("teams/team/", views.CurrentTeamAPIView.as_view(), name="current-team"),
+    path("event/leave_team/", views.LeaveTeamView.as_view(), name="leave-team"),
```
grahamhoyes: This actually makes the path `/api/event/event/leave_team/`.

How about we make it `/api/event/teams/leave/`, now that I'm realizing there's a teams namespace. So that would be `path("teams/leave/, ...)` here

hackathon_site/event/views.py
```diff
@@ -200,3 +208,36 @@ def post(self, request, *args, **kwargs):
         at once.
         """
         return super().post(request, *args, **kwargs)
+
+
+class LeaveTeamView(generics.GenericAPIView):
```
grahamhoyes: Because this doesn't have a `serializer_class` property or `get_serializer_class` method, the endpoint will crash (500) if you try to open it in the browser. 

Crashing isn't good, so you could do `serializer_class = TeamSerializer`. Then if you submit a get request, you'll get a 405 - Method not Allowed back.

That will make your `swagger_auto_schema` unnecessary, since the 201 will be annotated properly. The downside is that DRF / Swagger thinks the endpoint needs a team code to be in the request body, which isn't the case. But the API still works fine without it.

grahamhoyes: This should be in `event/api_views.py`

grahamhoyes (approved):
LGTM!

PR #305, opened 2021-08-30
ieee-36: POST: /api/event/teams/join/<team_code>/

arshinmar:
## Overview

- Resolves #305 
- Added a JoinTeamAPIView
- Added corresponding tests
- Added a "create event team" functionality with a customizable user number

## Unit Tests Created

- Check the JoinTeamTestCase in event/test_api.py

## Steps to QA

- @grahamhoyes you know best...


hackathon_site/event/api_urls.py
```diff
@@ -7,5 +7,10 @@
 urlpatterns = [
     path("users/user/", api_views.CurrentUserAPIView.as_view(), name="current-user"),
     path("teams/team/", views.CurrentTeamAPIView.as_view(), name="current-team"),
```
grahamhoyes, line 9: I just noticed that this view is in `views` instead of `api_views`, whoops. Will make a ticket to fix that

hackathon_site/event/api_views.py
```diff
@@ -1,10 +1,25 @@
 from django.db import transaction
 from django.db.models import Q
 
+from event.models import User, Team
```
grahamhoyes: We separate our imports into 3 sections:
- Standard library imports
- 3rd party imports (django, rest_framework, etc)
- Local imports (event, hardware, etc)

Can you rearrange these so that everything is together properly?

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
         response_data = response_serializer.data
         return Response(data=response_data, status=status.HTTP_201_CREATED,)
+
+
+class JoinTeamView(generics.GenericAPIView, mixins.RetrieveModelMixin):
+    permission_classes = [UserHasProfile]
+    serializer_class = TeamSerializer
+
+    def clean(self):
+        if not is_registration_open():
```
grahamhoyes: This is for the event, not registration.

From the ticket description, the actual check we need to have:

If the team they are currently on has any orders, HTTP 400 and don’t allow them to leave, with a message in the response.

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
         response_serializer = TeamSerializer(profile.team)
         response_data = response_serializer.data
         return Response(data=response_data, status=status.HTTP_201_CREATED,)
+
+
+class JoinTeamView(generics.GenericAPIView, mixins.RetrieveModelMixin):
+    permission_classes = [UserHasProfile]
+    serializer_class = TeamSerializer
+
+    def clean(self):
```
grahamhoyes: I don't think that `clean` is a method used by DRF views, it's a django form method

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+        if not is_registration_open():
+            raise forms.ValidationError(
```
grahamhoyes: Is your IDE properly configured, and did you test this? `forms` is never imported.

Nor should it be - look at how the `LeaveTeamView` raises a `ValidationError` (which is a DRF error, different from `forms.ValidationError`)

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+
+    @transaction.atomic
+    def post(self, request, team_code, *args, **kwargs):
+
+        profile = request.user.profile
+        current_team = profile.team
+
+        try:
+            team = Team.objects.get(team_code=team_code)
+        except:
```
grahamhoyes: It's bad practice to have a generic `except`, since it may swallow errors that we haven't thought about (your IDE should be warning you of this).

Instead, do `except Team.DoesNotExist`, which is the error that will be thrown if the team doesn't exist.

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+            raise ValidationError(
```
grahamhoyes: When something is not found, in most cases (including this case) we want to return a HTTP 404, not 400. 

You could do that by raising an `Http404` error from `django.http`, or constructing a drf `Response` with a 404 status code.

My recommendation though: the generic API view has a method called `get_object()` which will automatically look up the object for you and 404 if necessary. By defult it uses the `pk` argument on a view, but you can set it to filter on team code by adding `lookup_field = "team_code"` to your view class.

[Here's](https://github.com/encode/django-rest-framework/blob/master/rest_framework/generics.py#L75) the source code of `get_object()` if you're curious.

hackathon_site/event/api_views.py
grahamhoyes: Looks like you have a lot of duplicate imports in here too

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+                code="registration_closed",
+            )
+
+        return super().clean()
+
+    @transaction.atomic
+    def post(self, request, team_code, *args, **kwargs):
+
+        profile = request.user.profile
+        current_team = profile.team
```
grahamhoyes, line 81: Unused, that's concerning

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+
+        try:
+            team = Team.objects.get(team_code=team_code)
+        except:
+            raise ValidationError(
+                {"detail": "Team does not exist!"}, code=status.HTTP_400_BAD_REQUEST
```
grahamhoyes: We don't use "!" in our error responses

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+            )
+
+        if not team.profiles.exists():
+            raise ValidationError(
+                {"detail": "Cannot join a fully empty team!"},
```
grahamhoyes: Why not? Imo this isn't a check we need since there's no harm in joining an empty team. It shouldn't happen anyway since we clean up teams when they get empty.

hackathon_site/event/api_views.py
grahamhoyes: There's also code for that as part of the `LeaveTeamView` already

grahamhoyes: Ok, all of the checks for this are part of the post method already. So why is this method still here?

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+            ~Q(order__status="Cancelled"), Q(order__team=team),
+        )
+
+        if active_orders.exists():
+            raise ValidationError(
+                {"detail": "Cannot join a team with already processed orders"},
+                code=status.HTTP_400_BAD_REQUEST,
+            )
+
+        print(type(profile.team))
```
grahamhoyes: remove

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+
+        if not team.profiles.exists():
+            raise ValidationError(
+                {"detail": "Cannot join a fully empty team!"},
+                code=status.HTTP_400_BAD_REQUEST,
+            )
+        if team.profiles.count() >= Team.MAX_MEMBERS:
+            raise ValidationError({"detail": "Team is full"})
+
+        # Raise 400 if team has active orders
```
grahamhoyes: We don't want to let them leave a team if the team they are on has active orders (which is what the ticket says) - I don't see any reason to not let them join a team with orders, they just won't be able to leave it once they join

hackathon_site/event/api_views.py
```diff
@@ -66,3 +81,57 @@ def post(self, request, *args, **kwargs):
...
+        )
+
+        if active_orders.exists():
+            raise ValidationError(
+                {"detail": "Cannot join a team with already processed orders"},
+                code=status.HTTP_400_BAD_REQUEST,
+            )
+
+        print(type(profile.team))
+        profile.team = team
```
grahamhoyes, line 97: You haven't handled cleaning up their old team at all

hackathon_site/event/test_api.py
```diff
@@ -52,42 +52,77 @@ def test_user_has_profile(self):
         self.assertEqual(response.json(), serializer.data)
 
 
-class CurrentTeamTestCase(SetupUserMixin, APITestCase):
```
grahamhoyes: ... so you deleted an entire view's tests? Why?

hackathon_site/event/test_api.py
```diff
@@ -52,42 +52,77 @@ def test_user_has_profile(self):
...
+        self.view_name = "api:event:join-team"
 
-        self.view = reverse("api:event:current-team")
+    def _build_view(self, team_code):
+        return reverse(self.view_name, kwargs={"team_code": team_code})
+
+    def test_invalid_key(self):
+        self._login()
+        response = self.client.post(self._build_view("56ABD"))
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
grahamhoyes: Should be 404

hackathon_site/event/test_api.py
```diff
@@ -52,42 +52,77 @@ def test_user_has_profile(self):
...
+
+    def test_join_full_team(self):
+        self._login()
+        team = self._make_full_event_team(self_users=False)
+        print(team.team_code)
```
grahamhoyes: remove

hackathon_site/event/test_api.py
```diff
@@ -52,42 +52,77 @@ def test_user_has_profile(self):
...
+        response = self.client.post(self._build_view(team.team_code))
+        self.assertEqual(str(response.content), 'b\'{"detail":"Team is full"}\'')
```
grahamhoyes: Look at some of our other API tests. You can do `data = response.json()` to get a dict, then compare that dict rather than an ugly string like this

hackathon_site/event/test_api.py
```diff
@@ -52,42 +52,77 @@ def test_user_has_profile(self):
...
-    def test_user_has_profile(self):
+    def check_cannot_join(self):
+        old_team = self.profile.team
+        response = self.client.post(self._build_view(self.team.team_code))
+        self.user.refresh_from_db()
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(old_team.pk, self.user.profile.team.pk)
+        self.assertEqual(Team.objects.filter(team_code=old_team.team_code).count(), 1)
+
+    def test_leave_with_order(self):
```
grahamhoyes: this isn't the leave team view

hackathon_site/hackathon_site/tests.py
```diff
@@ -25,6 +26,20 @@ def setUp(self):
     def _login(self):
         self.client.login(username=self.user.username, password=self.password)
 
+    @staticmethod
+    def _apply_to_profile_as_user(user, team=None):
```
grahamhoyes: Why `_apply_to_profile`? What does this have to do with applying? Perhaps `_make_profile` would be a better name

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,73 +77,134 @@ def _get_random_email():
...
-    def _make_full_registration_team(self, team=None, self_users=True):
+    def _make_full_registration_team(self, team=None, self_users=True, num_users=4):
         if team is None:
             team = RegistrationTeam.objects.create()
 
         if self_users:
+
             user1 = self.user
+            cumulative_users = [user1]
+            if num_users > 1:
```
grahamhoyes: You can come up with a better way to write this using a for loop.

I don't care about using the painter's names either (unless we use them elsewhere in tests), you can just generate random values for all the fields.

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,73 +77,134 @@ def _get_random_email():
...
+            email1 = self._get_random_email()
+            user1 = User.objects.create_user(
+                username=email1,
+                password="foobar123",
+                email=email1,
+                first_name="John1",
+                last_name="Doe1",
+            )
+            cumulative_users = [user1]
+            if num_users > 1:
```
grahamhoyes: Same, if you want to make this configurable, this one is really easy to make using a for loop

hackathon_site/event/test_api.py
arshinmar: for some reason, I can't seem to resolve this issue. Will push code soon. The views code seems to make sense

hackathon_site/event/test_api.py
arshinmar: sorry about that

hackathon_site/event/api_urls.py
arshinmar: ok, sounds good

hackathon_site/event/api_views.py
arshinmar: resolved in new code

hackathon_site/event/api_views.py
arshinmar: removed in new code, part of edward's view, so thought it would be worth initially taking

hackathon_site/event/api_views.py
arshinmar: removed

hackathon_site/event/api_views.py
arshinmar: removed

hackathon_site/event/api_views.py
arshinmar: resolved in new code

hackathon_site/event/api_views.py
arshinmar: added get_object functionality in new code

hackathon_site/event/api_views.py
arshinmar: utilised in new code

hackathon_site/event/api_views.py
arshinmar: fixed

hackathon_site/event/api_views.py
arshinmar: removed

hackathon_site/event/api_views.py
arshinmar: done

hackathon_site/event/api_views.py
arshinmar: now, members cannot join another team if current team has active orders, as per expectation/description of JIRA issue

hackathon_site/event/api_views.py
arshinmar: now, members cannot join another team if current team has active orders, as per expectation/description of JIRA issue

if old team only had one member (themselves), the team is deleted.

hackathon_site/event/test_api.py
arshinmar: resolved now

hackathon_site/event/test_api.py
arshinmar: done

hackathon_site/event/test_api.py
arshinmar: done

hackathon_site/event/test_api.py
arshinmar: changed up the hardware consideration as per above comments.

hackathon_site/hackathon_site/tests.py
arshinmar: changed

hackathon_site/hackathon_site/tests.py
arshinmar: fixed

hackathon_site/hackathon_site/tests.py
arshinmar: done with a for loop

hackathon_site/event/api_views.py
```diff
@@ -1,3 +1,4 @@
+from event.models import Team
```
grahamhoyes: Already imported as `EventTeam` below

hackathon_site/event/api_views.py
```diff
@@ -66,3 +67,41 @@ def post(self, request, *args, **kwargs):
         response_serializer = TeamSerializer(profile.team)
         response_data = response_serializer.data
         return Response(data=response_data, status=status.HTTP_201_CREATED,)
+
+
+class JoinTeamView(generics.GenericAPIView, mixins.RetrieveModelMixin):
+    permission_classes = [UserHasProfile]
+    serializer_class = TeamSerializer
+    lookup_field = "team_code"
+    queryset = Team
```
grahamhoyes:
```suggestion
    queryset = EventTeam.objects.all()
```

hackathon_site/event/api_views.py
```diff
@@ -66,3 +67,41 @@ def post(self, request, *args, **kwargs):
...
+
+    @transaction.atomic
+    def post(self, request, team_code, *args, **kwargs):
+
+        profile = request.user.profile
+        current_team = profile.team
+
+        self.kwargs["lookup_url_kwarg"] = team_code
```
grahamhoyes: `lookup_url_kwarg` is the URL keyword argument that gets used in filtering. It basically does:

```python
team = EventTeam.objects.filter(**{self.lookup_field: self.kwargs[self.lookup_url_kwarg]})
```

Which we want to be

```python
EventTeam.objects.filter(team_code=team_code)
```

`self.lookup_url_kwarg` will default to `self.lookup_field`, so you don't need to do this. This is setting it on self.kwargs anyway, which is doesn't do anything.

hackathon_site/event/api_views.py
```diff
@@ -66,3 +67,41 @@ def post(self, request, *args, **kwargs):
...
+        team = self.get_object()
+
+        if team.profiles.count() >= Team.MAX_MEMBERS:
```
grahamhoyes:
```suggestion
        if team.profiles.count() >= EventTeam.MAX_MEMBERS:
```

hackathon_site/event/api_views.py
```diff
@@ -66,3 +67,41 @@ def post(self, request, *args, **kwargs):
...
+
+        profile.team = team
+        profile.save()
+
+        if not current_team.profiles.exists():
+            current_team.delete()
+
+        response_serializer = TeamSerializer(profile.team)
+        response_data = response_serializer.data
+        return Response(data=response_data, status=status.HTTP_201_CREATED,)
```
grahamhoyes: Response status here can just be 200 OK

hackathon_site/event/test_api.py
```diff
@@ -90,6 +90,102 @@ def test_user_has_profile(self):
         self.assertEqual(response.json(), serializer.data)
 
 
+class JoinTeamTestCase(SetupUserMixin, APITestCase):
+    def setUp(self):
+        super().setUp()
+        self.group = Group.objects.create(name="Test Users")
```
grahamhoyes: Unnecessary

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,74 +76,67 @@ def _get_random_email():
...
+                    username=random_email,
+                    password="foobar123",
+                    email=random_email,
+                    first_name="John{}".format(user_number),
+                    last_name="Doe{}".format(user_number),
+                )
+                self._apply_as_user(new_user, team)
+        return team
 
+    def _make_full_event_team(self, team=None, self_users=True, num_users=4):
```
grahamhoyes: Nit: Since this isn't making a full team, can just call it `_make_event_team`. Same for the registration one above, if you use pycharm to rename the method (right click > refactor > rename) it should update everywhere.

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,74 +76,67 @@ def _get_random_email():
             email = f"{uuid[:10]}@{uuid[10:20]}.com"
         return email
 
-    def _make_full_registration_team(self, team=None, self_users=True):
+    def create_user_set(self):
```
grahamhoyes:
```suggestion
    def _create_user_set(self):
```

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,74 +76,67 @@ def _get_random_email():
...
-                last_name="Carmichael",
-            )
-            user4 = self.user4 = User.objects.create_user(
-                username="lawren@harris.com",
-                password="wxyz7890",
-                email="lawren@harris.com",
-                first_name="Lawren",
-                last_name="Harris",
-            )
+        if self_users == True:
```
grahamhoyes:
```suggestion
        if self_users:
```

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,74 +76,67 @@ def _get_random_email():
...
-                email=email4,
-                first_name="John4",
-                last_name="Doe4",
-            )
-
-        self._apply_as_user(user1, team)
-        self._apply_as_user(user2, team)
-        self._apply_as_user(user3, team)
-        self._apply_as_user(user4, team)
+            for user_number in range(1, num_users + 1, 1):
```
grahamhoyes:
```suggestion
            for user_number in range(1, num_users + 1):
```

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,74 +76,67 @@ def _get_random_email():
...
+                random_email = self._get_random_email()
+                new_user = User.objects.create_user(
+                    username=random_email,
+                    password="foobar123",
+                    email=random_email,
+                    first_name="John{}".format(user_number),
```
grahamhoyes: Slightly more condensed:

```suggestion
                    first_name=f"John{user_number}",
```

Same with other format strings below

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,74 +76,67 @@ def _get_random_email():
...
+                    last_name="Doe{}".format(user_number),
+                )
+                self._apply_as_user(new_user, team)
+        return team
 
+    def _make_full_event_team(self, team=None, self_users=True, num_users=4):
+        if team is None:
+            team = EventTeam.objects.create()
+        if self_users == True:
```
grahamhoyes:
```suggestion
        if self_users:
```

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,74 +76,67 @@ def _get_random_email():
...
+            for user in self.create_user_set():
+                self._apply_as_user(user, team)
+        else:
+            for user_number in range(1, num_users + 1, 1):
```
grahamhoyes:
```suggestion
            for user_number in range(1, num_users + 1):
```

hackathon_site/hackathon_site/tests.py
```diff
@@ -62,74 +76,67 @@ def _get_random_email():
...
+                random_email = self._get_random_email()
+                new_user = User.objects.create_user(
+                    username=random_email,
+                    password="foobar123",
+                    email=random_email,
+                    first_name="John{}".format(user_number),
```
grahamhoyes:
```suggestion
                    first_name=f"John{user_number}",
```

hackathon_site/event/test_api.py
```diff
@@ -90,6 +90,102 @@ def test_user_has_profile(self):
...
+    def _build_view(self, team_code):
+        return reverse(self.view_name, kwargs={"team_code": team_code})
+
+    def test_join_and_delete(self):
+        """
+        When a member wants to join a team, if their current team only includes
+        them, their current team is deleted.
+        """
+        self._login()
+        self.client.post(self._build_view(self.team.team_code))
```
grahamhoyes: Instead of using a post request to put the user on a team the first time, can you just set it up directly from a model query? Helps keep this test isolated in what it is testing (that leaving an empty team deletes it)

hackathon_site/event/test_api.py
```diff
@@ -90,6 +90,102 @@ def test_user_has_profile(self):
...
+        sample_team = self._make_full_event_team(self_users=False, num_users=2)
+        response = self.client.post(self._build_view(sample_team.team_code))
+        self.user.refresh_from_db()
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.data["id"], self.user.profile.team.pk)
+        self.assertNotEqual(old_team.pk, self.user.profile.team.pk)
+        self.assertEqual(
+            Team.objects.filter(team_code=old_team.team_code).exists(), False
+        )
```
grahamhoyes:
```suggestion
        self.assertFalse(
            Team.objects.filter(team_code=old_team.team_code).exists()
        )
```

hackathon_site/event/api_views.py
```diff
@@ -66,3 +67,41 @@ def post(self, request, *args, **kwargs):
...
+    def post(self, request, team_code, *args, **kwargs):
+
+        profile = request.user.profile
+        current_team = profile.team
+
+        self.kwargs["lookup_url_kwarg"] = team_code
+        team = self.get_object()
+
+        if team.profiles.count() >= Team.MAX_MEMBERS:
+            raise ValidationError({"detail": "Team is full."})
```
grahamhoyes: Minor for consistency:

```suggestion
            raise ValidationError({"detail": "Team is full"})
```

grahamhoyes (commented):
Much better!

hackathon_site/event/api_views.py
```diff
@@ -66,3 +66,38 @@ def post(self, request, *args, **kwargs):
...
+
+
+class JoinTeamView(generics.GenericAPIView, mixins.RetrieveModelMixin):
+    permission_classes = [UserHasProfile]
+    serializer_class = TeamSerializer
+    lookup_field = "team_code"
+    queryset = EventTeam.objects.all()
+
+    @transaction.atomic
+    def post(self, request, team_code, *args, **kwargs):
```
grahamhoyes: Can actually remove `team_code` from here since it's unused (consumed implicitly by `get_object()` now)

hackathon_site/event/views.py
```diff
@@ -7,15 +7,20 @@
 from django.urls import reverse_lazy
 from django.views.generic.base import TemplateView
 from django.views.generic.edit import FormView
+from django.conf import settings
```
grahamhoyes, line 9: Take a look at this file again, lots of unused or duplicated imports (I'm guessing from starting working in this file before switching to api_views.py)

hackathon_site/hackathon_site/tests.py
grahamhoyes: Still have an unnecessary `,1` at the end (that's obvious for `range()`)

grahamhoyes (commented):
Just two minor things, then looks good!

grahamhoyes (approved):

PR #308, opened 2021-09-11
Ieee 69 get /event/teams

Leo6Leo:
## Overview

- Uses event.models.Team, NOT registration.models.Team

- Call it TeamListView, url route name team-list

- Permissions: event.teams.view ( ) 

- DjangoModelPermissions doesn’t actually check for view permissions on get requests, so do some googling (see comment below actually)

- Return a list of teams. We already have a team serializer, this should be easy.

- Use the team serializer that will be improved by 

   - Filters:

      - Search:

         - Team ID

         - Team code

         - team__user__first_name

         - team__user__last_name

## Unit Tests Created

- Not created yet because the new Team Serializer hasn't been developed yet. Refer to [IEEE-70 on Jira](https://ieeeuoft.atlassian.net/browse/IEEE-70)

## Steps to QA

- Create some applications and teams through the admin site, and play with them!!


hackathon_site/event/api_filters.py
```diff
@@ -0,0 +1,36 @@
...
+class IntegerCSVFilter(filters.BaseInFilter):
+    field_class = CSVInputIntegerField
+
+
+class TeamFilter(filters.FilterSet):
+    queryset = EventTeam
+    serializer_class = TeamSerializer
+
+    @staticmethod
+    def filter_in_stock(queryset, _, value):
```
grahamhoyes: Be careful to read what you copy-paste :stuck_out_tongue:. This should go

hackathon_site/event/views.py
```diff
@@ -200,3 +205,20 @@ def post(self, request, *args, **kwargs):
         at once.
         """
         return super().post(request, *args, **kwargs)
+
+
+class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
+    queryset = EventTeam.objects.all()
+    serializer_class = TeamSerializer
+
```
grahamhoyes, line 217: Now you need to attach the permissions class you defined:

```
permissions_classes = [FullDjangoModelPermissions]
```

grahamhoyes (commented):
Be careful about putting `/` in branch names - it's usually used to distinguish things like features, hotfixes, developer, etc. and can sometimes cause issues.

hackathon_site/event/views.py
```diff
@@ -7,16 +7,21 @@
...
 from django.conf import settings
+from django_filters import rest_framework as filters
+
 
 from hackathon_site.utils import is_registration_open
 from registration.forms import JoinTeamForm
 from registration.models import Team as RegistrationTeam
 
 from rest_framework import generics, mixins
+from rest_framework.filters import SearchFilter, OrderingFilter
```
YuYing-Liang: Where do you use OrderingFilter? (I'm only asking because I don't really know how to use it lol)

hackathon_site/event/views.py
Leo6Leo: @grahamhoyes  Graham, actually I have a question regarding this permission which I forgot to ask you in our last meeting. What is the purpose of this permission? The description in Jira said this "DjangoModelPermissions doesn’t actually check for view permissions on get requests, so do some googling". But I have tried that If I logged out under the case without adding this permission group, I will not be able to use get request. The page will say "not authorized". I cannot see any difference before and after adding this permission group.

hackathon_site/event/views.py
Leo6Leo: @YuYing-Liang  Ohh it's my mistake. When I copy and paste this line from my previous pull request, I forgot to check and remove this. stupid and awkward mistake again lol

Leo6Leo: Thanks Yuying for pointing that out!

Leo6Leo:
@grahamhoyes  I met the conflicts again.. I remembered that you told me I could use the conflict resolver in the pycharm when I met the same issue in [this PR](https://github.com/ieeeuoft/hackathon-template/pull/288#issuecomment-890399971). But I haven't find the conflict in the pycharm. Only the github is saying "this branch has conflicts".

hackathon_site/event/views.py
grahamhoyes: The DRF `DjangoModelPermissioms` doesn't check any permissions for GET requests, but the one that you've added does.

But now to test it, you'll need to give yourself that permission! On the admin site, edit your user and look for a permission like `Event | Team | View Team` or something like that and add it to yourself. You should be able to view the API after that. 

You'll also need tests for that. When not logged in, you should get a 401 Unauthorized. When logged in without that permission, you should get a 403 Forbidden. Finally with that permission, everything should work.

If you look up Django's permission system, you'll find how you can query and add permissions to a user. I recommend modifying the `_login` test method to take a list of permissions that it will add to the user. Then on your test, you can do something like `self.permisaions = <query for event.event_view permissions>`, and pass those to the login call.

Let me know if you need any help with that!

grahamhoyes:
> @grahamhoyes I met the conflicts again.. I remembered that you told me I could use the conflict resolver in the pycharm when I met the same issue in [this PR](https://github.com/ieeeuoft/hackathon-template/pull/288#issuecomment-890399971). But I haven't find the conflict in the pycharm. Only the github is saying "this branch has conflicts".

You have to start a merge first for the merge conflicts to appear.

```bash
$ git checkout develop
$ git pull
$ git checkout <this branch>
$ git merge develop
```

Then you should get the conflicts. After resolving with PyCharm, do `git commit` and you should be good to push.

hackathon_site/dashboard/frontend/yarn.lock
```diff
@@ -9028,9 +9028,9 @@ path-key@^3.0.0, path-key@^3.1.0:
   integrity sha512-ojmeN0qd+y0jszEtoY48r0Peq5dwMEkIlCOu6Q5f41lfkswXuKtYrhgoTpLnyIcHm24Uhqx+5Tqm2InSwLhE6Q==
 
 path-parse@^1.0.6:
```
grahamhoyes: I think when you were resolving merge conflicts, you kept your version of this file instead of the one from develop. Can you restore this file back to what's on develop?

```
$ git checkout develop -- hackathon_site/dashboard/frontent/yarn.lock
```

hackathon_site/requirements.txt
```diff
@@ -24,7 +24,7 @@ Jinja2==2.11.3
 MarkupSafe==1.1.1
 packaging==20.4
 pathspec==0.8.0
-Pillow==8.3.2
```
grahamhoyes: Same as yarn.lock, restore this back to what's on develop

hackathon_site/event/test_api.py
```diff
@@ -193,3 +194,84 @@ def test_leave_with_not_empty_team(self):
...
+
+        self.view = reverse("api:event:team-list")
+
+    def _login(self):
+        self.client.login(username=self.user.username, password=self.password)
+
+    def _login_permission(self):
+        self.client.login(username=self.user.username, password=self.password)
+        self.user.user_permissions.add(
+            Permission.objects.get(codename="view_team", pk="28")
```
grahamhoyes: Getting the permission by its primary key (the codename argument doesn't do anything when you specify the pk) isn't great, since it could change if the order permissions are created at changes (that's a detail internal to django).

You can filter by content type to get the app name instead, which will be more robust:

```python
permissions = Permission.objects.get(content_type__app_label="event", codename="view_team")
```

hackathon_site/event/test_api.py
```diff
@@ -193,3 +194,84 @@ def test_leave_with_not_empty_team(self):
...
+        self.team = Team.objects.create()
+        self.team2 = Team.objects.create()
+        self.team3 = Team.objects.create()
+
+        self.view = reverse("api:event:team-list")
+
+    def _login(self):
+        self.client.login(username=self.user.username, password=self.password)
+
+    def _login_permission(self):
```
grahamhoyes: We're going to need this in a lot of tests from now on, so it would be good to add it to our setup user mixin as part of the login function. Could you modify the `_login` method of `SetupUserMixin` to be:

```python
def _login(self, permissions=None):
    if permissions is not None:
        self.user.permissions.add(permissions)

```

Then to use it, in `setUp` you can set 

```python
self.permissions = Permission.objects.filter(content_type__app_label=...)
```

and call login in your later tests by doing `self._login(self.permissions)`

hackathon_site/event/test_api.py
```diff
@@ -193,3 +194,84 @@ def test_leave_with_not_empty_team(self):
...
+        self.user.user_permissions.add(
+            Permission.objects.get(codename="view_team", pk="28")
+        )
+
+    def _build_filter_url(self, **kwargs):
+        return (
+            self.view + "?" + "&".join([f"{key}={val}" for key, val in kwargs.items()])
+        )
+
+    def test_team_get_no_permission(self):
```
grahamhoyes: Still need a test for not logged in as well, which should have a status of HTTP 401 Unauthorized

hackathon_site/event/test_api.py
```diff
@@ -193,3 +194,84 @@ def test_leave_with_not_empty_team(self):
...
+        self._login()
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_team_get_permission(self):
```
grahamhoyes: nit: `test_team_get_has_permissions`, since it's not the permission we're getting

hackathon_site/hackathon_site/tests.py
```diff
@@ -22,7 +22,11 @@ def setUp(self):
             email="foo@bar.com",
         )
 
-    def _login(self):
+    def _login(self, permissions=None):
+        if permissions is not None:
+            for permission in permissions:
```
grahamhoyes, line 27: Does this work if you just do `self.user.permissions.add(permissions)`? `permissions` should be a queryset, so it should be able to add them all in a single query. Will help speed up our tests a tiny amount.

hackathon_site/event/views.py
```diff
@@ -1,21 +1,29 @@
...
+
+from django.conf import settings
+from django_filters import rest_framework as filters
+
 
 from hackathon_site.utils import is_registration_open
 from registration.forms import JoinTeamForm
 from registration.models import Team as RegistrationTeam
 
+from rest_framework import generics, mixins
```
grahamhoyes: Minor (just noticed this), can you move this to below the django imports so 3rd party imports are all grouped together?

grahamhoyes (commented):
Two minor things, looking good!

hackathon_site/event/permissions.py
```diff
@@ -12,3 +15,15 @@ def has_permission(self, request, view):
             and request.user
             and hasattr(request.user, "profile")
         )
+
+
+class FullDjangoModelPermissions(DjangoModelPermissions):
```
YuYing-Liang, line 20: maybe this should be names `FullDjangoModelViewPermissions` because I think we'll have to make another for change permissions right?

YuYing-Liang (commented):
Nice job Leo!

hackathon_site/hackathon_site/tests.py
Leo6Leo: I have tried this. for .add(), it only accepts 1 variable at a time. If it got more than 2 querysets (when I use Permission.objects.select(), it will just through out an error), I can attach the screenshot here later

Leo6Leo: so I add a for loop there, and the error is solved

hackathon_site/event/permissions.py
grahamhoyes: Nope, everything else is covered by `DjangoModelPermissions` (all the "unsafe" methods, so add / change / delete). All methods are covered by this `FullDjangoModelPermissions`

hackathon_site/hackathon_site/tests.py
grahamhoyes: Ah my bad, you're absolutely right!

grahamhoyes (approved):
Good work!

PR #309, opened 2021-09-16
IEEE-67: Created filtering for /hardware/orders api

YuYing-Liang:
## Overview

- Resolves #IEEE-67: https://ieeeuoft.atlassian.net/browse/IEEE-67 
- Added filtering for team_id, team_code and status for /hardware/orders api

## Unit Tests Created

- added unit tests for each filter and updated serializer tests to include team_code

## Steps to QA

## Problems I don't know how to solve

- I tried testing the ordering filter for created_at date but the date I gave to the Order model didn't work (but I don't know why)
- I couldn't find anywhere in django filters library documentation how to filter by descending
- not sure if there's another way to filter by team_code without putting the team_code in the order serializer/response (this was only needed for unit tests)


hackathon_site/hardware/permissions.py
```diff
@@ -0,0 +1,6 @@
+from rest_framework.permissions import DjangoModelPermissions
+
+
+class UserHasHardareOrderViewPermission(DjangoModelPermissions):
+    def __init__(self):
```
YuYing-Liang: Not quite sure how this works 😬, I followed a medium tutorial: https://medium.com/innoventes/django-protect-apis-using-djangomodelpermissions-cb1bf65c5b58

hackathon_site/hardware/permissions.py
```diff
@@ -0,0 +1,6 @@
+from rest_framework.permissions import DjangoModelPermissions
+
+
+class UserHasHardareOrderViewPermission(DjangoModelPermissions):
```
grahamhoyes: As we chatted about today, [here's](https://github.com/ieeeuoft/hackathon-template/pull/308/files#diff-a22107fd28a4dd92bbe5ad30046e848b0684042881843d5fe0fb2ee3c076cd7fR20) where this is implemented in @Leo6Leo's PR and will be available soon!

Permissions like this are model-independent, so we can just have one for the entire project

hackathon_site/hardware/serializers.py
```diff
@@ -41,10 +41,23 @@ def get_unique_hardware_count(obj: Category) -> int:
...
 class OrderListSerializer(serializers.ModelSerializer):
     hardware_set = HardwareSerializer(many=True, read_only=True)
+    team_code = serializers.SerializerMethodField()
 
     class Meta:
         model = Order
-        fields = ("id", "hardware_set", "team", "status", "created_at", "updated_at")
+        fields = (
+            "id",
+            "hardware_set",
```
grahamhoyes: for the API, we can break with what the field is called in django and just call this `hardware`. I realize this wasn't something you changed, I'm just notorious for noticing things too late :)

hackathon_site/hardware/serializers.py
```diff
@@ -41,10 +41,23 @@ def get_unique_hardware_count(obj: Category) -> int:
 
 class OrderListSerializer(serializers.ModelSerializer):
     hardware_set = HardwareSerializer(many=True, read_only=True)
+    team_code = serializers.SerializerMethodField()
 
     class Meta:
         model = Order
-        fields = ("id", "hardware_set", "team", "status", "created_at", "updated_at")
+        fields = (
```
grahamhoyes, line 77: The details below should go in a separate PR and Ticket to update the serializer to keep this PR clean, but we should think about what other information we'll want to capture as part of the order response.

We're probably going to use this endpoint on the participant side to get the orders to show on their dashboard. We're going to need to create a serializer for the `OrderItem`s and a new hardware serializer to compute and return that information. Going off Figma, we'll need things like:
- Quantity requested (I don't actually think we store this in the database, so maybe not)
- Quantity granted (will need some group bys in the queryset to compute efficiently)
- Whether it's been returned
- Returned time (looks like we don't store that in the database either? Should add a field for it to OrderItem)
- Part returned health

hackathon_site/hardware/test_api.py
```diff
@@ -376,6 +391,59 @@ def test_hardware_get_success(self):
...
+        self._login()
+
+        url = self._build_filter_url(team_id="1")
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = response.json()
+        results = data["results"]
+
+        returned_ids = [res["team"] for res in results]
```
grahamhoyes: Though this does work, you should be doing the assertions at the level of the returned order IDs since that's really what we're filtering

hackathon_site/hardware/test_api.py
```diff
@@ -376,6 +391,59 @@ def test_hardware_get_success(self):
...
+        self._login()
+
+        url = self._build_filter_url(team_code="ABCDE")
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = response.json()
+        results = data["results"]
+
+        returned_codes = [res["team_code"] for res in results]
```
grahamhoyes: Same for this and the rest below, assert on the order ID

hackathon_site/hardware/views.py
```diff
@@ -59,6 +60,12 @@ class OrderListView(generics.ListAPIView):
         "POST": OrderCreateSerializer,
     }
 
+    filter_backends = (filters.DjangoFilterBackend, OrderingFilter)
+    filterset_class = OrderFilter
+    ordering_fields = "created_at"
```
grahamhoyes: `ordering_fields` is an iterable, so you have to do `("created_at",)`

hackathon_site/hardware/serializers.py
```diff
@@ -41,10 +41,23 @@ def get_unique_hardware_count(obj: Category) -> int:
...
+            "team",
+            "team_code",
+            "status",
+            "created_at",
+            "updated_at",
+        )
+
+    @staticmethod
+    def get_team_code(obj: Order):
+        return obj.team.team_code
```
grahamhoyes, line 89: Because of this, django has to know to also query the linked team in addition to the Order when building the response. Currently it doesn't know to do that ahead of time, so it makes a separate database query for every team code it needs to pull. You can see that by opening up the Django Debug Toolbar on the browsable API (http://localhost:8000/api/hardware/orders/). For large responses, this will make things very slow.

We can update the queryset on the view to tell it to join in teams ahead of time:

```python
queryset = Order.objects.all().select_related("team")
```

grahamhoyes: Looking more into the debug toolbar, we also need to be joining in order items and hardware, since that's displayed as well. For just 2 orders, 11 database queries are being made (2 of which are for authentication, and 1 for generating the count field which we can't do anything about).

Let's improve the queryset even further:

```python
queryset = (
        Order.objects.all()
        .select_related("team")
        .prefetch_related("hardware_set", "hardware_set__categories")
    )
```

That gets it down to 7 database queries total, no matter how many new orders I add. As for `prefetch_related` vs `select_related`, I'd consult the Django docs for more details. I just try `select_related` first, and if it throws an error switch to `prefetch_related` (and let Pycharm to autocomplete the filed names for me).

hackathon_site/hardware/test_api.py
```diff
@@ -355,8 +357,21 @@ def setUp(self):
         OrderItem.objects.create(
             order=self.order_2, hardware=self.other_hardware,
         )
+        self.order_3 = Order.objects.create(
+            created_at=(datetime.now() + timedelta(days=1)).date(),
```
grahamhoyes: We specify `auto_now_add=True` on this field in the model, which means the created at timestamp is generated by the database. It's actually not even possible to override it like this, so you can remove this line and the datetime import.

hackathon_site/hardware/serializers.py
grahamhoyes: @EdwardJTL and I were talking about this and I forget if we ever wrote it down, but we'll also want to update this view later on so that:
- If the user has order view permissions, then the view behaves like it currently does.
- If the user doesn't have order view permissions, but has a profile, then we automatically filter and return only their team's orders

That way, we can use the same API endpoint for the participant and admin side of the site without having to duplicate a bunch of view logic and (especially) test code.

hackathon_site/hardware/views.py
```diff
@@ -59,6 +60,12 @@ class OrderListView(generics.ListAPIView):
...
+
+    permission_classes = (UserHasHardareOrderViewPermission,)
+
```
grahamhoyes, line 93: Since there appears to be a search box on figma, can you add a [search filter](https://www.django-rest-framework.org/api-guide/filtering/#searchfilter) and search fields as well? We're doing that with the other endpoints too, my bad for not updating the ticket when we started doing it.

Some search fields we might want, from looking at the Figma:
- Order ID
- Team code

Then you just need one search that does for example `_build_filter_url(search="<first 4 characters of a team code>"`. To make sure you don't run into any random issues with the randomly generated team codes, you can set the team codes to be AAAA, BBBB, etc in the setup method.

grahamhoyes (commented):
Great job, just a few minor things!

hackathon_site/hardware/serializers.py
YuYing-Liang: Do we need team_code?

YuYing-Liang: nvm I just saw ur search filter comment

hackathon_site/hardware/test_api.py
YuYing-Liang: how do I test the created_at ordering filter?

hackathon_site/hardware/serializers.py
grahamhoyes: team_code is also included in the select related for team (since it joins the entire team table)

hackathon_site/hardware/test_api.py
grahamhoyes: To start, you know that the timestamps will be increasing in the order you created the orders in the code, so that's your target ordering. 

In the test, you can do `self._build_filter_url(ordering="created_at")` for ascending (so your order ids should be [1, 2, 3]), and `ordering="-created_at"` for descending.

hackathon_site/hardware/serializers.py
YuYing-Liang: Sounds good! @Leo6Leo's incident API ticket does some of this (it has an OrderItemSerializer) [it looks very basic right now](https://github.com/ieeeuoft/hackathon-template/pull/300/files#diff-b2de03caba10b23014e48bd4882fd1da20367e83dd6cf30df1f14debac0e817bR42)

YuYing-Liang:
idk why tests are failing 😭

hackathon_site/hardware/test_api.py
YuYing-Liang: How accurate is the DateTimeField because I printed the response and it seems the orders are created at the exact same time I think

hackathon_site/hardware/permissions.py
YuYing-Liang: I will merge this PR only when his is done

hackathon_site/hardware/serializers.py
YuYing-Liang: I think you made a ticket for what you described above for the permissions https://ieeeuoft.atlassian.net/browse/IEEE-68

hackathon_site/hardware/test_api.py
grahamhoyes: Each one should be a separate database operation, so they should differ ever so slightly even if it's not printing a difference. Does the ordering not behave as expected?

YuYing-Liang: oh nvm it works 👍

hackathon_site/hardware/models.py
```diff
@@ -102,7 +102,7 @@ class Order(models.Model):
         ("Cancelled", "Cancelled"),
     ]
 
-    hardware_set = models.ManyToManyField(Hardware, through=OrderItem)
+    hardware = models.ManyToManyField(Hardware, through=OrderItem)
```
grahamhoyes, line 105: In my last comment I meant to just rename it in the serializer, I didn't realize we named the field `hardware_set` in the model to begin with :thinking:. I like this change (since in most cases we should be specifying an explicit related name rather than the default `<field>_set`).

Since you changed a model though, you will need to make a migration:

```bash
$ python manage.py makemigrations
```

Format the migration that gets created with black, and commit it.

hackathon_site/hardware/views.py
```diff
@@ -52,13 +52,18 @@ def get(self, request, *args, **kwargs):
 
 
 class OrderListView(generics.ListAPIView):
-    queryset = Order.objects.all()
+    queryset = Order.objects.all().select_related("team")
```
grahamhoyes: Still need to prefetch the related hardware:

```python
queryset = ( 
    Order.objects.all() .select_related("team") .prefetch_related("hardware", "hardware__categories") 
)
```

YuYing-Liang: Added a comment

hackathon_site/hardware/views.py
```diff
@@ -52,13 +51,22 @@ def get(self, request, *args, **kwargs):
 
 
 class OrderListView(generics.ListAPIView):
-    queryset = Order.objects.all()
+    queryset = (
+        Order.objects.all().select_related("team")
+        # Causing problems with filtering, will figure out later:
```
grahamhoyes: Minor:

```suggestion
        # TODO: Causing problems with queryset aggregations, will figure out later:
```

The `TODO` wll make it easier to find down the line (pycharm at least has a Todo section that I look through occasionally that tracks these comments)

grahamhoyes (commented):
Looks good! Just need to wait for #308 to be merged in first, then you can use the permissions from there.

Both PRs are one minor fix away, we'll let @Leo6Leo merge his first

grahamhoyes (approved):

PR #312, opened 2021-09-25
IEEE-19 and IEEE-137: PATCH Orders

### LINKED ISSUE #50: PATCH: /api/hardware/orders/<id>/
- Used for changing status to "submitted" (the only field that can change).
	- Should have a validation check that everything is valid:
		- Constraints met
		- Still in stock (someone might have snagged it)

arshinmar:
## Overview

Changed my mind about a couple of things. Both PATCH tickets corresponding to order status is done in the PR. Furthermore, instead of being done in the serializer, permission validation and status validation is done in the respective views for more clarity.

- Resolves #50 (being able to modify the status of the order) and IEEE-137 ticket.
- Created new serializer for status change method.
- Validated to ensure status change is from one of the following to another status option that is valid: 

```
For tech team

"Submitted": ["Cancelled", "Ready for Pickup"],
"Ready for Pickup": ["Picked Up"]

For reg users

"Submitted": ["Cancelled"]
```
- The above validation is done in the respective views (in the hardware and event folder respectively).
- Created 2 new views for patch capability for orders. 
- Patch ability is supported for both tech team and current users.
- There is also validation that reg users cannot change status of orders that are not from their team.

## Unit Tests Created

For tech team

- Test to ensure only allowed status changes are being made (whether checking the beginning status is allowed to change, or the end status is an acceptable final status based on initial status).
- Test to ensure login is required. 
- Test to ensure user has hardware change order capabilities.
- Test to ensure all-rules-passing success.

For regular users.

- Test to ensure only allowed status changes are being made (whether checking the beginning status is allowed to change, or the end status is an acceptable final status based on initial status).
- Test to ensure login is required. 
- Test to ensure user isn't changing orders for those who are not on their team.
- Test to ensure all-rules-passing success.

## Steps to QA

- Try creating a sample order and then testing the "orders/pk" route on swagger for both event and hardware..
- OR you could take a look at the test logic and see if it makes sense.

PS @YuYing-Liang your #320 PR was rlly helpful!


hackathon_site/hardware/serializers.py
```diff
@@ -39,6 +39,14 @@ def get_unique_hardware_count(obj: Category) -> int:
         return obj.hardware_set.annotate(Count("id", distinct=True)).count()
 
 
+class ChangeOrderStatusSerializer(serializers.ModelSerializer):
+    hardware_set = HardwareSerializer(many=True, read_only=True)
+
+    class Meta:
+        model = Order
+        fields = ("id", "hardware_set", "team", "status", "created_at", "updated_at")
```
grahamhoyes: Just an early heads up - the only field we want users to be able to change via the API is `status` (`updated_at` should also be set by the backend, but not the API). You can set the rest as read only fields

hackathon_site/hardware/serializers.py
```diff
@@ -39,6 +39,14 @@ def get_unique_hardware_count(obj: Category) -> int:
         return obj.hardware_set.annotate(Count("id", distinct=True)).count()
 
 
+class ChangeOrderStatusSerializer(serializers.ModelSerializer):
```
grahamhoyes: Nit, but for consistency with the others can you call this `OrderChangeSerializer`?

hackathon_site/hardware/views.py
```diff
@@ -86,3 +89,14 @@ def post(self, request, *args, **kwargs):
             return HttpResponseServerError()
         response_data = response_serializer.data
         return Response(response_data, status=status.HTTP_201_CREATED)
+
+    @transaction.atomic
```
grahamhoyes: Does this need to be atomic? This has an impact on performance, so we don't want to do it unless there's a race condition (which I don't think there is)

hackathon_site/hardware/api_urls.py
```diff
@@ -14,4 +14,12 @@
         views.HardwareDetailView.as_view(),
         name="hardware-detail",
     ),
+    re_path(
+        "hardware/(?P<order_id>[0-9]+)",
```
grahamhoyes: You can just do `<int:pk>` and use regular `path` instead of the regex

hackathon_site/hardware/serializers.py
```diff
@@ -39,6 +39,30 @@ def get_unique_hardware_count(obj: Category) -> int:
         return obj.hardware_set.annotate(Count("id", distinct=True)).count()
 
 
+class OrderChangeSerializer(serializers.ModelSerializer):
+    hardware_set = HardwareSerializer(many=True, read_only=True)
```
grahamhoyes: If this is here so that it returns the full updated order, should also mirror all the fields used by the `OrderListSerializer` (just with everything but `status` in read-only mode)

hackathon_site/hardware/serializers.py
```diff
@@ -39,6 +39,30 @@ def get_unique_hardware_count(obj: Category) -> int:
...
+
+    class Meta:
+        model = Order
+        fields = "status"
+
+    def validate(self, data):
+        current_status = data["status"]
+        requested_status = self.context["request"].status
+        change_options = {
```
grahamhoyes: Nice, this is a great way of doing this check.

Only tricky bit is that we don't want regular users to be able to mark their orders as Ready for Pickup or Picked Up. Is the request available in a serializer by any chance? We could check the permissions in here that way, or we have to just do this in the view directly instead of the serializer

hackathon_site/hardware/views.py
```diff
@@ -120,3 +128,19 @@ def post(self, request, *args, **kwargs):
             return HttpResponseServerError()
         response_data = response_serializer.data
         return Response(response_data, status=status.HTTP_201_CREATED)
+
+    def patch(self, request, order_id, *args, **kwargs):
+        if self.request.user.has_perm("hardware.change_order"):
+            order = self.get_object()
+            serializer = self.get_serializer_class()(
+                order, status="Submitted", partial=True
```
grahamhoyes: Is this just for testing?

hackathon_site/hardware/api_urls.py
```diff
@@ -14,4 +14,12 @@
         views.HardwareDetailView.as_view(),
         name="hardware-detail",
     ),
+    re_path(
+        "orders/(?P<order_id>[0-9]+)",
```
grahamhoyes: Can just be `orders/<int:pk>/` with a regular `path` instead of `re_path`. Same for the one above if you could change it as well, I must have missed that.

hackathon_site/hardware/api_urls.py
```diff
@@ -14,4 +14,12 @@
...
+        views.HardwareListView.as_view(),
```
grahamhoyes: This should be an `OrderDetailView`.

hackathon_site/hardware/api_urls.py
```diff
@@ -14,4 +14,12 @@
...
+        name="update-order",
```
grahamhoyes:
```suggestion
        name="order-detail",
```

hackathon_site/hardware/serializers.py
grahamhoyes: You should be able to get the request from `self.context["request"]`

hackathon_site/hardware/views.py
```diff
@@ -113,10 +122,30 @@ def get(self, request, *args, **kwargs):
...
+
         create_response = serializer.save()
         response_serializer = OrderCreateResponseSerializer(data=create_response)
         if not response_serializer.is_valid():
             logger.error(response_serializer.error_messages)
             return HttpResponseServerError()
         response_data = response_serializer.data
         return Response(response_data, status=status.HTTP_201_CREATED)
+
+    def patch(self, request, *args, **kwargs):
```
grahamhoyes, line 129: You should make a dedicated `OrderDetailView`, that uses `mixins.UpdateModelMixin` and the proper permissions classes, rather than doing permissions checks and validation yourself.

In general, we have two types of API views: list and detail. 

List views return a list of objects, and will have GET and POST methods (for creating new objects). Their urls are like `/api/hardware/orders/`. `mixins.ListModelMixin` and `mixins.CreateModelMixin` provide most of the behaviour you need for these, see `HardwareListView`.

Detail views pertain to a specific object that we know the ID of ahead of time, like `/api/hardware/orders/5/`. The have GET methods (that return a specific instance rather than a list response), and PATCH (or PUT) for updating and DELETE methods. `mixins.RetreiveModelMixin`, `mixins.UpdateModelMixin`, and `mixins.DeleteModelMixin` provide the behaviour.

hackathon_site/hardware/views.py
```diff
@@ -84,7 +86,9 @@ class OrderListView(generics.ListAPIView):
     serializer_method_classes = {
         "GET": OrderListSerializer,
         "POST": OrderCreateSerializer,
+        "PATCH": OrderChangeSerializer,
     }
+    lookup_field = "id"
```
grahamhoyes: If you use the `<int:pk>` syntax in the path, you don't need to have this. Django knows that `pk` ("primary key") will map to `id` for our project.

hackathon_site/hardware/views.py
```diff
@@ -102,6 +106,11 @@ def get_permissions(self):
             return [UserHasProfile()]
         if self.request.method == "GET":
             return [FullDjangoModelPermissions()]
+        if self.request.method == "PATCH":
+            if self.request.user.has_perm("hardware.change_order"):
+                return [FullDjangoModelPermissions()]
```
grahamhoyes: This came up in another PR, but you can just do `[(FullDjangoModelPermissions | UserHasProfile)()]`. Checking for a permission and then returning a permission class that will do that check again is unnecessary.

grahamhoyes:
One more discission we've been having in other PRs is whether we want to keep combining participant and admin functionality into the same endpoints, since it makes handling the different cases of what each group is allowed to do (like which statuses they can set) difficult.

I think for this one it's fine, we can just raise a validation error if a non-admin tries to mark an order as "ready for pickup", for example.

hackathon_site/hardware/views.py
arshinmar: interesting to learn!

hackathon_site/hardware/serializers.py
arshinmar: we will be handling this change via different routes for tech team vs reg user.

arshinmar:
@grahamhoyes would you like me to do both tech-team patch and reg-user patch in this PR?

hackathon_site/event/api_urls.py
```diff
@@ -20,4 +20,9 @@
         name="team-orders",
     ),
     path("teams/<int:pk>/", api_views.TeamDetailView.as_view(), name="team-detail"),
+    path(
+        "teams/team/order/<int:pk>/",
```
```suggestion
        "teams/team/orders/<int:pk>/",
```

hackathon_site/hardware/api_urls.py
```diff
@@ -9,9 +9,10 @@
...
     path("categories/", views.CategoryListView.as_view(), name="category-list"),
     path("incidents/", views.IncidentListView.as_view(), name="incident-list"),
-    re_path(
-        r"^hardware/(?P<pk>[0-9]+)/$",
+    path(
+        "hardware/<int:pk>/",
         views.HardwareDetailView.as_view(),
         name="hardware-detail",
     ),
+    path("orders/<int:pk>", views.OrderDetailView.as_view(), name="order-detail",),
```
grahamhoyes:
```suggestion
    path("orders/<int:pk>/", views.OrderDetailView.as_view(), name="order-detail",),
```

hackathon_site/event/api_urls.py
```diff
@@ -20,4 +20,9 @@
...
+        api_views.TeamOrderDetailView.as_view(),
+        name="order-detail",
```
```suggestion
        name="team-order-detail",
```

I'm not actually sure if having a unique name is required since the other `order-detail` view is in the hardware app. I don't think so, but doesn't hurt to add clarity.

hackathon_site/event/api_views.py
```diff
@@ -127,3 +123,40 @@ class TeamDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
...
+    serializer_class = OrderChangeSerializer
+    permission_classes = [UserHasProfile]
+
+    def patch(self, request, *args, **kwargs):
+        order = self.get_object()
+        change_options = {
+            "Submitted": ["Cancelled"],
+        }
+        current_status = order.status
+        order_team, user_team = (
```
grahamhoyes: Two things:
- Just define each of these separately on different lines, the tuple packing then unpacking is somewhat strange
- Use `order.team` and `request.user.profile.team` directly instead of the team code. Using team code will definitely still work since they are unique, but might as well just use the base team (which uses primary key for equality)

hackathon_site/event/api_views.py
```diff
@@ -127,3 +123,40 @@ class TeamDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
...
+            order.team.team_code,
+            request.user.profile.team.team_code,
+        )
+
+        if order_team != user_team:
+            raise ValidationError(
+                {"detail": "Cannot change the status for another team's order."},
+                code=status.HTTP_403_FORBIDDEN,
```
grahamhoyes: I think 403 is fine for this, but just something to keep in mind in future work: Raising this error message tells someone that that order ID does exist, even though it's not theirs. If this were something like a username instead of an order ID (or something else that you don't want to leak), you would actually want to make this a 404 instead.

hackathon_site/event/api_views.py
```diff
@@ -127,3 +123,40 @@ class TeamDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
...
+        change_options = {
+            "Submitted": ["Cancelled"],
+        }
+        current_status = order.status
+        order_team, user_team = (
+            order.team.team_code,
+            request.user.profile.team.team_code,
+        )
+
+        if order_team != user_team:
```
grahamhoyes, line 164: A thought: Could we do this check with DRF's [object permissions](https://www.django-rest-framework.org/api-guide/permissions/#object-level-permissions) check?

You would just need to write a `check_object_permissions` method like:

```python
    def check_object_permissions(self, request, obj: Order):
        order_team = obj.team
        user_team = request.user.profile.team

        if order_team != user_team:
            raise PermissionDenied("Can only change the status of your orders.")
```

That's probably the cleanest way for us to do the permissions checks for all these user-facing views. Then the actual validation logic (checking the status etc) can go in the serializer's `validate` method, and the patch method has to do nothing more than call `self.partial_update(request, *args, **kwargs)`.

hackathon_site/event/api_views.py
```diff
@@ -127,3 +123,40 @@ class TeamDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
...
+            raise ValidationError(
+                {"detail": "Cannot change the status for another team's order."},
+                code=status.HTTP_403_FORBIDDEN,
+            )
+        if current_status not in change_options:
```
grahamhoyes: Per above, could you move this logic into the serializer's `validate` method? (you also won't need to include the `detail` key in the validation errors in that case, just message strings). 

You'd have to have separate serializers for the team and regular change serializers, but you could add a `change_options` field on `OrderChangeSerializer` that `validate` uses, then just subclass it and change that field for the team one. Something like:

```python
class OrderChangeSerializer(serializers.ModelSerializer):
    hardware = HardwareSerializer(many=True, read_only=True)
    team_code = serializers.SerializerMethodField()

    change_options = {
        "Submitted": ["Cancelled", "Ready for Pickup"],
        "Ready for Pickup": ["Picked Up"],
    }

    class Meta:
        ...

    ...

    def validate(self, data):
        current_status = self.instance.status
        
        if current_status not in self.change_options:
            raise ValidationError("...")
        
        if data["status"] not in self.change_options[current_status]:
            raise ValidationError("...")
```

Then:

```python
class TeamOrderChangeSerializer(OrderChangeSerializer):
    change_options = {
        "Submitted": ["Cancelled"]
    }
```

That saves you the duplicate code in the views as well.

hackathon_site/hardware/views.py
```diff
@@ -118,3 +120,31 @@ def post(self, request, *args, **kwargs):
             return HttpResponseServerError()
         response_data = response_serializer.data
         return Response(response_data, status=status.HTTP_201_CREATED)
+
+
+class OrderDetailView(generics.GenericAPIView, mixins.UpdateModelMixin):
+    queryset = Order.objects.all()
+    serializer_class = OrderChangeSerializer
+    permission_classes = [FullDjangoModelPermissions]
+
```
grahamhoyes, line 128: See above, by moving all this logic into the serializer you can reuse it and have a much simpler view! No need for the `check_object_permissions` method on this view, since there aren't any object permissions in question.

hackathon_site/hardware/test_api.py
```diff
@@ -1258,3 +1258,63 @@ def test_no_remaining_quantities(self):
...
+        )
+        order = Order.objects.create(status="Submitted", team=self.team)
+        OrderItem.objects.create(order=order, hardware=hardware)
+        self.pk = order.id
+
+    def _build_view(self, pk):
+        return reverse(self.view_name, kwargs={"pk": pk})
+
+    def test_user_not_logged_in(self):
+        response = self.client.patch(self._build_view(self.pk))
```
grahamhoyes: You should probably still be submitting data as part of these two requests just to be safe, even though permissions checks happen before the data is looked at.

hackathon_site/hardware/test_api.py
```diff
@@ -1258,3 +1258,63 @@ def test_no_remaining_quantities(self):
...
+    def test_user_lack_perms(self):
+        self._login()
+        response = self.client.patch(self._build_view(self.pk))
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_successful_status_change(self):
+        self._login(self.change_permissions)
+        request_data = {"status": "Ready for Pickup"}
+        response = self.client.patch(self._build_view(self.pk), request_data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
grahamhoyes, line 1302: You should query the order from the database and make sure that it's status was successfully updated

hackathon_site/hardware/test_api.py
```diff
@@ -1258,3 +1258,63 @@ def test_no_remaining_quantities(self):
...
+    def test_unallowed_status_change(self):
+        self._login(self.change_permissions)
+        request_data = {"status": "Picked Up"}
+        response = self.client.patch(self._build_view(self.pk), request_data)
+        self.assertEqual(
+            response.json(),
+            {
+                "detail": "Cannot change the current status of the order to the desired order."
+            },
+        )
```
grahamhoyes, line 1317: Ditto, would be good to query the order and make sure it's status hasn't changed

hackathon_site/hardware/test_api.py
```diff
@@ -1258,3 +1258,63 @@ def test_no_remaining_quantities(self):
...
+
+    def test_failed_beginning_status(self):
```
grahamhoyes, line 1319: Not sure what this method means, what is beginning status? Also the same as above, check to make sure the order isn't changed.

hackathon_site/event/test_api.py
```diff
@@ -471,3 +471,65 @@ def test_team_get_has_permissions(self):
         data = response.json()
 
         self.assertEqual(expected_response[0], data)
+
+
+class TeamOrderDetailViewTestCase(SetupUserMixin, APITestCase):
```
grahamhoyes, line 618: Same comments for this one as for the other tests

grahamhoyes (commented):
Logic and view setup looks perfect to me, some comments on moving the validation checks into the serializer for reusability and on using object permissions.

Good work!

hackathon_site/event/api_views.py
arshinmar: ok got it

hackathon_site/event/api_views.py
arshinmar: cool I agree, I was thinking about how to best do this, I was initially trying to limit serializer complexity, but inheritance allows it to be really simple for sure.

hackathon_site/hardware/views.py
arshinmar: for sure, def cleaner

hackathon_site/hardware/test_api.py
arshinmar: Basically, you cannot change the status of an order that has status "Picked Up"

arshinmar: you can only change the status of an order if the status is "Submitted" or "Ready for Pickup"

hackathon_site/hardware/test_api.py
```diff
@@ -1258,3 +1258,70 @@ def test_no_remaining_quantities(self):
...
+        )
+        self.assertFalse(request_data["status"] == Order.objects.get(id=self.pk).status)
+
+    def test_failed_beginning_status(self):
+        self._login(self.change_permissions)
+        request_data = {"status": "Cancelled"}
+        order = Order.objects.create(status="Picked Up", team=self.team)
+        response = self.client.patch(self._build_view(order.id), request_data)
+        self.assertEqual(
+            response.json(), {"detail": ["Cannot change the status for this order."]},
```
Leo6Leo: I think it would be better to probably use arrays to store these "detail sentences" at the beginning of the test case. If we would like to change the prompt info in the future, it would be much easier to do that. How do you think Arsh?

arshinmar: Could you clarify what you mean by "prompt"?

Leo6Leo: oops I used the wrong word. I am referring to those "detail sentences": "Cannot change the status for thsi order"

arshinmar: Hmmm I don't think it would be helpful. A list array isn't really beneficial as it doesn't state which test requires which detail statement.

A dict would be a better choice that maps a test name string to a detail statement, but even then, it may be better visually to see the expected detail statement within the actual subtest.

Furthermore, we haven't really adopted that convention (the list array convention) in many other test cases, and I'm not sure if it's a good idea to cause that significant change here.

Leo6Leo: yeah, that's right. Dict would be better because we could map the test name easier. Something like this:
details = 
{
   "cannot_change _order" : "Cannot change the status for this order",
   "only_change" : "Can only change the status of your orders.",
  "..." : ... ,
}

hackathon_site/hardware/serializers.py
```diff
@@ -89,6 +89,56 @@ def get_team_code(obj: Order):
...
+    def get_team_code(obj: Order):
+        return obj.team.team_code
+
+    def validate(self, data):
+        current_status = self.instance.status
+
+        if current_status not in self.change_options:
+            raise ValidationError(
+                {"detail": "Cannot change the status for this order."},
+                code=status.HTTP_403_FORBIDDEN,
```
grahamhoyes: I think this one should still be a 400? That'll just make the logic on the frontend a bit easier, since we can generically catch 400s and display them in snackbars

hackathon_site/hardware/serializers.py
```diff
@@ -89,6 +89,56 @@ def get_team_code(obj: Order):
...
+    @staticmethod
+    def get_team_code(obj: Order):
+        return obj.team.team_code
+
+    def validate(self, data):
+        current_status = self.instance.status
+
+        if current_status not in self.change_options:
+            raise ValidationError(
+                {"detail": "Cannot change the status for this order."},
```
grahamhoyes: This doesn't need to be a dict with a "detail" key, you can just put the message in directly. The actual output is the same, so no need to change the tests

hackathon_site/hardware/test_api.py
```diff
@@ -1258,3 +1258,70 @@ def test_no_remaining_quantities(self):
...
+        request_data = {"status": "Ready for Pickup"}
+        response = self.client.patch(self._build_view(self.pk), request_data)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_successful_status_change(self):
+        self._login(self.change_permissions)
+        request_data = {"status": "Ready for Pickup"}
+        response = self.client.patch(self._build_view(self.pk), request_data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(request_data["status"], Order.objects.get(id=self.pk).status)
```
grahamhoyes, line 1303: This still doesn't guarantee that the field was actually changed - if the serializer discarded the field, this test would still pass. You'll need to test that the object's status is explicitly "Ready for Pickup". Same for the other tests.

hackathon_site/hardware/test_api.py
grahamhoyes: Ah gotcha. Might want to throw a docstring on the method to make it clear how this method is different from `test_unallowed_status_change`

hackathon_site/hardware/serializers.py
grahamhoyes: Also if you want a 400 status code, you don't need to specify it (that's the default)

hackathon_site/hardware/serializers.py
```diff
@@ -89,6 +89,56 @@ def get_team_code(obj: Order):
...
+                code=status.HTTP_403_FORBIDDEN,
+            )
+
+        if data["status"] not in self.change_options[current_status]:
+            raise ValidationError(
+                {
+                    "detail": "Cannot change the current status of the order to the desired order."
```
grahamhoyes: Could be fancy here: `f"Cannot change the status of an order from {current_status} to {data['status']}."`

hackathon_site/hardware/test_api.py
arshinmar: I'm not sure I understand. I thought I did check whether the object's status (Order.objects.get(id=self.pk).status)) is explicitly "Ready for Pickup" (request_data["status"])

hackathon_site/hardware/test_api.py
arshinmar: @grahamhoyes what do you think? I personally don't wanna change, but as it is a recommendation that differs from typically how we do things in the django tests, I thought your input would be useful here

hackathon_site/hardware/test_api.py
arshinmar: are you suggesting if somehow, it took out the "status" field from request_data, in the self.client.patch?

hackathon_site/hardware/test_api.py
grahamhoyes: I agree with Arsh, I don't think it makes sense to store the them at the test class level and index into them for each test for two reasons:
1. Each detail string only appears once in the `OrderListPatchTestCase` anyway, so there's only one place to change them now anyway.
2. It's best to keep things defined as close to the test assertion as possible, since it makes things easier to follow (unless it's truly generic to all the methods, like `self.view_name` or something). In this case, if a test failed because that detail string changed, we can see directly from the test output what the issue is.

hackathon_site/hardware/test_api.py
grahamhoyes: I am not doing very good with PRs lately, I was reading that as response["status"] 🤦‍♂️. You're totally right, this is perfect as-is!

hackathon_site/hardware/serializers.py
arshinmar: When this is done 
```
raise ValidationError(
                f"Cannot change the status of an order from {current_status} to {data['status']}.",
                code=status.HTTP_400_BAD_REQUEST,
            )
```
it leads to the error detail json to be {"non_field_errors": error_message} instead of {"detail": error_message}

arshinmar: let me know how you want to continue

grahamhoyes: Ah right, that brings up a good point - errors thrown in `validate` cause a generic `non_field_errors` key. 

Could you try making this method `validate_status` and see what happens? I think then it will throw a field error specific to the status, which in theory is useful on the frontend.

Sorry for all the round about changes on this, it's been quite some time since I've done serializer validation.

hackathon_site/hardware/serializers.py
```diff
@@ -89,6 +89,54 @@ def get_team_code(obj: Order):
...
+    team_code = serializers.SerializerMethodField()
+
+    change_options = {
+        "Submitted": ["Cancelled", "Ready for Pickup"],
+        "Ready for Pickup": ["Picked Up"],
+    }
+
+    class Meta:
+        model = Order
+        fields = OrderListSerializer.Meta.fields
```
grahamhoyes, line 101: If you're going to do this, want to just define the class as `class OrderChangeSerializer(OrderListSerializer)`, then you won't have to specify the `hardware`, `team_code`, and `get_team_code` fields / methods? Meta class stays the same

grahamhoyes: Also, this will only work because `hardware` and `team_code` are on the list of read only fields. If eg hardware needed to be modified too by passing a different hardware ID, then you wouldn't be able to subclass because of the nested serializer.

arshinmar: cool

hackathon_site/hardware/serializers.py
arshinmar: ye nw, will change to status

grahamhoyes (approved):
💯

PR #313, opened 2021-09-25
Connect Category Store

grahamhoyes:
## Overview

- Connects the category API to our redux store, and the inventory filter where categories are displayed to the store
- Refactors the inventory filter component slightly to use redux hooks instead of connect()

## Unit Tests Created

- Integration style tests for the inventory page and filter, to make sure data is fetched and category filters are submitted
- Error tests for the thunk
- Thunk test for exhaustively querying the paginated API

## Steps to QA

- Create some categories and hardware via the Admin site if you haven't already
- Visit the Inventory page, make sure categories load properly
- Submit category filters, make sure they still work
- Bonus: Add over 100 categories to make sure pagination works, though I'm confident it does from the tests

### DIFF
--- a/hackathon_site/dashboard/frontend/src/api/api.ts
+++ b/hackathon_site/dashboard/frontend/src/api/api.ts
@@ -46,7 +46,8 @@ const makeConfig = () => ({
 });
 
 export const stripHostname = (url: string): string => {
-    return url.replace(SERVER_URL, "");
+    const parsed = new URL(url);
+    return parsed.pathname + parsed.search + parsed.hash;
 };
 
 export const get = <T>(
--- a/hackathon_site/dashboard/frontend/src/api/api.ts
+++ b/hackathon_site/dashboard/frontend/src/api/api.ts
@@ -45,6 +45,10 @@ const makeConfig = () => ({
     withCredentials: true,
 });
 
+export const stripHostname = (url: string): string => {
+    return url.replace(SERVER_URL, "");
+};
+
 export const get = <T>(
     uri: string,
     params?: { [key: string]: any }
--- a/hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.tsx
+++ b/hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.tsx
@@ -42,29 +42,38 @@ describe("<InventoryFilter />", () => {
      * the level of the (mocked) API requests.
      */
 
+    const prepopulateStore = () => {
+        const apiResponse = makeMockApiListResponse(mockCategories);
+        mockedGet.mockResolvedValue(apiResponse);
+
+        const store = makeStore();
+        store.dispatch(getCategories());
+
+        return store;
+    };
+
     it("Submits selected filters", async () => {
-        const { getByText } = render(<InventoryFilter />);
+        const store = prepopulateStore();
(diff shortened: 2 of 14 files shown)


hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.tsx
```diff
@@ -72,19 +82,21 @@ describe("<InventoryFilter />", () => {
...
 
     it("Clears filters when toggled", async () => {
-        const { getByText } = render(<InventoryFilter />);
+        const store = prepopulateStore();
+        const { getByText, findByText } = render(<InventoryFilter />, { store });
 
         const orderByNameButton = getByText("Z-A");
         const inStockCheckbox = getByText("In stock");
-        const FPGACheckbox = getByText("FPGA");
+        const firstCheckbox = await findByText(mockCategories[0].name);
```
grahamhoyes, line 90: Calling this out for anyone who's curious - `getByText()` will error immediately if the text isn't there, `findByText` will return a promise (with I think a 5 second timeout) for the data to appear.

Since it takes a bit of time for the action dispatched in `prepopulateStore()` to make it's way through, hit the (mocked)  API, etc, this is just waiting for the categories to be available in the store / filters.

Could also use this instead of `await waitFor(() => {...})` in some other places, but I think it's a bit more obvious what's going on with that one in most cases.

hackathon_site/dashboard/frontend/src/api/api.ts
```diff
@@ -45,6 +45,11 @@ const makeConfig = () => ({
     withCredentials: true,
 });
 
+export const stripHostname = (url: string): string => {
+    const parsed = new URL(url);
```
YuYing-Liang, line 49: Oh okay, yeah I guess this works better than JavaScript replace lol

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.tsx
```diff
@@ -142,11 +154,25 @@ describe("<InventoryFilter />", () => {
         });
     });
 
-    it("Renders options for each category", () => {
-        const { getByText } = render(<InventoryFilter />);
+    it("Renders loading bar, then options for each category", async () => {
```
YuYing-Liang, line 157: Not the right line but do we need a 'describe' for each test? E.g. having multiple 'it()' tests for InventoryFilter

YuYing-Liang (approved):

hackathon_site/dashboard/frontend/src/api/api.ts
grahamhoyes: Had to actually switch to this since in production `SERVER_URL` will just be an empty string, but the API will come back with the full `https://example.com/api/...`.

hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.tsx
grahamhoyes: Technically no - a `describe()` wraps a set of tests, which themselves are either `it()` or `test()` (both the same function, I just use whichever sounds better in my head for a given test description).

You could have `it()`/`test()` tests at the top level of the file, and you can also have nested `describe()`s. They just help to organize the code and the output from jest in the console.

PR #314, opened 2021-09-27
Upgrade Django to v3.2

grahamhoyes:
## Overview

- Keeping on top of things, this upgrades Django from v3.1 to v3.2
- No backwards incompatible changes that we need to worry about. The only thing that needed changing was specifyng the type of the default primary key field for models, as described [here](https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys)

## Unit Tests Created

- N/a

## Steps to QA

- Shouldn't need anything, tests should cover it

### DIFF
--- a/hackathon_site/hackathon_site/settings/__init__.py
+++ b/hackathon_site/hackathon_site/settings/__init__.py
@@ -136,6 +136,8 @@ DATABASES = {
     }
 }
 
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
+
 # Cache
 # https://docs.djangoproject.com/en/3.1/topics/cache/
 REDIS_URI = os.environ.get("REDIS_URI", "172.17.0.1:6379/1")
--- a/hackathon_site/requirements.txt
+++ b/hackathon_site/requirements.txt
@@ -7,7 +7,7 @@ click==7.1.2
 coreapi==2.3.3
 coreschema==0.0.4
 dj-rest-auth==1.0.6
-Django==3.1.13
+Django==3.2.7
 django-cors-headers==3.3.0
 django-debug-toolbar==2.2.1
 django-filter==2.4.0


hackathon_site/hackathon_site/settings/__init__.py
```diff
@@ -136,6 +136,8 @@
     }
 }
 
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
```
YuYing-Liang, line 139: why do we need to change this?

hackathon_site/requirements.txt
```diff
@@ -7,7 +7,7 @@ click==7.1.2
 coreapi==2.3.3
 coreschema==0.0.4
 dj-rest-auth==1.0.6
-Django==3.1.13
+Django==3.2.7
```
YuYing-Liang, line 10: 👍

hackathon_site/hackathon_site/settings/__init__.py
Leo6Leo: I think this is used to avoid unwanted migrations in the future (Saw it in Django doc) @YuYing-Liang

YuYing-Liang: Ooo okay ty!

YuYing-Liang (approved):

hackathon_site/hackathon_site/settings/__init__.py
grahamhoyes: This used to be an implicit default, but in future versions of django it's going to change to `BigAutoFiled` which would generate migrations for every model. In Django 3.2 you'll get a warning if you haven't explicitly set one in settings.

https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys

PR #315, opened 2021-09-29
Added hardware count selector to inventory page

YuYing-Liang:
## Overview

- Resolves [IEEE-90](https://ieeeuoft.atlassian.net/browse/IEEE-90)
- Added a hardware count selector in hardware slice
- Displays SHOWING {number of hardware items queried} OF {total hardware items} ITEMS
- Displays total number of items beside search bar
- Displays "Load More" button if the amount of hardware items retrieved is less than the total number of items

## Unit Tests Created

- Default hardware count
- hardware count with filtering by category

## Steps to QA

- Use the inventory page normally (filter by any category, try ordering too) and see if the hardware counts change in the places described above


hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -34,13 +35,48 @@ describe("Inventory Page", () => {
...
 
+        // by default the hardwareUri has a limit of 100
+        if (mockHardware.length <= 100) {
+            expect(queryByText(/load more/i)).not.toBeInTheDocument();
+        } else {
+            expect(getByText(/load more/i)).toBeInTheDocument();
+        }
+    });
+
+    test("Hardware count", async () => {
```
YuYing-Liang: I'm not sure how I should test the hardware count when filtering by category, I was thinking returning a filtered response from the api but that feels like cheating lol. The other way is going through the process of applying the filter and then checking but we already test the filtering part in another test so that feels redundant. What do you guys think?

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -6,6 +6,7 @@ import Inventory from "pages/Inventory/Inventory";
 import { mockCategories, mockHardware } from "testing/mockData";
 
 import { get } from "api/api";
+import { within } from "@testing-library/react";
```
grahamhoyes: Minor, can you import this from `testing/utils`? That module re-exports everything from RTL (which they [recommend](https://testing-library.com/docs/react-testing-library/setup/#custom-render))

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -34,13 +35,48 @@ describe("Inventory Page", () => {
...
+            null,
+            mockHardware.length
+        );
+
+        when(mockedGet)
+            .calledWith(hardwareUri, {})
+            .mockResolvedValue(hardwareApiResponse);
+
+        const { getByText, getByTestId } = render(<Inventory />);
+        const { getByText: getTotalCountText } = within(
```
grahamhoyes: Cool, I didn't know about `within`!

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -34,13 +35,48 @@ describe("Inventory Page", () => {
...
+            getByTestId("inventory-total-count")
```
grahamhoyes: When it comes to UI testing, the general guideline is "test what users see". `getByTestId` breaks that principle, and should really only be used as a last resort when things like `getByText` can't be used (such as for testing that loading bars appear).

In this case, do we absolutely need to use `getByTestId`? Or would it be sufficent just to use the regular `getByText()` on the entire page? In all of our other tests, we don't really test _where_ things appear, only _that_ they appear. Too much of the former leads back to snapshot tests, which get annoying when you want to update the UI or something that shouldn't break functional tests like these.

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
```diff
@@ -107,7 +114,12 @@ const Inventory = () => {
                                 </Hidden>
 
                                 <div className={styles.inventoryBodyToolbarRefresh}>
-                                    <Typography variant="body2">123 items</Typography>
+                                    <Typography
+                                        variant="body2"
+                                        data-testid="inventory-total-count"
+                                    >
+                                        {count} items
```
grahamhoyes: Hmm, does `{count} items` make sense? With that terminoligy, I would expect that to be the number of items displayed, not total.

The figma uses `{count} total results`, could we switch this to that or `{count} results`? (I prefer the shorter one, I defer to you for which one to pick).

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
```diff
@@ -117,17 +129,21 @@ const Inventory = () => {
                         <InventoryGrid />
                         <Divider className={styles.inventoryLoadDivider} />
                         <Typography variant="subtitle2" align="center" paragraph>
-                            SHOWING 100 OF 123 ITEMS
+                            {count > 0
+                                ? `SHOWING ${items.length} OF ${count} ITEMS`
+                                : "NO ITEMS FOUND"}
```
grahamhoyes, line 136: The inventory grid component also shows "No items found". 

How about if there are no items, we remove the `Divider` above and remove the no items message from either the inventory or inventory grid component? The message in the inventory grid looks kinda wonky, so I'd suggest removing that one.

Will need tests for this in the inventory and inventory grid tests too.

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -34,13 +35,48 @@ describe("Inventory Page", () => {
...
 
-        const { getByText } = render(<Inventory />);
+        const { getByText, queryByText } = render(<Inventory />);
 
         await waitFor(() => {
             expect(getByText(mockHardware[0].name)).toBeInTheDocument();
             expect(getByText(mockCategories[0].name)).toBeInTheDocument();
         });
 
+        // by default the hardwareUri has a limit of 100
```
grahamhoyes: This test doesn't really make sense

- `mockHardware` is always going to be less than 100 elements. It doesn't change as part of the test, and we're never going to add that many things to it
- "100" is not hard coded anywhere in the front end, and is not a special number. We could just as easily change the backend to use 50 or 200 items per page (to test this PR, I set it to 1 item per page).

This test is passing because `mockHardware` happens to be less than 100 elements, and the count generated by `makeMockApiListResponse` is the length of the data you pass in (ie, "load more" will never appear). If we put 101 elements in that mock data array, this test would fail.

I would remove this block entirely, and let the functionality be covered by your new test cases

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
grahamhoyes: Is that something we need to test?

We already know that:
- The hardware API will always return a `count`. This is guaranteed by our backend
- That filtering hits the same hardware API, just with filter parameters applied. The inventory filter tests guarantee this.
    - Side note - looks like I never wrote a test for the actual api `get` function, so no guarantees that it's actually building filter params correct. Whoops 😬 . If you wanted to do that, you could create an api.test.ts file and use jest to mock axios (or just make a backlog ticket for it)

So in short, applying filters shouldn't have any effect on the `count` parameter, so there's no need to write tests that include applying filters (unless I'm missing something).

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -34,13 +35,48 @@ describe("Inventory Page", () => {
...
+        // Mock hardware api response with a limit of mockHardware.length - 2
+        const limit = mockHardware.length - 2;
+        const hardwareApiResponse = makeMockApiListResponse(
+            mockHardware.slice(0, limit),
+            null,
+            null,
+            mockHardware.length
+        );
+
+        when(mockedGet)
```
grahamhoyes, line 58: Since you're rendering the full inventory page, you should also mock the categories response that it expects. Right now it will just pass a bunch of mock objects through which won't break anything, but in theory they could if we ever incorporate more error checking.

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
grahamhoyes: You should break this into two test cases - one where the load more button appears, and one where it doesn't.

Something like

```typescript
it("Shows count and load more button when there is more hardware to fetch", async () => {
    ...
});

it("Shows count and no load more button when there is no more hardware to fetch", async () => {
    ...
});
```

Descriptive test names are also preferred (which is why I prefer using `it()` since it reads nicer).

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -34,13 +35,48 @@ describe("Inventory Page", () => {
...
+
+        await waitFor(() => {
+            expect(getByText(mockHardware[0].name)).toBeInTheDocument();
+        });
+
+        expect(getTotalCountText(`${mockHardware.length} items`)).toBeInTheDocument();
+        expect(
+            getByText(`SHOWING ${limit} OF ${mockHardware.length} ITEMS`)
+        ).toBeInTheDocument();
         expect(getByText(/load more/i)).toBeInTheDocument();
```
grahamhoyes, line 75: See above, but you should also have a test case that covers when this button shouldn't be displayed.

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: Ah okay, I used test() just to see what you think lol
The reason why I didn't break it up was because I thought the first test case would not show the load more button (since I don't think we'd mock 100 items) and so if I made another test for it, the test would look almost the same as the first test.

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
YuYing-Liang: The shorter the better, ppl don't like to read 👍

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: I see, I see (pun intended), I will use getByText().

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
YuYing-Liang: oh right, I forgot to test when there are no items!
What do we need to test in the inventory grid if we are removing that message?

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: I don't understand how it would fail because there is an else block but I agree, I will test the load more button in separate tests.

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
YuYing-Liang: Also since we are removing the divider, in this case we can use getByTestId() right?

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: Ignore the comment above, I read your previous comments (sometimes github shows the comments out of order)

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
grahamhoyes: If we added a bunch more elements to `mockHardware`, we'd go into the else block, and the test would fail because count === hardware.length, so the "load more" button won't be there

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
grahamhoyes: For testing the inventory grid, you can probably just remove the test for the "no items found" message

For the divider, tbh I wouldn't bother unit testing it since it's a minor UI thing. But if you'd like to (I'm fine either way), then a test ID would be the way to go

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: oh I right, we are mocking the endpoint so there is no automatic limit 🤦‍♀️

hackathon_site/dashboard/frontend/src/components/inventory/InventoryGrid/InventoryGrid.tsx
```diff
@@ -36,12 +36,7 @@ export const InventoryGrid = () => {
                             currentStock={item.quantity_remaining}
                         />
                     </Grid>
-                ))
-            ) : (
-                <Grid xs={12} item className={styles.Item}>
-                    <Typography>No items found</Typography>
```
grahamhoyes, line 42: Can remove the typography import since it's not used anymore

grahamhoyes (commented):
Super minor old import to remove, then looks good to me!

grahamhoyes (approved):

PR #317, opened 2021-10-01
IEEE-81 Team Detail View

Leo6Leo:
## Overview

- permissions: event.teams.view 
- look at code for /event/teams/team but add a query for team_code

## Unit Tests Created

- EventTeamCodeListsViewTestCase


hackathon_site/event/views.py
```diff
@@ -226,3 +227,43 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
...
+        filter = {}
+        for field in self.lookup_fields:
+            if self.kwargs[field]:  # Ignore empty fields.
+                filter[field] = self.kwargs[field]
+        obj = get_object_or_404(queryset, **filter)  # Lookup the object
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+
+class TeamCodeView(MultipleFieldLookupMixin, generics.GenericAPIView):
```
grahamhoyes: Take a look at DRFs [RetrieveModelMixin](https://www.django-rest-framework.org/api-guide/generic-views/#retrievemodelmixin) - It should do what we need by just setting the lookup field. The get method then just needs to be:

```python
```

Take a look at the `CurrentUserAPIView` for an example of how to use it (you don't need the `get_object` method that's in there, that one had custom behaviour to get the object from the request user rather than a URL kwarg)

hackathon_site/event/views.py
```diff
@@ -226,3 +227,43 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
 
     def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
+
+
+class MultipleFieldLookupMixin:
```
grahamhoyes: Cool! But is this strictly necessary? Are we going to have more than one lookup field for this endpoint?

hackathon_site/event/api_urls.py
```diff
@@ -14,4 +14,7 @@
     ),
     path("teams/leave_team/", api_views.LeaveTeamView.as_view(), name="leave-team"),
     path("teams/", views.TeamListView.as_view(), name="team-list"),
+    path(
+        "teams/<str:team_code>", views.TeamCodeView.as_view(), name="team-code-detail"
```
grahamhoyes: We'll have a discussion about this today during the meeting, but I think this should be `teams/<int:pk>/` to do the lookup based off of primary key (which is `id`).

I think the use case for this endpoint is to eg fetch a team given an order response. The order response contains just the team's ID, not their team code, so that's what we need the lookup to be based off of if that is the intended use case.

grahamhoyes:
```suggestion
        "teams/<str:team_code>", views.TeamCodeView.as_view(), name="team-detail"
```

hackathon_site/event/test_api.py
```diff
@@ -4,6 +4,8 @@
 from rest_framework.test import APITestCase
 from hackathon_site.tests import SetupUserMixin
 from django.contrib.auth.models import Permission
+import json
```
grahamhoyes: Minor, move these to the top of the file above the 3rd party imports (separated by a blank line)

hackathon_site/event/views.py
grahamhoyes: Can you move this to `event/api_views.py`? That's something I should have caught with all the other API views we've done as well (IEEE-88 is a ticket to move all the other ones to where they should be).

grahamhoyes:
```suggestion
class TeamDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
```

Calling things `<model>DetailView` and `<model>ListView` is a common convention. Usually they each have separate GET methods (detail view returns a single instance while list view returns a list), the detail view has PATCH (modify) and DELETE methods, and the list view has a POST (create) method

hackathon_site/event/views.py
```diff
@@ -226,3 +227,43 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
...
+    lookup_fields = ["team_code"]
+
+    filter_backends = (filters.DjangoFilterBackend, SearchFilter)
+    filterset_class = TeamFilter
+    search_fields = (
+        "team_code",
+        "id",
+        "profiles__user__first_name",
+        "profiles__user__last_name",
+    )
```
grahamhoyes: Since this is a detail view, not a list view, there's no searching or filtering. It will always return the exact object that was requested, not a list of them.

hackathon_site/event/test_api.py
```diff
@@ -366,3 +368,47 @@ def test_name_search_filter(self):
...
+        self.team = Team.objects.create()
+        self.team2 = Team.objects.create()
+        self.team3 = Team.objects.create()
+        self.permissions = Permission.objects.filter(
+            content_type__app_label="event", codename="view_team"
+        )
+        super().setUp()
+        self.view = reverse("api:event:team-code-detail", args=[self.team.team_code])
+
+    def _build_filter_url(self, **kwargs):
```
grahamhoyes: Don't need this, no filtering needed

hackathon_site/event/test_api.py
```diff
@@ -366,3 +368,47 @@ def test_name_search_filter(self):
...
+            ).data
+        )
+
+        data = response.json()
+
+        expected_response = json.loads(response_list_obj)
+
+        self.assertEqual(expected_response[0], data)
+
+    def test_team_get_not_login(self):
```
grahamhoyes: Minor, can you move this to the top of the test cases? It's nice to just get the not logged in and no permissions tests out of the way at the beginning.

hackathon_site/event/test_api.py
```diff
@@ -366,3 +368,47 @@ def test_name_search_filter(self):
...
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_team_get_has_permissions(self):
+        self._login(self.permissions)
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        queryset = Team.objects.filter(team_code=self.team)
+
+        # need to provide a request in the serializer context to produce absolute url for image field
```
grahamhoyes: Shouldn't be any image urls here

hackathon_site/event/test_api.py
```diff
@@ -366,3 +368,47 @@ def test_name_search_filter(self):
...
+        response_list_obj = json.dumps(
```
grahamhoyes: Once you switch over to the retrieve model mixin, the response won't be a list anymore

grahamhoyes:
Also just want to say that given the lack of context in the ticket and background on what a DRF detail view is, this is a great first shot!

hackathon_site/event/views.py
Leo6Leo: Because in the ticket, it says we need to use team_code as the condition to retrieve the data. I have tested by using  RetrieveModelMixin, but it uses <int:pk> as the default condition and the only solution I found to use team_code instead of <int:pk> as the query condition is to use MultipleFieldLookupMixin.(this method is introduced in RESTFUL document) @grahamhoyes

hackathon_site/event/views.py
Leo6Leo: If we would like to use team_code as the query condition, then yes. If we would like to change the condition to team_id(int:pk), then MultipleFieldLookupMixin is not necessary. That's what I found out so far.

hackathon_site/event/views.py
grahamhoyes: Makes sense! I think you did right by the ticket, I'm questioning if the ticket itself is correct in what it's asking to do. There's some threads on slack discussing it too

Leo6Leo:
image
Mail Test failed

hackathon_site/event/api_views.py
```diff
@@ -101,3 +106,12 @@ def post(self, request, *args, **kwargs):
         response_serializer = TeamSerializer(profile.team)
         response_data = response_serializer.data
         return Response(data=response_data, status=status.HTTP_200_OK,)
+
+
+class TeamIDView(mixins.RetrieveModelMixin, generics.GenericAPIView):
```
grahamhoyes:
```suggestion
class TeamDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
```

hackathon_site/event/api_urls.py
```diff
@@ -14,4 +14,5 @@
     ),
     path("teams/leave_team/", api_views.LeaveTeamView.as_view(), name="leave-team"),
     path("teams/", views.TeamListView.as_view(), name="team-list"),
+    path("teams/<str:pk>", api_views.TeamIDView.as_view(), name="team-detail"),
```
grahamhoyes: `pk` will query the model's primary key, which for all of our models is the `id` field, which is an int. So this path should be `teams/<int:pk>/` (also note the trailing slash, which we use for all URL routes)

hackathon_site/event/test_api.py
```diff
@@ -366,3 +368,39 @@ def test_name_search_filter(self):
         results = data["results"]
         returned_ids = [res["team_code"] for res in results]
         self.assertCountEqual(returned_ids, [self.team2.team_code])
+
+
+class EventTeamCodeListsViewTestCase(SetupUserMixin, APITestCase):
```
grahamhoyes: Lots going on in this class name! This can just be `EventTeamDetailViewTestCase` (this is for the detail view, not the list view).

hackathon_site/event/test_api.py
```diff
@@ -5,6 +5,8 @@
 from hackathon_site.tests import SetupUserMixin
 from django.contrib.auth.models import Permission
 
+from collections import OrderedDict
```
grahamhoyes: Unused import

Leo6Leo:
Mail test failed again. Re-run the tests did not work.

hackathon_site/event/api_urls.py
```diff
@@ -14,4 +14,5 @@
     ),
     path("teams/leave_team/", api_views.LeaveTeamView.as_view(), name="leave-team"),
     path("teams/", views.TeamListView.as_view(), name="team-list"),
+    path("teams/<int:pk>", api_views.TeamDetailView.as_view(), name="team-detail"),
```
grahamhoyes: Needs a trailing slash at the end (`"teams/<int:pk>/"`)

grahamhoyes (commented):
Just one missing character, then looks good!

Sorry about those mailing tests always failing at midnight :(

grahamhoyes (approved):

PR #318, opened 2021-10-06
Clean phone numbers in application

GauravRanganath:
## Overview

- Resolves IEEE-59
- Removed all non numeric characters before saving to database

## Steps to QA

- Try submitting an application with a phone number that has non-numeric characters (ex: + 1 123-456 7890) and check that non-numeric characters are removed in the database

### DIFF
--- a/hackathon_site/registration/forms.py
+++ b/hackathon_site/registration/forms.py
@@ -11,7 +11,7 @@ from hackathon_site.utils import is_registration_open
 from registration.models import Application, Team, User
 from registration.widgets import MaterialFileInput
 
-
+import re
 class SignUpForm(UserCreationForm):
     """
     Form for registering a new user account.
@@ -171,6 +171,7 @@ class ApplicationForm(forms.ModelForm):
 
         self.instance.user = self.user
         self.instance.team = team
+        self.instance.phone_number = re.sub("[^0-9]", "", self.instance.phone_number)
 
         if commit:
             self.instance.save()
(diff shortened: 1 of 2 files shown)


hackathon_site/registration/forms.py
```diff
@@ -11,6 +11,8 @@
 from registration.models import Application, Team, User
 from registration.widgets import MaterialFileInput
 
+import re
```
grahamhoyes: Can you move this up to below the `dateutil` import, with the rest of the stdlib imports?

hackathon_site/registration/forms.py
```diff
@@ -171,6 +173,7 @@ def save(self, commit=True):
 
         self.instance.user = self.user
         self.instance.team = team
+        self.instance.phone_number = re.sub("[^0-9]", "", self.instance.phone_number)
```
grahamhoyes, line 175: Needs a unit test, you can put it in the sign up form tests

GauravRanganath: I think it makes more sense to put it in the application form tests no?

grahamhoyes: Whoops, yes you're right!

hackathon_site/registration/forms.py
grahamhoyes: Still need to move this to below the `dateutil` import to keep our imports organized

grahamhoyes: Still needs to be moved up where it should be

grahamhoyes (approved):

PR #319, opened 2021-10-07
IEEE-68: added permissions and tests for getting orders for regular participants

YuYing-Liang:
## Overview

- Resolves [IEEE-68](https://ieeeuoft.atlassian.net/browse/IEEE-68)
- Added permission in order list view for regular participants
- if a user has a profile, it will filter by their team's orders, otherwise it will throw a 403 error
- if a user has change permissions, they get to see all orders

## Unit Tests Created

- Testing only view permissions with no profile
- Testing only view permissions with a profile
- Testing change permissions
- Testing filtering team's orders with view permissions and a profile
- Testing filtering another team's orders with with view permissions and a profile

## Steps to QA

### DIFF
--- a/hackathon_site/hardware/test_api.py
+++ b/hackathon_site/hardware/test_api.py
@@ -481,9 +481,6 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         self.view_permissions = Permission.objects.filter(
             content_type__app_label="hardware", codename="view_order"
         )
-        self.all_hardware_permissions = Permission.objects.filter(
-            content_type__app_label="hardware"
-        )
         self.view = reverse("api:hardware:order-list")
 
     def _build_filter_url(self, **kwargs):
@@ -495,22 +492,16 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         response = self.client.get(self.view)
         self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
 
-    def test_user_has_no_view_permissions(self):
+    def test_user_has_no_view_permissions_or_profile(self):
         self._login()
         response = self.client.get(self.view)
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
-    def test_user_has_view_permissions_no_profile(self):
-        self._login(self.view_permissions)
-        response = self.client.get(self.view)
-        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
-
-    def test_user_has_view_permissions_with_profile(self):
-        Profile.objects.create(user=self.user, team=self.team)
+    def test_user_has_view_permissions_but_no_profile(self):
         self._login(self.view_permissions)
         response = self.client.get(self.view)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        queryset = Order.objects.all().filter(team_id=self.team.id)
+        queryset = Order.objects.all()
 
         # need to provide a request in the serializer context to produce absolute url for image field
         expected_response = OrderListSerializer(
@@ -520,8 +511,23 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
 
         self.assertEqual(expected_response, data["results"])
 
-    def test_user_has_change_permissions(self):
-        self._login(self.all_hardware_permissions)
+    def test_user_has_no_view_permissions_with_profile(self):
+        Profile.objects.create(user=self.user, team=self.team)
+        self._login()
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        queryset = Order.objects.filter(team_id=self.team.id)
+
+        expected_response = OrderListSerializer(
+            queryset, many=True, context={"request": response.wsgi_request}
+        ).data
+        data = response.json()
+
+        self.assertEqual(expected_response, data["results"])
+
+    def test_user_has_view_permissions_and_profile(self):
+        Profile.objects.create(user=self.user, team=self.team)
+        self._login(self.view_permissions)
         response = self.client.get(self.view)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         queryset = Order.objects.all()
@@ -535,7 +541,7 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         self.assertEqual(expected_response, data["results"])
 
     def test_hardware_get_success(self):
-        self._login(self.all_hardware_permissions)
+        self._login(self.view_permissions)
 
         response = self.client.get(self.view)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
@@ -549,7 +555,7 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         self.assertEqual(expected_response, data["results"])
 
     def test_team_id_filter(self):
-        self._login(self.all_hardware_permissions)
+        self._login(self.view_permissions)
 
         url = self._build_filter_url(team_id="1")
 
@@ -562,7 +568,7 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         self.assertCountEqual(returned_ids, [self.order.id, self.order_2.id])
 
     def test_team_code_filter(self):
-        self._login(self.all_hardware_permissions)
+        self._login(self.view_permissions)
 
         url = self._build_filter_url(team_code="ABCDE")
 
@@ -575,7 +581,7 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         self.assertCountEqual(returned_ids, [self.order_3.id, self.order_4.id])
 
     def test_status_filter(self):
-        self._login(self.all_hardware_permissions)
+        self._login(self.view_permissions)
 
         url = self._build_filter_url(status="Cart")
 
@@ -588,7 +594,7 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         self.assertCountEqual(returned_ids, [self.order.id])
 
     def test_created_at_ordering_ascending(self):
-        self._login(self.all_hardware_permissions)
+        self._login(self.view_permissions)
 
         url = self._build_filter_url(ordering="created_at")
         response = self.client.get(url)
@@ -603,7 +609,7 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         )
 
     def test_created_at_ordering_descending(self):
-        self._login(self.all_hardware_permissions)
+        self._login(self.view_permissions)
 
         url = self._build_filter_url(ordering="-created_at")
         response = self.client.get(url)
@@ -618,7 +624,7 @@ class OrderListViewGetTestCase(SetupUserMixin, APITestCase):
         )
 
     def test_search_filter(self):
-        self._login(self.all_hardware_permissions)
+        self._login(self.view_permissions)
 
         url = self._build_filter_url(search="ABCDE")
         response = self.client.get(url)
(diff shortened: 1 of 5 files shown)


hackathon_site/hardware/views.py
```diff
@@ -101,7 +109,10 @@ def get_permissions(self):
         if self.request.method == "POST":
             return [UserHasProfile()]
         if self.request.method == "GET":
-            return [FullDjangoModelPermissions()]
+            if self.request.user.has_perm("hardware.change_order"):
+                return [FullDjangoModelPermissions()]
```
YuYing-Liang: This requires the user to have view permissions and it throws a forbidden error if a user only has change permissions. I don't think this is a big deal, we just have to remember to give tech team view and change permissions.

Also I was trying to combine permissions with `and` and `or` but it wasn't working ://// I'll try to look into it more because this if statement is a bit messy

hackathon_site/hardware/views.py
```diff
@@ -101,7 +109,10 @@ def get_permissions(self):
...
+            else:
+                return [UserHasProfile()]
         return [permissions.IsAuthenticated()]
```
YuYing-Liang, line 104: I think this return is useless ... I will remove

hackathon_site/hardware/test_api.py
```diff
@@ -493,8 +500,28 @@ def test_user_has_no_view_permissions(self):
...
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_user_has_view_permissions_with_profile(self):
+        Profile.objects.create(user=self.user, team=self.team)
+        self._login(self.view_permissions)
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        queryset = Order.objects.all().filter(team_id=self.team.id)
+
+        # need to provide a request in the serializer context to produce absolute url for image field
```
Leo6Leo: Don't forget to remove this comment

hackathon_site/hardware/test_api.py
```diff
@@ -598,7 +627,34 @@ def test_search_filter(self):
...
+        results = data["results"]
+
+        returned_ids = [res["id"] for res in results]
+        self.assertCountEqual(returned_ids, [self.order_2.id])
+
+    def test_filter_from_another_team_with_only_view_permissions(self):
+        Profile.objects.create(user=self.user, team=self.team)
+        self._login(self.view_permissions)
+
+        url = self._build_filter_url(team_code="ABCDE")
```
Leo6Leo: Yuying are you trying to test with this wrong team_code intentionally?

hackathon_site/hardware/views.py
Leo6Leo: Why would you think the return is useless? Do you mean also remove the "else" statement together?

Leo6Leo:
Amazing work Yuying! You have a really great coding habit!!! Detailed explanation on the complex functions/module

hackathon_site/hardware/views.py
YuYing-Liang: I originally thought it was useless because we only have post and get requests for order list view but on second thought I'll keep it in in case we make other requests

hackathon_site/hardware/test_api.py
YuYing-Liang: yeah I'm testing if a user with a profile and only view permissions can view another team's orders, so that's why the assertion is an empty array

hackathon_site/hardware/test_api.py
YuYing-Liang: yes! thanks for catching that :)

hackathon_site/hardware/views.py
```diff
@@ -101,7 +109,10 @@ def get_permissions(self):
         if self.request.method == "POST":
             return [UserHasProfile()]
         if self.request.method == "GET":
-            return [FullDjangoModelPermissions()]
+            if self.request.user.has_perm("hardware.change_order"):
```
grahamhoyes: Can we just Or these two permissions together? `return [FullDjangoModelPermissions() | UserHasProfile()]` might work

hackathon_site/hardware/views.py
```diff
@@ -91,6 +87,18 @@ class OrderListView(generics.ListAPIView):
...
+    def get_queryset(self):
+        queryset = (
+            Order.objects.all().select_related("team")
+            # TODO: Causing problems with queryset aggregations, will figure out later:
+            # .prefetch_related("hardware", "hardware__categories")
+        )
+        if self.request.user.has_perm("hardware.change_order"):
+            return queryset
+        else:
+            user_profile = Profile.objects.get(user_id=self.request.user.id)
```
grahamhoyes: Can just use `self.request.user.profile`

grahamhoyes (commented):
Haven't read through all the tests yet since I've got to go make a turkey, but some minor comments

hackathon_site/hardware/views.py
grahamhoyes: Ok apparently you have to Or the classes together, then instantiate. `return [(FullDjangoModelPermissions | UserHasProfile)()]` is kinda gross, but it works.

Slightly cleaner example here: https://www.django-rest-framework.org/api-guide/viewsets/#introspecting-viewset-actions

hackathon_site/hardware/views.py
```diff
@@ -91,6 +87,18 @@ class OrderListView(generics.ListAPIView):
     ordering_fields = ("created_at",)
     search_fields = ("team__team_code", "id")
 
+    def get_queryset(self):
+        queryset = (
+            Order.objects.all().select_related("team")
+            # TODO: Causing problems with queryset aggregations, will figure out later:
+            # .prefetch_related("hardware", "hardware__categories")
+        )
+        if self.request.user.has_perm("hardware.change_order"):
```
grahamhoyes: For get requests, this should be checking for `hardware.view_order`. Can you wrap this block in an `if self.request.method == "GET"`, then just return queryset by default for other methods

hackathon_site/hardware/views.py
grahamhoyes: That's to be expected - `FullDjangoModelPermissions` checks the relevant permission for each method. Tech team should get change + view permissions for most models.

hackathon_site/hardware/test_api.py
```diff
@@ -493,8 +500,28 @@ def test_user_has_no_view_permissions(self):
         response = self.client.get(self.view)
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
-    def test_user_has_view_permissions(self):
-        self._login(self.permissions)
+    def test_user_has_view_permissions_no_profile(self):
+        self._login(self.view_permissions)
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
```
grahamhoyes: If they have view permissions, they should be given access. Tech team won't have a profile (though technically they could), but should be able to view orders.

Conversely, regular participants won't have view permissions, but they'll get to see their own orders because they have a profile.

hackathon_site/hardware/test_api.py
```diff
@@ -493,8 +500,28 @@ def test_user_has_no_view_permissions(self):
...
+
+    def test_user_has_view_permissions_with_profile(self):
+        Profile.objects.create(user=self.user, team=self.team)
+        self._login(self.view_permissions)
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        queryset = Order.objects.all().filter(team_id=self.team.id)
```
grahamhoyes: Minor: don't need to do `.all()` when doing any other method to `objects`

hackathon_site/hardware/test_api.py
```diff
@@ -493,8 +500,28 @@ def test_user_has_no_view_permissions(self):
...
+
+        # need to provide a request in the serializer context to produce absolute url for image field
+        expected_response = OrderListSerializer(
+            queryset, many=True, context={"request": response.wsgi_request}
+        ).data
+        data = response.json()
+
+        self.assertEqual(expected_response, data["results"])
+
+    def test_user_has_change_permissions(self):
```
grahamhoyes: I left a comment on the view as well, but get requests should be governed only by view permissions

hackathon_site/hardware/test_api.py
```diff
@@ -508,7 +535,7 @@ def test_user_has_view_permissions(self):
         self.assertEqual(expected_response, data["results"])
```
grahamhoyes, line 513: One more test you could add: Make sure that when a user has a profile and view permissions, they can still see everything.

hackathon_site/hardware/test_api.py
```diff
@@ -508,7 +535,7 @@ def test_user_has_view_permissions(self):
...
 
     def test_hardware_get_success(self):
-        self._login(self.permissions)
+        self._login(self.all_hardware_permissions)
```
grahamhoyes: It would be better to keep all of these scoped to just the view permissions that they originally were

hackathon_site/hardware/test_api.py
```diff
@@ -598,7 +627,34 @@ def test_search_filter(self):
         results = data["results"]
 
         returned_ids = [res["id"] for res in results]
-        self.assertCountEqual(returned_ids, [self.order_3.id])
+        self.assertCountEqual(returned_ids, [self.order_3.id, self.order_4.id])
+
+    def test_status_filter_with_only_view_permissions(self):
```
grahamhoyes: We know that filters are applied after calling `get_queryset` - that's something that DRF guarantees for us. We also know that the status filter works (since it's tested above). Hence, there's no need for this test

hackathon_site/hardware/test_api.py
```diff
@@ -598,7 +627,34 @@ def test_search_filter(self):
...
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = response.json()
+        results = data["results"]
+
+        returned_ids = [res["id"] for res in results]
+        self.assertCountEqual(returned_ids, [self.order_2.id])
+
+    def test_filter_from_another_team_with_only_view_permissions(self):
```
grahamhoyes: Again, if a user has view permissions, they should get access to everything - that's what tech team will have.

And as above, we know how the team code filter works, so this shouldn't actually require a test

hackathon_site/hardware/views.py
```diff
@@ -91,6 +86,22 @@ class OrderListView(generics.ListAPIView):
...
+            # TODO: Causing problems with queryset aggregations, will figure out later:
+            # .prefetch_related("hardware", "hardware__categories")
+        )
+        if self.request.method == "GET":
+            if self.request.user.has_perm("hardware.view_order"):
+                return queryset
+            elif hasattr(self.request.user, "profile"):
+                return queryset.filter(team_id=self.request.user.profile.team_id)
+            else:
+                return queryset
```
YuYing-Liang: this is kind of messy but (I don't know why) I think get_queryset runs before get_permissions because for the test case where a user has no permissions or profile, django throws a 'User has no profile' exception. I added this check and the test now passes fine but I don't know how to fix this 🤔
@grahamhoyes @Leo6Leo

hackathon_site/hardware/views.py
```diff
@@ -2,14 +2,15 @@
...
-from django.http import HttpResponseServerError
+from django.http import HttpResponseServerError, HttpResponseForbidden
 from drf_yasg.utils import swagger_auto_schema
 
 from rest_framework import generics, mixins, status, permissions
 from rest_framework.response import Response
 from rest_framework.filters import SearchFilter, OrderingFilter
 
 from event.permissions import UserHasProfile, FullDjangoModelPermissions
+from event.models import Profile
```
grahamhoyes: Unused

hackathon_site/hardware/views.py
```diff
@@ -2,14 +2,15 @@
 
 from django_filters import rest_framework as filters
 from django.db import transaction
-from django.http import HttpResponseServerError
+from django.http import HttpResponseServerError, HttpResponseForbidden
```
grahamhoyes: `HttpResponseForbidden` is also unused now

hackathon_site/hardware/views.py
grahamhoyes: Honestly I think this is fine.

I am starting to question our decision to have the APIs behave differently depending on your authentication level, though. Adding on to the discussion in #320, maybe it would have been cleaner to just leave this endpoint as admin-only and have something like `/api/event/teams/team/orders/` to return the current team's order.

The other problem that this brings up is that if we use this view on the participant side frontend to display their orders, it won't work properly if someone actually has order view permissions. Mostly that will happen when we are developing it locally, which could be really annoying.

What do you think @YuYing-Liang @Leo6Leo? We could keep this implicit filtering, or just start having dedicated endpoints for participants and admins.

YuYing-Liang: Hmmm you are right, I like the idea of separate endpoints and it prevents these weird permission issues. I'll make the changes in this PR

hackathon_site/hardware/views.py
YuYing-Liang: lol why doesn't black remove unused imports 😢

grahamhoyes: Because sometimes it doesn't know if they are actually used. For example our ci.py file seems to have unused imports, but in reality they get used by Django.

I'd like to eventually add Flake8 so we get warnings about unused imports, but there's a lot of cleanup we'd need to do to support it.

hackathon_site/hardware/views.py
grahamhoyes: Cool. Let's make this endpoint require just the Django model permissions then, and add a new one at the path above that filters by the current team. Can do that in this PR if you want, or make another one.

grahamhoyes (approved):
Awesome!

PR #320, opened 2021-10-15
Ieee 66 patch api for profiles

YuYing-Liang:
## Overview

- Resolves [IEEE-66](https://ieeeuoft.atlassian.net/browse/IEEE-66)
- Added a ProfileModifySerializer for new patch api request
- Added ProfileDetailView with permissions and patch request --> viewable to users with profile & users with event.change_profile permissions

## Unit Tests Created

Tested users who...
- are not logged in
- have no permissions and no profile
- have no permissions but have a profile
- have permissions and no profile
- have permissions and profile
- trying to change acknowledge_rules and e_signature after they are already set

## Steps to QA

- pull branch & use swagger to trigger api


hackathon_site/event/serializers.py
```diff
@@ -1,3 +1,5 @@
+from abc import ABC
```
Leo6Leo: Hey Yuying, why import abc?

hackathon_site/event/test_api.py
```diff
@@ -366,3 +366,118 @@ def test_name_search_filter(self):
...
+        self.assertEqual(self._get_expected_response(profile), data)
+
+        # Testing if user can see another user's profile
+        response2 = self.client.patch(
+            self._build_view(self.profile.id), self.request_body
+        )
+        data2 = response2.json()
+
+        # User can only see their own profile even if they query another user's profile id
+        self.assertEqual(response2.status_code, status.HTTP_200_OK)
```
Leo6Leo: If user can only see their own profile, should the status code here be 403 or 401?

hackathon_site/event/serializers.py
YuYing-Liang: Yes whoops, I will remove

hackathon_site/event/test_api.py
YuYing-Liang: True, that makes more sense, I'll change that.

Right now the user sees their own profile regardless of what they query but that could get confusing

hackathon_site/event/views.py
```diff
@@ -226,3 +227,34 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
...
+    def patch(self, request, *args, **kwargs):
+        if self.request.user.has_perm("event.change_profile"):
+            profile = self.get_object()
+        elif request.user.profile:
+            profile = request.user.profile
+        else:
+            return HttpResponseForbidden()
+
+        data = request.data
+        profile.id_provided = data["id_provided"]
```
grahamhoyes: these two fields are something that only admins should be able to check (id provided when they visit the tech team, attended when they sign in on the day-of).

Do you think it would be easier if we leave this endpoint as only accessible to admins, and make a separate `PATCH /event/profiles/profile/`  that regular users can use? Similar to we have `/event/users/user/` and `/events/teams/team/` which deal with the currently logged in user and their team.

hackathon_site/event/api_urls.py
```diff
@@ -13,5 +13,10 @@
         name="join-team",
     ),
     path("teams/leave_team/", api_views.LeaveTeamView.as_view(), name="leave-team"),
+    re_path(
+        r"^profiles/(?P<id>[0-9]+)/$",
```
grahamhoyes: It's usually a convention to do `re_path("profiles/<int:pk>/`. `pk` (primary key) is the default lookup argument for django, which will end up doing an `id` lookup in the database. Saves you from having to set the `lookup_field` in the view

hackathon_site/event/views.py
```diff
@@ -226,3 +227,34 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
 
     def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
+
+
+class ProfileDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
+
+    queryset = Profile.objects.all()
+    serializer_class = ProfileModifySerializer
```
grahamhoyes: This only gets used if you use `mixins.UpdateModelMixin` (which will also run validators on the data). Right now there isn't any actual data validation going on in the plain `patch` method

hackathon_site/event/serializers.py
```diff
@@ -34,6 +36,17 @@ class Meta:
         )
 
 
+class ProfileModifySerializer(serializers.Serializer):
```
grahamhoyes: Use a `serializers.ModelSerializer`, then you don't need to specify the fields manually below. Just add their names to the `Meta.fields` tuple

hackathon_site/event/test_api.py
grahamhoyes: Yeah this is kinda weird, normally it should 403 if given another user's profile.

Taking this plus my comments below, it might be better to make this view admin-only and make a dedicated view for viewing and changing the user's profile like we have with user and team already

hackathon_site/event/views.py
YuYing-Liang: yeah I think that's a lot cleaner

hackathon_site/event/views.py
YuYing-Liang: I'll change RetrieveModelMixin to UpdateModelMixin then

hackathon_site/event/api_urls.py
```diff
@@ -13,5 +13,15 @@
         name="join-team",
     ),
     path("teams/leave_team/", api_views.LeaveTeamView.as_view(), name="leave-team"),
+    re_path(
+        r"^profiles/(?P<pk>[0-9]+)/$",
+        views.ProfileDetailView.as_view(),
+        name="profile-detail",
+    ),
```
```suggestion
    path(
        "profiles/<int:pk>/,
        views.ProfileDetailView.as_view(),
        name="profile-detail",
    ),
```

hackathon_site/event/api_urls.py
```diff
@@ -13,5 +13,15 @@
...
+    path(
+        "profiles/profile",
```
grahamhoyes: Needs a trailing slash (all of our endpoints have them, or should at least)

hackathon_site/event/api_views.py
```diff
@@ -101,3 +106,28 @@ def post(self, request, *args, **kwargs):
...
+
+class CurrentProfileView(mixins.UpdateModelMixin, generics.GenericAPIView):
+
+    queryset = Profile.objects.all()
+    serializer_class = ProfileModifySerializer
+    permission_classes = [UserHasProfile]
+
+    @transaction.atomic
+    def patch(self, request, *args, **kwargs):
+        profile = request.user.profile
```
grahamhoyes: Could you implement a `self.get_object` method that will return this? It then becomes super easy to add a get method that just calls `self.retrieve()` (from `mixins.RetrieveModelMixin`)

hackathon_site/event/api_views.py
```diff
@@ -101,3 +106,28 @@ def post(self, request, *args, **kwargs):
...
         return Response(data=response_data, status=status.HTTP_200_OK,)
+
+
+class CurrentProfileView(mixins.UpdateModelMixin, generics.GenericAPIView):
+
+    queryset = Profile.objects.all()
+    serializer_class = ProfileModifySerializer
+    permission_classes = [UserHasProfile]
+
+    @transaction.atomic
```
grahamhoyes: Don't think this needs to be atomic. This can slow down all atomic views quite a bit since it puts the database in transaction mode.

hackathon_site/event/api_views.py
```diff
@@ -101,3 +106,28 @@ def post(self, request, *args, **kwargs):
...
+    def patch(self, request, *args, **kwargs):
+        profile = request.user.profile
+
+        data = request.data
```
grahamhoyes: I think there is a way to do updates with Serializers? Just specify which fields are read-only. That way you get validation that the fields are of the correct type etc

hackathon_site/event/views.py
```diff
@@ -226,3 +227,28 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
 
     def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
+
+
+class ProfileDetailView(mixins.UpdateModelMixin, generics.GenericAPIView):
```
grahamhoyes: Can you move this over to `api_views.py`?

hackathon_site/event/api_views.py
```diff
@@ -101,3 +106,28 @@ def post(self, request, *args, **kwargs):
...
+        profile.id_provided = data["id_provided"]
```
grahamhoyes: Since this endpoint is the one particpants will be using, they shouldn't be able to set `id_provided` and `attended`.

Methinks the best thing would be to create two serializers, one for this view and one for the one that tech team uses. You might be able to subclass the existing profile serialzier, and just set which fields are read-only in each case.

hackathon_site/event/api_views.py
YuYing-Liang: should I remove `mixins.UpdateModelMixin` then? or add `mixins.RetrieveModelMixin`? or is this just for the future for when we want to make a get request?

grahamhoyes: For now this is fine, eventually we can add a get method and the retrieve model mixin

hackathon_site/event/serializers.py
```diff
@@ -32,6 +32,15 @@ class Meta:
             "e_signature",
             "team",
         )
+        read_only_fields = ("id", "team")
+
+
+class UserProfileSerializer(ProfileSerializer):
```
grahamhoyes: Minor, would call this `CurrentProfileSerializer` just to be consistent with the view.

hackathon_site/event/api_views.py
```diff
@@ -101,3 +106,53 @@ def post(self, request, *args, **kwargs):
...
+class CurrentProfileView(mixins.UpdateModelMixin, generics.GenericAPIView):
+
+    queryset = Profile.objects.all()
+    serializer_class = UserProfileSerializer
+    permission_classes = [UserHasProfile]
+
+    def get_object(self):
+        return self.request.user.profile
+
+    def patch(self, request, *args, **kwargs):
```
grahamhoyes, line 131: Since you have `UpdateModelMixin`, you can use the `self.partial_update` method instead of doing this logic yourself:

```
def patch(self, request, *args, **kwargs):
    return self.partial_update(request, *args, **kwargs)
```

In order to do that, you'll need to move the logic into the serializer. There are two places you could do that: in the `validate` method (which returns the validated data), or in the `update` method. In this case, since we're not actually validating anything, the `update` method should be fine. 

Here's how you'd move that logic into and `update` method on the serializer (I think, this is untested):

```python
    def update(self, instance: Profile, validated_data):
        acknowledge_rules = validated_data.pop("acknowledge_rules", False)
        e_signature = validated_data.pop("e_signature", None)

        if not instance.acknowledge_rules and acknowledge_rules:
            instance.acknowledge_rules = acknowledge_rules
        if not instance.e_signature and e_signature:
            instance.e_signature = e_signature

        return super().update(instance, validated_data)
```

That's also safe against them only submitting one field or the other.

hackathon_site/event/api_views.py
```diff
@@ -101,3 +106,53 @@ def post(self, request, *args, **kwargs):
...
+
+
+class ProfileDetailView(mixins.UpdateModelMixin, generics.GenericAPIView):
+
+    queryset = Profile.objects.all()
+    serializer_class = ProfileSerializer
+    permission_classes = [FullDjangoModelPermissions]
+
+    @transaction.atomic
+    def patch(self, request, *args, **kwargs):
```
grahamhoyes, line 118: This can be similar to the `CurrentProfileView`, in that it can use `self.partial_update`.

I don't think this one will require any special `update` method on the `ProfileSerializer`. In theory tech team shouldn't be modifying the `e_signature` or `acknowledge_rules` fields, but since we may eventually build the whole admin functionality into react we can keep it.

hackathon_site/event/test_api.py
```diff
@@ -366,3 +366,136 @@ def test_name_search_filter(self):
...
+
+        # First, normally update profile
+        response = self.client.patch(self.view, self.request_body)
+        data = response.json()
+        expected_response = self.expected_response
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(expected_response, data)
+
+        # Then update profile with a different request, changing acknowledge & e signature
+        new_request_body = {
```
grahamhoyes: One test you'll want to have: What if they only provide a single field? I think the current code will crash.

hackathon_site/event/api_views.py
```diff
@@ -1,18 +1,19 @@
...
-from django.shortcuts import get_object_or_404
-from django_filters import rest_framework as filters
 
 from rest_framework import generics, mixins, status
 from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
-from rest_framework.filters import SearchFilter
 
 
+from event.models import Profile
```
grahamhoyes: Minor, combine these two imports with the ones below

hackathon_site/event/serializers.py
```diff
@@ -32,6 +32,43 @@ class Meta:
...
         )
+        read_only_fields = ("id", "team")
+
+    def update(self, instance: Profile, validated_data):
+        id_provided = validated_data.pop("id_provided", False)
+        attended = validated_data.pop("attended", False)
+        acknowledge_rules = validated_data.pop("acknowledge_rules", False)
+        e_signature = validated_data.pop("e_signature", None)
+
+        if id_provided is not None:
```
grahamhoyes: I think you don't actually need to have the `id_provided` and `attended` in this method (no need to pop them from validated data either). They'll be handled by the `super().update()` call, since there's no special logic associated with them

hackathon_site/event/serializers.py
```diff
@@ -32,6 +32,43 @@ class Meta:
             "e_signature",
             "team",
         )
+        read_only_fields = ("id", "team")
+
+    def update(self, instance: Profile, validated_data):
+        id_provided = validated_data.pop("id_provided", False)
+        attended = validated_data.pop("attended", False)
+        acknowledge_rules = validated_data.pop("acknowledge_rules", False)
```
grahamhoyes: Now that I think about it, should this serializer even be allowed to modify `acknowledge_rules` and `e_signature`? Those are somewhat important fields, it might be a bad idea to have them editable over the API. Thoughts?

If you concur, you can add them to the read only fields for this serializer.

arshinmar (dismissed):

hackathon_site/event/serializers.py
```diff
@@ -32,6 +32,31 @@ class Meta:
...
+        return super().update(instance, validated_data)
+
+
+class CurrentProfileSerializer(ProfileSerializer):
+    class Meta(ProfileSerializer.Meta):
+        read_only_fields = (
+            "id",
+            "team",
+            "id_provided",
+            "attended",
```
grahamhoyes, line 47: One last change, you'll need to add `team` to this too (since there are dedicated views with special logic for changing teams)

hackathon_site/event/test_api.py
```diff
@@ -370,6 +370,150 @@ def test_name_search_filter(self):
         self.assertCountEqual(returned_ids, [self.team2.team_code])
 
 
+class ProfileDetailViewTestCase(SetupUserMixin, APITestCase):
```
grahamhoyes, line 373: Could you also add a quick round of tests for each of these views that modifying the read only fields throws a validation error (status 400)? You can just do a single test method for each with a for loop or something to make the requests

hackathon_site/event/serializers.py
YuYing-Liang: Isn't team already there?

grahamhoyes: I am completely blind apparently :man_facepalming:

hackathon_site/event/serializers.py
```diff
@@ -32,6 +32,36 @@ class Meta:
             "e_signature",
             "team",
         )
+        read_only_fields = ("id", "team", "acknowledge_rules", "e_signature")
+
+    def update(self, instance: Profile, validated_data):
+        if not validated_data:
```
YuYing-Liang: I had to raise a custom error because modifying read only fields/having an empty request doesn't automatically throw an error

hackathon_site/event/serializers.py
```diff
@@ -32,6 +32,35 @@ class Meta:
...
+            raise serializers.ValidationError("no modifiable fields provided")
+        acknowledge_rules = validated_data.pop("acknowledge_rules", False)
+        e_signature = validated_data.pop("e_signature", None)
+
+        if not instance.acknowledge_rules and acknowledge_rules:
+            instance.acknowledge_rules = acknowledge_rules
+        if not instance.e_signature and e_signature:
+            instance.e_signature = e_signature
+
+        return serializers.ModelSerializer.update(self, instance, validated_data)
```
YuYing-Liang, line 61: had to use the original update method because calling super() would call the ProfileSerializer update method and since we pop off the acknowledge_rules and e_signature, it would always throw a validation error

hackathon_site/event/serializers.py
grahamhoyes: That's fine actually, the empty request part is valid since this is a patch request (meaning partial update). You can remove this error.

I didn't realize that submitting read only fields didn't throw an error though 🤔. Maybe it's easier to have a quick test case that just asserts what the read only fields (`Serializer.read_only_fields`) are on the serializer directly, so that nobody accidentally changes them.

hackathon_site/event/serializers.py
grahamhoyes: Good catch! Could you add in a comment explaining that?

hackathon_site/event/serializers.py
YuYing-Liang: what do you mean by on the serializer directly?

grahamhoyes: I mean add something in the test case like:

```
def test_readonly_serializer_fields(self):
    self.assertEqual(CurrentTeamOrderListView.read_only_fields, ("id", "team", "acknowledge_rules", "e_signature"))
```

You can put that in the test class for the view.

We know that DRF will use the read_only_fields properly, we just want a test to make sure nobody accidentally changes it without considering the implications. 

You can go ahead and remove this `if not validate_data` check as well, since I guess it conforms to the semantics of a PATCH request as it was.

hackathon_site/event/test_api.py
grahamhoyes: Per the comments above, it looks like DRF's default behaviour is to just ignore the read only fields.

That's fine by me - so we can either remove those tests (sorry :( ) and add the one above that would assert on `CurrentTeamOrderListView.read_only_fields` directly, or you could modify the tests you added to make sure that the read only fields weren't modified (but the former is probably easier).

grahamhoyes (approved):

PR #321, opened 2021-10-17
Convert Cart Page and Components to Typescript, Connect CartCard to Redux

grahamhoyes:
## Overview

Resolves IEEE-112

Converts the `CartCard` component and `Cart` page to Typescript, and connects the former to redux to grab hardware. This was more of a rewrite than a simple refactor, since the (implicit) interfaces we assumed for the cart originally don't match up with what we have decided (and can now enforce with Typescript).

The `CartCard` component now just takes a hardware ID and quantity, and pulls the hardware from the redux store. I wanted it to just take a regular `Hardware` object, but pulling the relevant objects from the store inside the `Cart` component was rather difficult and would have led to inefficient re-renders of the entire `Cart` component in some cases. The downside is that tests are a little harder to write this way.

A new `makeStoreWithEntities` test helper was added, which pre-fills the store with entities (currently only supports hardware and categories, more have to be added manually). This does rely a bit on knowing the internal workings of redux, but typescript will yell at us if that ever changes.

Finally, adds an argument to the `getHardwareWithFilters` thunk to allow preserving existing hardware in the store. This is used by the cart page if anything new needs to be fetched.

## Unit Tests Created

- Added a new `makeStoreWithEntities` helper function
- Updated the `Cart` and `CartCard` tests. Some tests were removed since they'll need to change with redux, comments left in their place.
- New test that the Cart page fetches missing hardware on load
- Some extra basic tests for the hardware and category thunks, to make using the new `makeStoreWithEntities` safer

## Steps to QA

- The current mocked cart items use hardware with IDs 1, 2, and 3, so make sure you have at least those added (via the admin site) otherwise weird stuff will happen. Then just visit the cart page.


hackathon_site/dashboard/frontend/src/components/cart/CartCard/CartCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+import CloseIcon from "@material-ui/icons/Close";
+
+import { useSelector } from "react-redux";
+import { RootState } from "slices/store";
+import { hardwareSelectors } from "slices/hardware/hardwareSlice";
+
+const makeSelections = (quantity_remaining: number) => {
+    const items = [];
+
+    for (let i = 1; i < quantity_remaining + 1; i++) {
```
YuYing-Liang, line 22: would map or forEach work better here?
`[...Array(quantity_remaining).keys()].map(i => <> ... </>)`

hackathon_site/dashboard/frontend/src/pages/Cart/Cart.tsx
```diff
@@ -1,40 +1,34 @@
...
 import CartSummary from "components/cart/CartSummary/CartSummary";
-import { cartItems, cartQuantity } from "testing/mockData";
-
-const handleRemove = (id) => {
-    alert("Remove item with id " + id + " from the cart");
-};
+import { cartItems } from "testing/mockData";
 
 const Cart = () => {
+    const cartQuantity = cartItems.reduce((accum, item) => accum + item.quantity, 0);
```
YuYing-Liang: good use of reduce

hackathon_site/dashboard/frontend/src/testing/utils.tsx
```diff
@@ -81,3 +84,37 @@ export const makeMockApiListResponse = <T extends unknown>(
...
+}
+
+/**
+ * Given a store, a collection of entities, and a mocked api.get function,
+ * prepopulate the store by emulating API responses.
+ *
+ * Due to the asynchronous nature of dispatches, tests using this function
+ * may need to wait after calling it for the store to be populated.
+ */
+export const fillStoreWithEntities = (
```
YuYing-Liang: should we change the inventory unit tests to use this too?

hackathon_site/dashboard/frontend/src/components/cart/CartCard/CartCard.tsx
grahamhoyes: I was going to do that originally, but I thought it would look weird to people. But since you suggested it, I'll do it!

hackathon_site/dashboard/frontend/src/testing/utils.tsx
grahamhoyes: I'm not really happy with this function tbh, it's still annoying to use. I wanted to just manually insert the entities into the store (saves having to wait for the thunks to propagate), but I was worried that we'd then end up not properly testing that components do end up hitting the API.

I think that as long as we have some tests for components that hit the APIs, it's fine to just prepopulate the state for the rest of them. I'm going to add a few direct tests for the thunks which should also cover the API functionality, then change this function to not use mock API responses anymore.

YuYing-Liang (dismissed):

hackathon_site/dashboard/frontend/src/components/cart/CartCard/CartCard.tsx
grahamhoyes: Changed my mind, this actually became quite messy for some reason. Sticking with the easier to read for loops!

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -116,13 +116,21 @@ const hardwareSlice = createSlice({
...
-        });
+        builder.addCase(
+            getHardwareWithFilters.fulfilled,
+            (state, { payload, meta }) => {
+                state.isLoading = false;
+                state.error = null;
+                state.next = payload.next;
+                state.count = payload.count;
+
+                if (meta.arg?.keepOld) {
```
YuYing-Liang, line 127: When would we use keep old? refresh button?

YuYing-Liang:
Looks good! I think this can be done in a different PR but the Product Overview panel limits the quantity dropdown based on the hardware & category constraints (max_per_team) so it'd be nice to do it for the cart page too

YuYing-Liang (approved):

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
grahamhoyes: I use it on the cart page when some hardware is missing (which will happen if we eventually only persist the cart to local storage). Rather than replacing all the hardware in the store, I just fetch the ones that are missing.

Originally I wrote a separate thunk for it, but this was far easier.

grahamhoyes:
> Looks good! I think this can be done in a different PR but the Product Overview panel limits the quantity dropdown based on the hardware & category constraints (max_per_team) so it'd be nice to do it for the cart page too

Good point! Made IEEE-136 for that

PR #322, opened 2021-10-22
IEEE-26: added get next hardware page thunk and unit test

YuYing-Liang:
## Overview

- Resolves #IEEE-26
- Added getNextHardwarePage thunk
- Added reducers for this thunk to update adapter with more hardware

[Short demo here](https://share.vidyard.com/watch/1wk8okwUbY1WfMdqhYZ1q6?)

## Unit Tests Created

- Test to see if load more buttons loads more hardware

## Steps to QA


hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -108,4 +111,48 @@ describe("Inventory Page", () => {
...
+            .mockResolvedValue(categoryApiResponse);
+
+        const { getByText } = render(<Inventory />);
+
+        await waitFor(() => {
+            expect(getByText(mockHardware[0].name)).toBeInTheDocument();
+        });
+
+        fireEvent.click(getByText(/load more/i));
+        await waitFor(() => {
```
YuYing-Liang, line 163: these assertions are failing but I don't know why :(

hackathon_site/dashboard/frontend/src/api/api.ts
```diff
@@ -58,6 +58,8 @@ export const get = <T>(
 
     if (params) {
         uri += "?" + new URLSearchParams(params).toString();
+    } else {
+        uri = uri.endsWith("/") ? uri.substr(0, uri.length-1) : uri;
```
grahamhoyes: Our URIs have to end with a slash (unless I misreviewed a PR somewhere and we missed one)

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -6,6 +6,9 @@ import Inventory from "pages/Inventory/Inventory";
 import { mockCategories, mockHardware } from "testing/mockData";
 
 import { get } from "api/api";
+import { fireEvent } from "@testing-library/react";
```
grahamhoyes: Minor, can import this from `"testing/utils"`, since that re-exports everything from RTL

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -6,6 +6,9 @@ import Inventory from "pages/Inventory/Inventory";
...
+import { makeStore } from "../../slices/store";
```
grahamhoyes: Use absolute imports instead of relative imports

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -68,6 +68,32 @@ export const getHardwareWithFilters = createAsyncThunk<
...
+export const getHardwareNextPage = createAsyncThunk<
+    APIListResponse<Hardware>,
+    void,
+    { state: RootState; rejectValue: RejectValue; dispatch: AppDispatch }
+    >(
+        `${hardwareReducerName}/getHardwareNextPage`,
+    async (_, { dispatch, getState, rejectWithValue }) => {
+        try {
+            const nextURL = stripHostname(hardwareNextSelector(getState()) ?? "")
+            const response = await get<APIListResponse<Hardware>>(nextURL);
```
grahamhoyes: If `nextURL` is empty, this will end up just fetching the home page I think. Needs some way of not running if the `next` field in the state is none.

Two options for that:
- Return nothing (I think this will require manually building an axios response because typescript)
- Reject and display a snackbar saying there's no more hardware to fetch

The first option would be the cleanest I think, if you can get it working.

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -128,6 +154,23 @@ const hardwareSlice = createSlice({
...
+
+        builder.addCase(getHardwareNextPage.pending, (state) => {
+            state.isLoading = true;
+            state.error = null;
+        });
+
+        builder.addCase(getHardwareNextPage.fulfilled, (state, { payload }) => {
+            state.isLoading = false;
+            state.error = null;
+            state.next = payload.next;
```
grahamhoyes: Should also update `state.count`, just in case hardware is added/removed and the count changes or something.

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
grahamhoyes: If you want to test that everything is going through, I think you can print out `mockedGet.mock.calls` to make sure it's being called enough times. You can also do an `expect(mockedGet).toHaveBeenCalledWith(stuff)` to test that the calls are working

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
```diff
@@ -117,7 +126,11 @@ const Inventory = () => {
                                     <Typography variant="body2">
                                         {count} results
                                     </Typography>
-                                    <IconButton color="primary" aria-label="Refresh">
+                                    <IconButton
```
grahamhoyes, line 132: Need a quick test that this button works too!

grahamhoyes:
One thing I noticed while testing this (which you can do by changing the page size [here](https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/hackathon_site/settings/__init__.py#L174)): When we're loading more hardware, the entire hardware list gets replaced with a loading wheel.

When there's a hundred items there and we're fetching a hundred more, that might get annoying. I think it might be best to add a new `isMoreLoading` key to the store, and use that to stick a loading wheel in the "load more" button and leave the `isLoading` key for when we're replacing everything?

hackathon_site/dashboard/frontend/src/api/api.ts
YuYing-Liang: the reason why I added this is because cleanURI function adds a "/" at the end of the endpoint but this is the "next" property in the response which already has the queries attached to it and doesn't need that extra "/"

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: I will try to remember to do this all the time lol

hackathon_site/dashboard/frontend/src/api/api.ts
grahamhoyes: Hmm good point, but the problem with this is that if you make a get request without any params (like to `/api/hardware/hardware/`), the trailing slash we added is going to be removed.

Do you think you could fix up the actual `cleanURI` function, or come up with an alternative instead? Something based around a `new URL` object could be done instead.

Perhaps we make a new `makeFullURL` function, that takes in a URI (just a path, possibly with query params) and:
- Removes the leading slash
- Does ``url = `${SERVER_URL}/${uri}` ``
- Parses the resulting url with `new URL`, makes sure `pathname` ends with a slash, then reconstructs the URL like we do in `stripHostname`

Here's a quick dump from a testing console:

```
> uri = 'http://localhost:8000/api/hardware/?foo=bar'
'http://localhost:8000/api/hardware/?foo=bar'
> parsed = new URL(uri)
URL {
  href: 'http://localhost:8000/api/hardware/?foo=bar',
  origin: 'http://localhost:8000',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost:8000',
  hostname: 'localhost',
  port: '8000',
  pathname: '/api/hardware/',
  search: '?foo=bar',
  searchParams: URLSearchParams { 'foo' => 'bar' },
  hash: ''
}
> uri = 'http://localhost:8000/api/hardware?foo=bar'
'http://localhost:8000/api/hardware?foo=bar'
> parsed = new URL(uri)
URL {
  href: 'http://localhost:8000/api/hardware?foo=bar',
  origin: 'http://localhost:8000',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost:8000',
  hostname: 'localhost',
  port: '8000',
  pathname: '/api/hardware',
  search: '?foo=bar',
  searchParams: URLSearchParams { 'foo' => 'bar' },
  hash: ''
}
```

YuYing-Liang: I made a stripHostnameReturnFilters function to return the searchParams, it works, but code is kind of ugly lol

hackathon_site/dashboard/frontend/src/api/api.ts
```diff
@@ -60,6 +79,7 @@ export const get = <T>(
         uri += "?" + new URLSearchParams(params).toString();
     }
 
+    console.log(uri);
```
YuYing-Liang: i will remove

hackathon_site/dashboard/frontend/src/api/api.ts
YuYing-Liang: fyi, I tried changing the page size to 1 (just so there would be many api calls) and the categories api is called infinitely, I printed out the url that was being called and it also has an extra / at the end. I tried fixing it but the unit tests started failing so I left that out lol

hackathon_site/dashboard/frontend/src/api/api.ts
```diff
@@ -50,6 +50,25 @@ export const stripHostname = (url: string): string => {
     return parsed.pathname + parsed.search + parsed.hash;
 };
 
+export const stripHostnameReturnFilters = (
+    url: string | null
+): { path: string; filters?: { [key: string]: any } } => {
+    if (url) {
+        const parsed = new URL(url);
+        const filters: { [key: string]: any } = {};
+        if (parsed.search && filters) {
```
grahamhoyes, line 59: The `&& filters` here isn't doing anything, since the truthyness of an empty object in Javascript is true for some reason

hackathon_site/dashboard/frontend/src/api/api.ts
```diff
@@ -50,6 +50,25 @@ export const stripHostname = (url: string): string => {
...
+            parsed.searchParams.forEach((val, key) => (filters[key] = val));
+        }
+        return JSON.stringify(filters) !== "{}"
```
grahamhoyes: `Object.keys(filters).length !== 0`

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -1,14 +1,24 @@
...
 
 import Inventory from "pages/Inventory/Inventory";
 import { mockCategories, mockHardware } from "testing/mockData";
 
-import { get } from "api/api";
+import { get, stripHostnameReturnFilters } from "api/api";
 
 jest.mock("api/api");
 const mockedGet = get as jest.MockedFunction<typeof get>;
+const mockStripHostnameReturnFilters =
```
grahamhoyes: Do we have to mock this? I think it would be better to just mock the API endpoints (via `get`), but leave this function as-is.

If you still want to mock this function, you should add an explicit unit test for it to be safe.

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -68,6 +70,55 @@ export const getHardwareWithFilters = createAsyncThunk<
...
+    `${hardwareReducerName}/getHardwareNextPage`,
+    async (_, { dispatch, getState, rejectWithValue }) => {
+        try {
+            const nextFromState = hardwareNextSelector(getState()) ?? "";
+            if (nextFromState) {
+                const { path, filters } = stripHostnameReturnFilters(nextFromState);
+                const response = await get<APIListResponse<Hardware>>(path, filters);
+                return response.data;
+            }
+            // return empty response if there is no nextURL
```
grahamhoyes, line 87: Not so sure this is a good idea. If there is no nextURL, we should just be returning nothing. You can update the return type of the thunk (line 74) to be `APIListResponse<Hardware> | null`, and handle the null case accordingly in the reducer.

This fake response is going to end up buing inserted into the redux store.

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
```diff
@@ -126,6 +179,28 @@ const hardwareSlice = createSlice({
 
         builder.addCase(getHardwareWithFilters.rejected, (state, { payload }) => {
             state.isLoading = false;
+            state.isMoreLoading = false;
+            state.error = payload?.message || "Something went wrong";
+        });
+
+        builder.addCase(getHardwareNextPage.pending, (state) => {
+            state.isLoading = false;
```
grahamhoyes, line 174: Since this thunk doesn't touch `isLoading` anymore, you don't actually have to explicitly set it in any of its reducers. Same with the `getHardwareWithFilters` reducers, you don't need to be setting `state.isMoreLoading` (unless I'm missing something?)

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
```diff
@@ -108,4 +118,91 @@ describe("Inventory Page", () => {
...
+            null,
+            hardwareUri,
+            mockHardware.length
+        );
+        const categoryApiResponse = makeMockApiListResponse(mockCategories);
+
+        when(mockedGet)
+            .calledWith(hardwareUri, {})
+            .mockResolvedValue(hardwareApiResponse);
+        when(mockStripHostnameReturnFilters)
```
grahamhoyes: Going off the comment above, I don't think you should be mocking this. You know what data is being passed in to this function, so you should know what it's output is. There aren't any side-effects either, so it's safe to not mock I think?

Even better, you can do a `expect(mockedGet).toHaveBeenCalledWith(/* whatever you expect stripHostnameReturnFiletrs to return */)`, that way you have a test for that function.

hackathon_site/dashboard/frontend/src/api/api.ts
grahamhoyes: Good catch, the category API thunk is definitely broken! Not even sure how we would write a unit test that catches the infinite case :thinking:. I made IEEE-128 for us to fix that.

hackathon_site/dashboard/frontend/src/api/api.ts
YuYing-Liang: but filters is an optional argument

hackathon_site/dashboard/frontend/src/api/api.ts
YuYing-Liang: I thought of doing this but Object.keys() is expensive right because it creates an array?

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: yeah so I didn't think I needed to mock this either but the unit tests will fail if I don't :( I started logging a bunch of things to figure out why it was failing and the url is undefined after passing through that function 😔 I'm not sure why though but mocking it fixed that 🤷‍♀️

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: ah okay, going off my reply to your comment above lol I'll try this to see if it solved my problem

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
YuYing-Liang: okay, I'll try that

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
YuYing-Liang: yeah true, I was setting them just in case but I guess it's redundant

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
YuYing-Liang: nvm lol it doesn't work :( I need the when()

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
YuYing-Liang: do i need to test the slice for this case?

hackathon_site/dashboard/frontend/src/api/api.ts
grahamhoyes: D'oh, good point :man_facepalming:

hackathon_site/dashboard/frontend/src/api/api.ts
grahamhoyes: Tbh I'm not sure if it actually creates a new array, or if it is doing an efficient iterator over the keys that are already stored. That being said, `Object.keys(obj).length === 0` is the most common way of checking for an empty object that I've seen.

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
grahamhoyes: That would make me think that there might be something wrong with the function if it is returning undefined, since that isn't even a valid return type for it. If it helps, set a breakpoint in there and then run the test in the pycharm debugger!

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts
grahamhoyes: As long as there's a test somewhere that triggers the return null case, you're good (even if it's part of the inventory tests)

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.test.tsx
grahamhoyes: Aha, I know why it's returning undefined! When you do `jest.mock("api/api")`, that actually mocks everything in that module (mocking is the only thing I really don't like about Jest, it's so much easier in Python).

I got around this in the [category slice tests](https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/dashboard/frontend/src/slices/hardware/categorySlice.test.ts#L18) by mocking just the get function. Or rather, I restore all the exports to their original implementation and only mock get:

```javascript
jest.mock("api/api", () => ({
    ...jest.requireActual("api/api"),
    get: jest.fn(),
}));
```

Should work for you, and allow you to use the actual `stripHostnameReturnFilters`.

YuYing-Liang:
I accidentally ran prettier on some config files 😬

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.test.ts
```diff
@@ -7,16 +7,19 @@ import {
     hardwareReducerName,
     initialState,
     getHardwareWithFilters,
-    isLoadingSelector,
```
YuYing-Liang, line 10: removed because it wasn't being used

hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.test.ts
```diff
@@ -103,3 +106,67 @@ describe("getHardwareWithFilters thunk", () => {
...
     });
 });
+
+describe("getHardwareNextPage thunk", () => {
+    it("Updates the store on API success", async () => {
+        const limit = mockHardware.length - 2;
+        const nextURL = `http://localhost:8000/api/hardware/hardware/?offset=${limit}`;
+        const { path, filters } = stripHostnameReturnFilters(nextURL);
+
+        const hardwareWithFiltersResponse = makeMockApiListResponse(
```
YuYing-Liang, line 116: idk how to pre-populate the store so I mocked getHardwareWithFilters for these 2 new tests

grahamhoyes: You can use `const store = makeStoreWithEntities({ hardware: yourHardwareArray })`

grahamhoyes: Nevermind, since that doesn't actually populate the next URL you can stick with the mocked responses!

hackathon_site/dashboard/frontend/src/pages/Inventory/Inventory.tsx
```diff
@@ -142,8 +160,13 @@ const Inventory = () => {
                                 size="large"
                                 fullWidth={true}
                                 disableElevation
+                                onClick={getMoreHardware}
                             >
-                                Load more
+                                {isMoreLoading ? (
+                                    <CircularProgress size={25} />
```
grahamhoyes: Can you tweak the test to also check that this appears, and then disappears once the data is loaded? You'll need to add a test id to it.

That becomes our integration-y test that the `isMoreLoading` selector is actually working. I do something similar here: https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/dashboard/frontend/src/components/inventory/InventoryFilter/InventoryFilter.test.tsx#L157

grahamhoyes (commented):
One tiny thing that still needs to be tested, but everything works great!

grahamhoyes (approved):

PR #323, opened 2021-10-23
IEEE-15: POST and IEEE-138: POST Incidents

arshinmar:
## Overview
 
- Resolves #IEEE-15  and #IEEE-138 (posting Incidents)
- Created a post view titled IncidentListView in the event folder.
- Reconfigured permissions method for hardware API to appropriately handle Incidents.
- For event team, overrode the perform_create method to ensure posting team is the incident's team.
- Event api is for UserHasProfile permission and Hardware api is for FullDjangoModel permissions.

## Unit Tests Created

- For hardware, made sure that you had the right permissions to post an incident.
- For event, made sure you had to have a profile and you could only post an incident for your own team.
- Also made sure that you had to be logged in.

## Steps to QA

- Test in swagger, and make sure tests are formatted properly.


hackathon_site/hardware/serializers.py
```diff
@@ -89,6 +88,111 @@ def get_team_code(obj: Order):
         return obj.team.team_code
 
 
+class IncidentCreateSerializer(serializers.Serializer):
```
grahamhoyes: This is waaaaaay more than you need. I realize it was copy-pasted from the Order serializer, but why? Orders was an extremely custom use case (that involved creating a bunch of `OrderItems` under the hood). Creating an Incident is very straightforward, and just requires a barebones serializer that looks like:

```python
class IncidentCreateSerializer(serializers.ModelSerializer):
    class Meta:
        model = Incident
        fields = (
            "id",
            "state",
            "time_occurred",
            "description",
            "order_item",
            "created_at",
            "updated_at",
        )
```

I forget how nested serializers are handled during create calls, you might actually be able to just tuse the pre-existing `IncidentSerializer` and just add `read_only_fields = ("team_id",)` to the Meta class.

hackathon_site/hardware/views.py
```diff
@@ -62,9 +69,31 @@ class IncidentListView(mixins.ListModelMixin, generics.GenericAPIView):
...
+            return self.serializer_method_classes[self.request.method]
+        except (KeyError, AttributeError):
+            return super().get_serializer_class()
+
     def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
 
+    @transaction.atomic
+    @swagger_auto_schema(responses={201: IncidentCreateResponseSerializer})
+    def post(self, request, *args, **kwargs):
```
grahamhoyes, line 75: This view should use `mixins.CreateModelMixin`, then this view can just be:

```python
def post(self, request, *args, **kwargs):
```

hackathon_site/hardware/views.py
```diff
@@ -62,9 +69,31 @@ class IncidentListView(mixins.ListModelMixin, generics.GenericAPIView):
...
+    def get_serializer_class(self):
+        try:
+            return self.serializer_method_classes[self.request.method]
+        except (KeyError, AttributeError):
+            return super().get_serializer_class()
+
     def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
 
+    @transaction.atomic
```
grahamhoyes: This view doesn't need to be atomic, there aren't any race conditions

hackathon_site/hardware/views.py
```diff
@@ -62,9 +69,31 @@ class IncidentListView(mixins.ListModelMixin, generics.GenericAPIView):
     filter_backends = (filters.DjangoFilterBackend, SearchFilter)
     filterset_class = IncidentFilter
 
+    def get_serializer_class(self):
```
grahamhoyes, line 66: tbh I don't remember why we implemented this in the Order view. We should have just had an if/elif statement on `self.request.method` in here and returned the appropriate one, rather than having the map and try/except.

grahamhoyes (commented):
There's no need for this method to be as complicated as the order endpoint, you can just use [`CreateModelMixin`](https://www.django-rest-framework.org/api-guide/generic-views/#createmodelmixin) and some simple serializer logic.

It also doesn't really make sense to start by copy-pasting hundreds of lines of code from another (very different) endpoint without first considering what the use case for this one is, or asking for help if you need it.

hackathon_site/hardware/test_api.py
```diff
@@ -607,6 +608,58 @@ def test_search_filter(self):
...
+    def test_incorrect_permissions(self):
+        self._login()
+        response = self.client.post(self.view, self.request_data, format="json")
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_successful_post(self):
+        self._login(self.permissions)
+        self.profile = Profile.objects.create(user=self.user, team=self.team)
+        response = self.client.post(self.view, self.request_data, format="json")
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
YuYing-Liang, line 667: make an assertion on the response data as well

hackathon_site/hardware/views.py
```diff
@@ -46,11 +47,13 @@ def get(self, request, *args, **kwargs):
...
 
-class IncidentListView(mixins.ListModelMixin, generics.GenericAPIView):
+class IncidentListView(
+    mixins.ListModelMixin, mixins.CreateModelMixin, generics.GenericAPIView
+):
     queryset = Incident.objects.all().select_related(
         "order_item", "order_item__order__team", "order_item__hardware"
     )
-    serializer_class = IncidentSerializer
+    serializer_class = IncidentListSerializer
```
YuYing-Liang: do you need this if you have the get_serializer_class() method? I think that method overwrites serializer_class property

YuYing-Liang: hmm actually, nvm keep it, it defaults to that serializer if we're not making a get or post request

grahamhoyes: I'd probably agree, you can remove it if you have a `get_serializer_class` method. And if there's no method on the view for the other HTTP verbs, they API will give a 405 Method Not Allowed anyway.

hackathon_site/event/api_urls.py
```diff
@@ -14,6 +14,11 @@
     ),
     path("teams/leave_team/", api_views.LeaveTeamView.as_view(), name="leave-team"),
     path("teams/", views.TeamListView.as_view(), name="team-list"),
+    path(
+        "teams/team/incidents",
```
```suggestion
        "teams/team/incidents/",
```

The trailing slash strikes again!

hackathon_site/event/api_views.py
```diff
@@ -109,6 +109,25 @@ def post(self, request, *args, **kwargs):
...
 
+class IncidentListView(
+    mixins.ListModelMixin, mixins.CreateModelMixin, generics.GenericAPIView
+):
+    queryset = Incident.objects.all()
+
+    serializer_class = IncidentCreateSerializer
+    permission_classes = [UserHasProfile]
+
+    def perform_create(self, serializer):
```
grahamhoyes, line 121: Do you need a call to `super().perform_create(serializer)` at the end of this method?

hackathon_site/event/api_views.py
```diff
@@ -109,6 +109,25 @@ def post(self, request, *args, **kwargs):
         return Response(data=response_data, status=status.HTTP_200_OK,)
 
 
+class IncidentListView(
+    mixins.ListModelMixin, mixins.CreateModelMixin, generics.GenericAPIView
+):
+    queryset = Incident.objects.all()
+
+    serializer_class = IncidentCreateSerializer
+    permission_classes = [UserHasProfile]
```
grahamhoyes, line 119: What about tech team creating incidents? Separate view for that?

hackathon_site/hardware/serializers.py
```diff
@@ -65,7 +65,12 @@ class Meta:
 
     @staticmethod
     def get_team_id(obj: Incident):
-        return obj.order_item.order.team.id
+        intermediary = (
+            obj["order_item"]["order"]
+            if isinstance(obj, OrderedDict)
```
grahamhoyes: Huh, what makes this happen? Our type annotation that this is an `Incident` object is incorrect it seems

hackathon_site/hardware/serializers.py
```diff
@@ -89,6 +94,20 @@ def get_team_code(obj: Order):
         return obj.team.team_code
 
 
+class IncidentCreateSerializer(IncidentListSerializer):
```
grahamhoyes: I think in this case, you don't want to subclass `IncidentListSerializer`, because we don't want the nested `order_item` serializer. Looking at the request body in django:

```
{
  "state": "Heavily Used",
  "time_occurred": "2021-11-19T00:56:12.694Z",
  "description": "string",
  "order_item": {
    "hardware": 0,
    "order": 0,
    "part_returned_health": "Healthy"
  }
}
```

We would have to specify the full `order_item` in the request (according to swagger anyway), which is unnecessary. It would be better if `order_item` were just the OrderItem id. 

This is contrary to my [comment](https://github.com/ieeeuoft/hackathon-template/pull/312/files#r752782211) on #312 where we can subclass, because the nested serializer field doesn't need to be part of the request.

hackathon_site/hardware/serializers.py
```diff
@@ -89,6 +94,20 @@ def get_team_code(obj: Order):
...
+    class Meta:
+        model = Incident
+        fields = IncidentListSerializer.Meta.fields
+        read_only_fields = ("team_id",)
+
+    def create(self, validated_data):
```
grahamhoyes: I think once you change this to not subclass the list serializer, the default `create` method will work

hackathon_site/hardware/test_api.py
```diff
@@ -607,6 +608,58 @@ def test_search_filter(self):
...
+        )
+
+        self.incident = Incident.objects.create(
+            state="Broken",
+            description="Description",
+            order_item=self.order_item,
+            time_occurred=datetime(2022, 8, 8, tzinfo=settings.TZ_INFO),
+        )
+
+        self.request_data = IncidentCreateSerializer(self.incident).data
```
grahamhoyes: I think it would actually be better to hard code the json we'd be submitted as part of the test rather than using the serializer to generate it, since that's how the frontend is going to use it after all. Plus then it would be easier to notice the nested field that is in here I think.

hackathon_site/event/api_views.py
grahamhoyes: Wow I am shockingly blind lately, it's in the hardware views 🤦 . Can you call this view `TeamIncidentListView` or something so they don't both have the same class name?

hackathon_site/hardware/test_api.py
arshinmar: yes will do

hackathon_site/event/api_views.py
arshinmar: fixed this

hackathon_site/event/api_views.py
arshinmar: yep sounds good

hackathon_site/hardware/serializers.py
arshinmar: sorry bout this

hackathon_site/hardware/test_api.py
arshinmar: yep will do

hackathon_site/hardware/serializers.py
arshinmar: yep

hackathon_site/hardware/serializers.py
arshinmar: yep

hackathon_site/event/test_api.py
```diff
@@ -613,3 +614,91 @@ def test_team_get_has_permissions(self):
...
+        self.order_item = OrderItem.objects.create(
+            order=self.order, hardware=self.hardware,
+        )
+
+        self.request_data = {
+            "state": "Broken",
+            "time_occurred": "2022-08-08T01:18:00-04:00",
+            "description": "Description",
+            "order_item": self.order_item.id,
+            "team_id": 1,
```
grahamhoyes: This shouldn't need to be in the request, it's a method field on the serializer and not actually going into the model.

If itt throws an error without it, can you add `team_id` to the read only fields for the serializer? Might as well add `created_at` and `update_at` to the read only fields too, since they are set automatically.

hackathon_site/hardware/serializers.py
```diff
@@ -89,6 +91,18 @@ def get_team_code(obj: Order):
         return obj.team.team_code
 
 
+class IncidentCreateSerializer(serializers.ModelSerializer):
```
grahamhoyes: `IncidentCreateSerializer` is defined twice?

I think you're fine with removing this one, the way you have it above with `IncidentListSerializer` subclassing `IncidentCreateSerializer` should work (and is a clever way of just adding the nested serializers for specific fields, we should keep that in mind).

hackathon_site/hardware/views.py
```diff
@@ -62,9 +60,32 @@ class IncidentListView(mixins.ListModelMixin, generics.GenericAPIView):
     filter_backends = (filters.DjangoFilterBackend, SearchFilter)
     filterset_class = IncidentFilter
 
+    def get_queryset(self):
+        queryset = Incident.objects.all()
+        if self.request.method == "GET":
+            return queryset.select_related(
```
grahamhoyes: You can just return this in all cases (and just keep it as the `queryset` property on the view, don't need the `get_queryset` method).

Since your serializer includes a method field for getting the team id, it would be efficient to at least include `order_item__order__team` in both cases. Technically the related hardware isn't required for the post case, so if you want to keep the `get_queryset` method you can `select_related` team accordingly in the POST block

hackathon_site/hardware/views.py
```diff
@@ -62,9 +60,32 @@ class IncidentListView(mixins.ListModelMixin, generics.GenericAPIView):
...
+
+    def get_serializer_class(self):
+        if self.request.method == "GET":
+            return IncidentListSerializer
+        elif self.request.method == "POST":
+            return IncidentCreateSerializer
+
+    def get_permissions(self):
+        if self.request.method == "POST":
+            return [FullDjangoModelPermissions()]
```
grahamhoyes: This should be the permissions class used no matter what.

As it is defined right now, anyone with an account could view all incidents.

arshinmar: this was the way it was before I started. I'm assuming there had to be a reason for why the author of the GET feature didn't have permissions for GET requests.

@Leo6Leo was there a specific reason why permissions weren't there for the GET feature?

YuYing-Liang: I think before we were sharing views between ppl with profiles and ppl with permissions but now we've switched to separate views for each so this view should have FullDjangoModelPermissions() for the GET as well since this is the view only for ppl with permissions

YuYing-Liang:
just a general question @arshinmar @grahamhoyes what's the difference/purpose of "state" and "part returned health"
image
They kind of seem the same to me

YuYing-Liang:
Another question, should making an incident report for an item change the quantity available for that hardware item (based on the part health)? Since I think incidents are for all returned items not just damaged ones

hackathon_site/hardware/views.py
grahamhoyes: Back when this view was originally implemented in #300, permissions were just left out to keep things simple. There's a comment saying they'll be added later [here](https://github.com/ieeeuoft/hackathon-template/pull/300#discussion_r715983131), looks like we never made a ticket for it though.

And as YuYing said, now it's pretty easy and we can just do `FullDjangoModelPermissions()` on all the admin-side views (you can just set `permission_classes = [FullDjangoModelPermissions]` on the view, no need for `get_permissions`).

Do you think you could really quickly add the basic checks for not logged in, logged in without permissions, and has permissions to `IncidentListsViewTestCase` (and maybe rename that to `IncidentListViewTestCase`? Should be a quick copy-paste.

grahamhoyes:
> just a general question @arshinmar @grahamhoyes what's the difference/purpose of "state" and "part returned health" image They kind of seem the same to me

State is specifically for when creating an incident, and it's declared by the participant. It's what the tech team will see when they get an incident report. Part returned health is declared by the tech team upon receiving it and marking it as returned.

Is it overkill to have the two fields? Probably.

grahamhoyes:
> Another question, should making an incident report for an item change the quantity available for that hardware item (based on the part health)? Since I think incidents are for all returned items not just damaged ones

No, because an incident is filed before the part is returned. But we should actually factor `part_returned_health` into the calculation for quantity remaining now that you mention it, if it's anything other that "Healthy" we probably shouldn't put it back into the pool. Made IEEE-146 for that.

hackathon_site/event/test_api.py
arshinmar: yep done

hackathon_site/hardware/views.py
arshinmar: ok the get_queryset method is removed.

hackathon_site/event/test_api.py
grahamhoyes: Still needs to be removed, not a required part of the request

hackathon_site/event/api_views.py
```diff
@@ -109,6 +110,26 @@ def post(self, request, *args, **kwargs):
...
+
+    def perform_create(self, serializer):
+        incident_team = serializer.validated_data["order_item"].order.team
+        user_team = self.request.user.profile.team
+
+        if incident_team != user_team:
+            raise PermissionDenied("Can only post incidents for your own team.")
```
grahamhoyes: Nit:
```suggestion
            raise PermissionDenied("Can only create incidents for your own team.")
```

Just making the message more user-friendly if we expose these messages somewhere (not that our frontend will let someone create an incident for another team)

hackathon_site/hardware/serializers.py
```diff
@@ -45,10 +45,28 @@ class Meta:
...
+            "description",
+            "order_item",
+            "created_at",
+            "updated_at",
+        )
+        read_only_fields = ("team_id", "created_at", "updated_at")
 
-    order_item = OrderItemSerializer()
+    @staticmethod
+    def get_team_id(obj: Incident):
```
grahamhoyes: This isn't being used by `IncidentCreateSerializer`. Likewise, having `team_id` in `read_only_fields` isn't doing anything.

In order to use this, you need `team_id = serializers.SerializerMethodField()` like is on the list serializer.

grahamhoyes: I think I commented wrong on my last review saying which of the create serializers to delete, the second one was actually the one to keep, sorry!. The definitions for both should be:

```python
class IncidentCreateSerializer(serializers.ModelSerializer):
    team_id = serializers.SerializerMethodField()

    class Meta:
        model = Incident
        fields = (
            "id",
            "state",
            "time_occurred",
            "description",
            "order_item",
            "team_id",
            "created_at",
            "updated_at",
        )
        read_only_fields = ("created_at", "updated_at")

    @staticmethod
    def get_team_id(obj: Incident) -> int:
        return obj.order_item.order.team.id


class IncidentListSerializer(IncidentCreateSerializer):
    order_item = OrderItemSerializer()
```

The only difference is that the list serializer uses a nested serializer for the order item. Everything else can stay the same.

grahamhoyes (approved):

PR #324, opened 2021-10-30
IEEE-39 extend the team serializer to include team members

Leo6Leo:
Please discard this pull request for now


hackathon_site/event/serializers.py
```diff
@@ -10,18 +10,15 @@ class Meta:
...
-            "team_code",
-            "created_at",
-            "updated_at",
-        )
+        model = User
+        fields = ("id", "first_name", "last_name", "email")
 
 
 class ProfileSerializer(serializers.ModelSerializer):
+    user = UserInSerializer(read_only=True)
```
grahamhoyes, line 20: Gotta be careful with this - the `UserSerializer` uses `ProfileSerializer`, so you'll end up with infinite recursion I think. You may need to define separate `ProfileInUserSerializer` and `ProfileInTeamSerializer`s

hackathon_site/event/serializers.py
```diff
@@ -41,3 +38,21 @@ class UserSerializer(serializers.ModelSerializer):
...
+        fields = (
+            "id",
+            "team_code",
+            "created_at",
+            "updated_at",
+            "profiles",
+        )
+
+    @staticmethod
+    def get_team_code(obj: Profile):
```
grahamhoyes, line 57: Not necessary, since `team_code` is an actual field on a team

grahamhoyes:
You'll need to fix this branch, since it looks like you branched off the branch you were using in #317 instead of off of develop, which makes following the changes difficult. Make sure that in the future, you always branch off of develop.

I'm not really sure how you ended up here. Looking at the commits in both, I think `f2c88c1` is the last commit that both branches share, but there are similar ones that were applied to both (like the "fix the pk type issue" commit, which shouldn't exist in this branch anymore).

There might be a way to fix this with a rebase, but I can't figure one out since there are some changes that were manually applied to both. I think you best bet is to:
- Rename this branch locally to something else with `git branch -M <new branch name>`
- Checkout and pull develop, then create a new branch with the same name as this one: `git checkout -b IEEE-39-Extend_the_Team_Serializer_to_include_team_members`
- Start cherry-picking the commits that you need, eg `git cherry-pick c438c1b`. Look at each commit in the list above on github manually to figure out which ones you need.
- Force push with `git push --force`

hackathon_site/event/serializers.py
Leo6Leo: yeah that's true. But to my surprise, it worked. I am afraid there will be some potential issues appear in the future if wrote like this

PR #325, opened 2021-10-31
IEEE-117 Add interfaces for common API types to api/types.ts

Mustaballer:
## Overview

- Resolves #IEEE-117
- Added interfaces for Profile, User, Team, and Order to api/types.ts

## Unit Tests Created

- None were created

## Steps to QA

- None

### DIFF
--- a/hackathon_site/dashboard/frontend/src/api/types.ts
+++ b/hackathon_site/dashboard/frontend/src/api/types.ts
@@ -61,7 +61,7 @@ export interface User {
     profile: Profile;
 }
 
-type UserWithoutProfile = Omit<User, "profile"| "team">;
+type UserWithoutProfile = Omit<User, "profile" | "team">;
 type ProfileWithoutTeamNumber = Omit<Profile, "team">
 
 
--- a/hackathon_site/dashboard/frontend/src/api/types.ts
+++ b/hackathon_site/dashboard/frontend/src/api/types.ts
@@ -42,3 +42,45 @@ export interface Category {
     max_per_team: number;
     unique_hardware_count: number;
 }
+
+/** Event API */
+export interface Profile {
+    id: number;
+    id_provided: boolean;
+    attended: boolean;
+    acknowledge_rules: boolean;
+    e_signature: string;
+    team: number;
+}
+
+interface UserWithoutProfile {
+    id: number;
+    first_name: string;
+    last_name: string;
+    email: string;
+}
+
+export interface User extends UserWithoutProfile {
+    profile: Profile;
+}
+
+export interface Team {
+    id: number;
+    team_code: string;
+    created_at: Date;
+    updated_at: Date;
+    profiles: Profile & { user: UserWithoutProfile }[];
+}
+
+/** Orders API **/
+export type OrderStatus = "Submitted" | "Ready for Pickup" | "Picked Up" | "Cancelled";
+
+export interface Order {
+    id: number;
+    hardware: Hardware[];
+    team: number;
+    team_code: string;
+    status: OrderStatus;
+    created_at: Date;
+    updated_at: Date;
+}
--- a/hackathon_site/dashboard/frontend/src/api/types.ts
+++ b/hackathon_site/dashboard/frontend/src/api/types.ts
@@ -53,23 +53,24 @@ export interface Profile {
     team: number;
 }
 
-interface UserWithoutProfile {
+export interface User {
     id: number;
     first_name: string;
     last_name: string;
     email: string;
-}
-
-export interface User extends UserWithoutProfile {
     profile: Profile;
 }
+
+type UserWithoutProfile = Omit<User, "profile"| "team">;
+type ProfileWithoutTeamNumber = Omit<Profile, "team">
+
 
 export interface Team {
     id: number;
     team_code: string;
-    created_at: Date;
-    updated_at: Date;
-    profiles: Profile & { user: UserWithoutProfile }[];
+    created_at: string;
+    updated_at: string;
+    profiles: ({ profile: ProfileWithoutTeamNumber } & { user: UserWithoutProfile })[];
 }
 
 /** Orders API **/
@@ -81,6 +82,6 @@ export interface Order {
     team: number;
     team_code: string;
     status: OrderStatus;
-    created_at: Date;
-    updated_at: Date;
+    created_at: string;
+    updated_at: string;
 }
--- a/hackathon_site/dashboard/frontend/src/api/types.ts
+++ b/hackathon_site/dashboard/frontend/src/api/types.ts
@@ -62,8 +62,7 @@ export interface User {
 }
 
 type UserWithoutProfile = Omit<User, "profile" | "team">;
-type ProfileWithoutTeamNumber = Omit<Profile, "team">
-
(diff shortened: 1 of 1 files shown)


hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -42,3 +42,45 @@ export interface Category {
     max_per_team: number;
     unique_hardware_count: number;
 }
+
+/** Event API */
```
Leo6Leo, line 46: just to be consistent. add one more star lolllll.

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -42,3 +42,45 @@ export interface Category {
...
+    team_code: string;
+    created_at: Date;
+    updated_at: Date;
+    profiles: Profile & { user: UserWithoutProfile }[];
+}
+
+/** Orders API **/
+export type OrderStatus = "Submitted" | "Ready for Pickup" | "Picked Up" | "Cancelled";
+
+export interface Order {
```
Leo6Leo, line 78: just asking: all interfaces here are used to tell react what type of each variable is right? Does it like a header file in C/C++ ? What is the purpose of these statements?

hackathon_site/dashboard/frontend/src/api/types.ts
grahamhoyes: All of the ones above only have one closing star, so this is fine. No idea why, but it's normal to start a block comment like this with a double star :man_shrugging:.

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -42,3 +42,45 @@ export interface Category {
...
+}
+
+export interface User extends UserWithoutProfile {
+    profile: Profile;
+}
+
+export interface Team {
+    id: number;
+    team_code: string;
+    created_at: Date;
```
grahamhoyes: Unfortunately when this comes back from the API, it will just be a datetime string in ISO 8601 format so the type should be `string`. Same for `updated_at` below.

To make matters worse, Redux technically can't store date values either (it will sometimes work, but since they aren't serializable, it's not guaranteed or recommended). 

We probably won't end up actually using this anywhere on the frontend, so it doesn't really matter. If we did, we'll just have to convert to a `Date` as part of the component for rendering purposes.

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -42,3 +42,45 @@ export interface Category {
...
+    updated_at: Date;
+    profiles: Profile & { user: UserWithoutProfile }[];
```
grahamhoyes: Great use of typescript magic! Two things:

First, this needs an extra set of brackets to work properly `(Profile & { user: UserWithoutProfile })[]`

Second, from looking at #324, the nested `profile` don't have a `team` property which this currently expects. You can use the [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) utility type to remove it from `Profile`, but that might start to get messy.

hackathon_site/dashboard/frontend/src/api/types.ts
grahamhoyes: `interface`s (and `type`s) are the core piece of Typescript. They are type annotiations that tell you what type a variable is (which in the case of javascript, usually means what "shape" an object is).

You would use it like

```typescript
const myOrder: Order = {
  id: 5,
  hardware: [ /* Some hardware */ ],
  team: 2,
  team_code: "ABCDE",
  status: "Submitted",
  created_at: now,
  updated_at: now 
} 
```

If you didn't build the `myOrder` object properly (say you tried to put a random string in the `OrderStatus` field), then the typescript compiler (and your IDE) would throw an error. This **only** happens at compile time - typescript code ends up just getting compiled to javascript, where all of the type checking doesn't exist.

Headers in c/c++ are a bit more complicated, and are used by the compiler to link together references to different routines in different programs you compile together. Typescript is much easier to understand!

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -42,3 +42,45 @@ export interface Category {
...
+/** Orders API **/
+export type OrderStatus = "Submitted" | "Ready for Pickup" | "Picked Up" | "Cancelled";
+
+export interface Order {
+    id: number;
+    hardware: Hardware[];
+    team: number;
+    team_code: string;
+    status: OrderStatus;
+    created_at: Date;
```
grahamhoyes: Ditto about needing to be `string`

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -42,3 +42,45 @@ export interface Category {
...
+}
+
+interface UserWithoutProfile {
+    id: number;
+    first_name: string;
+    last_name: string;
+    email: string;
+}
+
+export interface User extends UserWithoutProfile {
```
grahamhoyes: As I mentioned above, it might be a little cleaner to write this the other way around and do

```typescript
export interface User {
    id: number;
    first_name: string;
    last_name: string;
    email: string;
    profile: Profile;
}

type UserWithoutProfile = Omit<User, "profile">;
```

grahamhoyes (commented):
Great start!

hackathon_site/dashboard/frontend/src/api/types.ts
Mustaballer: Okay so what I did was omit the team property from Profile, and from the UserWithoutProfile. The changes are in the latest commit, I'm not too sure how to clean it up or if it works.

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -42,3 +42,45 @@ export interface Category {
...
+
+export interface User {
+    id: number;
+    first_name: string;
+    last_name: string;
+    email: string;
+    profile: Profile;
+}
+
+type UserWithoutProfile = Omit<User, "profile" | "team">;
```
grahamhoyes: User doesn't have a team property, can that be removed from the omit list?

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -42,3 +42,45 @@ export interface Category {
...
+
+type UserWithoutProfile = Omit<User, "profile">;
+type ProfileWithoutTeamNumber = Omit<Profile, "team">;
+
+export interface Team {
+    id: number;
+    team_code: string;
+    created_at: string;
+    updated_at: string;
+    profiles: ({ profile: ProfileWithoutTeamNumber } & { user: UserWithoutProfile })[];
```
YuYing-Liang: I think this will create
`profiles: [ { profile: ProfileWithoutTeam, user: UserWithoutProfile}]`
but we want
`profiles: [ { ... <properties of ProfileWithoutTeam>, user: UserWithoutProfile}]`
so maybe try:
```suggestion
    profiles: (ProfileWithoutTeamNumber & { user: UserWithoutProfile })[];
```

YuYing-Liang:
Looks good! just need to quickly merge with develop, let us know if you need any help with that :)

YuYing-Liang (approved):

PR #326, opened 2021-11-08
Ieee 95 add hardware product info to inventory

YuYing-Liang:
## Overview

- Resolves #IEEE-95
- converted ProductOverview to typescript
- added notes property in hardware type
- converted uiSlice to typescript
- added inventory object to uiSlice which included ProductOverview open state
- connected store to inventory grid

## Unit Tests Created

- None, I should write some though


hackathon_site/dashboard/frontend/src/components/inventory/InventoryGrid/InventoryGrid.tsx
```diff
@@ -2,15 +2,21 @@ import React from "react";
 import Grid from "@material-ui/core/Grid";
 import styles from "pages/Inventory/Inventory.module.scss";
 import Item from "components/inventory/Item/Item";
-
-import { useSelector } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import { hardwareSelectors, isLoadingSelector } from "slices/hardware/hardwareSlice";
 import { LinearProgress } from "@material-ui/core";
+import { setProductOverviewItem } from "../../../slices/ui/uiSlice";
```
grahamhoyes: Switch these to absolute imports.

If pycharm made these imports for you, you can make it use absolute paths from Settings > Editor > Code Style > Javascript > Imports tab > and check "Use paths relative to the project, resource or sources root".

hackathon_site/dashboard/frontend/src/components/inventory/InventoryGrid/InventoryGrid.tsx
```diff
@@ -2,15 +2,21 @@ import React from "react";
...
+import { Hardware } from "../../../api/types";
+import { RootState } from "../../../slices/store";
```
grahamhoyes: Unused

hackathon_site/dashboard/frontend/src/slices/ui/uiSlice.ts
```diff
@@ -0,0 +1,129 @@
+import {createSelector, createSlice, PayloadAction} from "@reduxjs/toolkit";
+import {Hardware} from "../../api/types";
```
grahamhoyes: Ditto with the absolute imports

hackathon_site/dashboard/frontend/src/slices/ui/uiSlice.ts
```diff
@@ -0,0 +1,129 @@
...
+        hardwareItemBeingViewed: Hardware | null,
+        isProductOverviewVisible: boolean,
+    },
+    snackbars: ({
+        message: string,
+        options: {
+            key: number
+        },
+        dismissed: boolean,
+    })[] | []
```
grahamhoyes: Don't think you need the ` | []` part? Any array type allows for that array to be empty.

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -37,25 +44,23 @@ const createQuantityList = (number) => {
...
-    constraintMax,
+    maxPerTeam,
     handleSubmit,
     handleChange,
-    requestFailure,
     values: { quantity },
-}) => {
-    const dropdownNum = !constraintMax
+}: AddToCartFormProps) => {
+    const dropdownNum = !maxPerTeam
```
grahamhoyes: Something I just noticed is that `max_per_team` for hardware and category are both nullable, so those type definitions should be updated.

Then after that, what should this do if we set the max per team to 0 for some reason? This count should probably stay as zero. Once max per team is nullable in the type definition, can this be changed to `const dropdownNum = maxPerTeam === null ? quantityAvailable : Math.min(quantityAvailable, maxPerTeam)`

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -184,56 +212,89 @@ const MainSection = ({ name, total, quantityAvailable, categories, img }) => {
...
-                    quantityAvailable={detail.quantityAvailable}
-                    constraintMax={detail.constraintMax}
-                />
+    const isProductOverviewVisible: boolean = useSelector(
+        isProductOverviewVisibleSelector
+    );
+    const hardware: Hardware | null = useSelector(hardwareBeingViewedSelector);
+    const allCategories = useSelector(categorySelectors.selectEntities);
+
+    const categories: (Category | undefined)[] = hardware
```
grahamhoyes: Think you could write a `selectCategoriesByIds`  selector similar to the hardware one [here](https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/dashboard/frontend/src/slices/hardware/hardwareSlice.ts#L167)? It would get you the same information, but it would be memoized so slightly more efficient.

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -184,56 +212,89 @@ const MainSection = ({ name, total, quantityAvailable, categories, img }) => {
...
+        ? hardware.categories.map((categoryId) => allCategories[categoryId])
+        : [];
+
+    const constraints: (string | undefined)[] =
+        categories.length > 0
+            ? [`Max ${hardware?.max_per_team} of this item`].concat(
```
grahamhoyes: If `hardware.max_per_team` is null (as mentioned above), won't want to have this message

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -184,56 +212,89 @@ const MainSection = ({ name, total, quantityAvailable, categories, img }) => {
...
+
+    const closeProductOverview = () => dispatch(removeProductOverviewItem());
+
+    return (
+        <SideSheetRight
+            title="Product Overview"
+            isVisible={isProductOverviewVisible}
+            handleClose={closeProductOverview}
+        >
+            {hardware && (
```
grahamhoyes: What else should we do if hardware is null? It would be rather strange to just show an empty drawer.

Should we attempt to fetch that hardware? You can use a `useEffect` hook and if hardware is null and hardware is not loading, dispatch actions to clear the hardware filters, set the filters to the ID we want, then dispatch `getHardwareWithFilters({ keepOld: true })`. And render a loading wheel while that's loading or something.

What about if somehow this gets launched somehow with a hardware that doesn't actually exist? I think other parts of the site (like the cart page) will just keep making infinite requests, so that's fine.

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -184,56 +212,89 @@ const MainSection = ({ name, total, quantityAvailable, categories, img }) => {
...
+        isProductOverviewVisibleSelector
+    );
+    const hardware: Hardware | null = useSelector(hardwareBeingViewedSelector);
+    const allCategories = useSelector(categorySelectors.selectEntities);
+
+    const categories: (Category | undefined)[] = hardware
+        ? hardware.categories.map((categoryId) => allCategories[categoryId])
+        : [];
+
+    const constraints: (string | undefined)[] =
```
grahamhoyes: I think it would be best to remove any undefined items before constructing this array, since right now you'll end up with "Max undefined of items under category undefined` sometimes

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -165,13 +180,26 @@ const DetailInfoSection = ({
     );
 };
 
-const MainSection = ({ name, total, quantityAvailable, categories, img }) => {
+interface MainSectionProps {
+    name: string;
+    quantityAvailable: number;
+    quantityRemaining: number;
+    categories: (string | undefined)[];
```
grahamhoyes: This should just be a `string[]`, make the component that passes it in filter out the undefined values

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -110,40 +118,47 @@ export const EnhancedAddToCartForm = ({
...
         </Formik>
     );
 };
 
+interface DetailInfoSectionProps {
+    manufacturer: string;
+    modelNumber: string;
+    datasheet: string;
+    notes: string;
+    constraints: (string | undefined)[];
```
grahamhoyes: Same, just make this a `string[]` and let the outer component filter

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -89,14 +94,17 @@ export const AddToCartForm = ({
...
     quantityAvailable,
-    constraintMax,
-}) => {
-    const onSubmit = (formikValues) => {
-        handleSubmit(formikValues.quantity);
+    maxPerTeam,
+}: EnhancedAddToCartFormProps) => {
+    const onSubmit = (formikValues: { quantity: string }) => {
+        // TODO: add an item to cart store
+        alert(`Ordering ${formikValues.quantity} of this item`);
```
grahamhoyes: #328!

grahamhoyes (commented):
Some initial thoughts, good work overall though!

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
YuYing-Liang: I don't see how this modal would be rendered with hardware that doesn't exist because it's accessed from the inventory/dashboard page (I guess tech team could accidentally delete it). Fetching more hardware makes sense but I don't see the case where that would be needed

YuYing-Liang: also if hardware is null, how do we attempt to fetch hardware? we don't know what hardware we want

grahamhoyes: Oh good point, this isn't like the cart cards where we have a hardware ID to fetch. 

As long as we make sure that the hardware is in the store before dispatching an action to load it, we should be fine. The edge case I was considering is that we only get hardware IDs back from the order API, so we just need to make sure all of those hardware are loaded before rendering anywhere (which is probably required anyway).

YuYing-Liang:
I'll create a new ticket to fix all the nullable fields on the frontend

hackathon_site/dashboard/frontend/src/testing/mockData.tsx
```diff
@@ -133,19 +133,18 @@ export const cartQuantity = 5;
 export const userEmail = "graham@email.com";
 
 // Product Overview
-export const productInformation = {
+export const productInformation: Hardware = {
```
grahamhoyes: Could technically just use `mockHardware[0]` in place of this, that might be best given the rest of our mock data is designed to work with it.

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -35,6 +36,12 @@ export interface HardwareFilters {
     ordering?: HardwareOrdering;
 }
 
+/** Hardware Product Overview Types */
+export interface ProductOverviewItem extends Omit<Hardware, "categories"> {
```
grahamhoyes: This is unused, did you have plans for it?

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,337 @@
...
+    const isProductOverviewVisible: boolean = useSelector(
+        isProductOverviewVisibleSelector
+    );
+    const hardware = useSelector(hardwareBeingViewedSelector);
+    const categories = useSelector((state: RootState) =>
+        selectCategoriesByIds(state, hardware?.categories || [])
+    );
+    if (categories.length > 0) {
+        categoryNames = categories.map(
+            (category) => category?.name ?? `Category ${category?.id}`
```
grahamhoyes: This can still technically end up being `Category undefined` if something breaks somewhere.

Could you do `categories.filter(category => !!category).map(category => category.name)`?

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,337 @@
...
+                            quantityAvailable={hardware.quantity_available}
+                            hardwareId={hardware.id}
+                            name={hardware.name}
+                            maxPerTeam={maxPerTeam}
+                        />
+                    )}
+                </div>
+            ) : (
+                <Typography variant="subtitle2" align="center" paragraph>
+                    Unable to display hardware. Please refresh page and try again.
```
grahamhoyes: Super minor:

```suggestion
                    Unable to display hardware. Please refresh the page and try again.
```

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,337 @@
...
+    maxPerTeam,
+}: EnhancedAddToCartFormProps) => {
+    const dispatch = useDispatch();
+    const currentQuantityInCart =
+        useSelector((state: RootState) => cartSelectors.selectById(state, hardwareId))
+            ?.quantity ?? 0;
+
+    const onSubmit = (formikValues: { quantity: string }) => {
+        const numQuantity: number = parseInt(formikValues.quantity);
+        if (currentQuantityInCart + numQuantity <= (maxPerTeam || quantityAvailable)) {
```
grahamhoyes: Since you default `maxPerTeam` to `Infinity`, I don't think there's any situation where it's undefined, meaning this check would always be true. Also what about maxPerTeam of 0? (No idea why we'd do that, but should handle the constraint still).

You might want to find another way of handling `maxPerTeam` in the outer component other than using `Infinity`, since it gets hard to reason about what's going on.

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,337 @@
...
+            dispatch(addToCart({ hardware_id: hardwareId, quantity: numQuantity }));
+            dispatch(
+                displaySnackbar({
+                    message: `Added ${numQuantity} ${name} item(s) to your cart.`,
```
grahamhoyes, line 124: Could we make this a `success` variant? I was so excited to see that adding to cart worked that I wanted it to green :)

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,337 @@
...
+                })
+            );
+        } else {
+            dispatch(
+                displaySnackbar({
+                    message: `You've already added the maximum amount of this item to your cart.`,
```
grahamhoyes: And this should probably be a warning variant

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+            .reduce((prev, curr) => Math.min(prev, curr));
+
+        const { getByText, getByRole, queryByText } = render(
+            <ProductOverview showAddToCartButton={true} />,
+            {
+                store,
+            }
+        );
+
+        fireEvent.mouseDown(getByRole("button", { name: "Qty 1" }));
```
grahamhoyes, line 55: Are you sure this is doing anything? What component has a name of `Qty 1`?

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+            store,
+        });
+
+        const button = getByText("Add to cart");
+
+        await waitFor(() => {
+            fireEvent.click(button);
+        });
+
+        expect(cartSelectors.selectAll(store.getState())).toEqual([initialCart]);
```
grahamhoyes: I could never get snackbars to render during tests properly, but if you wanted to check for the error message you could look in the snackbars array in the UI slice perhaps?

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
grahamhoyes: Ah, I see that it was a leftover from the original .js tests. I still don't think it's doing anything though

grahamhoyes (commented):
I had a huge smile on my face while clicking stuff and adding it to the cart, we're getting somewhere! Great job!

grahamhoyes:
Random UI observation: We need to change the mouse to a hover pointer or whatever it is when hovering over the inventory items

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
YuYing-Liang: yeah for sure, i thought maybe adding an extra filter() was too much but ig there's no other way

YuYing-Liang: Even with the filter, typescript is saying category could be undefined :(

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
YuYing-Liang: will do!💚

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
YuYing-Liang: I think I figured it out! :))

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
YuYing-Liang: will remove

YuYing-Liang: hmm actually I think this is the dropdown role

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
YuYing-Liang: I still use infinity but I initialize maxPerTeam to null

hackathon_site/dashboard/frontend/src/App.tsx
```diff
@@ -82,7 +82,7 @@ const ConnectedApp = () => (
                 <SnackbarProvider>
                     <SnackbarNotifier />
                     <ConnectedRouter history={history}>
-                        <UnconnectedApp />
+                        <UnconnectedApp />i
```
grahamhoyes: 👁️

hackathon_site/dashboard/frontend/src/components/general/SnackbarNotifier/SnackbarNotifier.js
```diff
@@ -27,42 +27,44 @@ export const UnconnectedSnackbarNotifier = ({
             setDisplayedSnackbars(displayedSnackbars);
         };
 
-        snackbars.forEach(({ message, options, dismissed }) => {
-            if (dismissed) {
-                closeSnackbar(options.key);
-                return;
-            }
+        if (snackbars?.length) {
```
grahamhoyes, line 30: Was there a bug where this was undefined somehow?

hackathon_site/dashboard/frontend/src/pages/Cart/Cart.tsx
```diff
@@ -87,18 +87,6 @@ const Cart = () => {
                             );
                         })
                     )}
-                    {/*TODO: Remove this button when something else has been hooked up to the store*/}
-                    <Button
```
grahamhoyes, line 91: Can remove the `Button` and `addToCart` imports

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.js
```diff
@@ -47,12 +46,8 @@ const Dashboard = () => {
     return (
         <>
             <Header />
-            <ProductOverview
-                detail={productInformation}
-                addToCart={addToCart}
```
grahamhoyes, line 52: Can remove the `const addToCart` placeholder function above

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
grahamhoyes:
```typescript
categoryNames = categories.filter((category): category is Category => !!category)
                                   .map(category => category.name);
```

Type guards to the rescue!

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,356 @@
...
+    );
+    const hardware = useSelector(hardwareBeingViewedSelector);
+    const categories = useSelector((state: RootState) =>
+        selectCategoriesByIds(state, hardware?.categories || [])
+    );
+    if (categories.length > 0) {
+        categoryNames = categories.map(
+            (category) => category?.name ?? `Category ${category?.id}`
+        );
+        categories.filter((category) => !!category).map((category) => category?.name);
```
grahamhoyes: Can remove this once the filter/map above works

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,356 @@
...
+            : [];
+        maxPerTeam = hardware?.max_per_team ?? Infinity;
+        for (const category of categories) {
+            if (category?.max_per_team) {
+                constraints.push(
+                    `Max ${category.max_per_team} of items under category ${category.name}`
+                );
+                maxPerTeam = Math.min(category.max_per_team, maxPerTeam);
+            }
+        }
```
grahamhoyes: Change the declaration of `maxPerTeam` to `let maxPerTeam: number | undefined;`. Then:

```suggestion
        maxPerTeam = hardware?.max_per_team;
        for (const category of categories) {
            if (category?.max_per_team) {
                constraints.push(
                    `Max ${category.max_per_team} of items under category ${category.name}`
                );
                maxPerTeam =
                    maxPerTeam === undefined
                        ? category.max_per_team
                        : Math.min(category.max_per_team, maxPerTeam);
            }
        }
```

The use of Infinity otherwise concerns me

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,356 @@
...
+            </form>
+        </>
+    );
+};
+
+interface EnhancedAddToCartFormProps {
+    quantityAvailable: number;
+    hardwareId: number;
+    name: string;
+    maxPerTeam: number | null;
```
grahamhoyes, line 105: Change to `maxPerTeam: number | undefined` as well

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
grahamhoyes:
```suggestion
        if (currentQuantityInCart + numQuantity <= (maxPerTeam ?? quantityAvailable)) {
```

No more problem with `Infinity`, and we can handle a weird constraint of max 0 per team

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,356 @@
...
+            </MenuItem>
+        );
+    }
+
+    return entry;
+};
+
+interface AddToCartFormProps extends FormikValues {
+    quantityAvailable: number;
+    maxPerTeam: number | null;
```
grahamhoyes, line 52:
```suggestion
    maxPerTeam: number | undefined;
```

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,356 @@
...
+}
+export const AddToCartForm = ({
+    quantityAvailable,
+    maxPerTeam,
+    handleSubmit,
+    handleChange,
+    values: { quantity },
+}: AddToCartFormProps) => {
+    const dropdownNum =
+        maxPerTeam !== null
```
grahamhoyes, line 62:
```suggestion
        maxPerTeam !== undefined
```

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
grahamhoyes: You can keep it as `null` if you want instead of `undefined`, just initialize it as such and ignore the rest of my comments on the matter.

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
grahamhoyes: Seems you're right because the tests break if it's removed, but I don't see anything with that role unless mui is doing it under the hood? @lisa-sa-li do you know anything about this?

hackathon_site/dashboard/frontend/src/App.tsx
YuYing-Liang: oops

hackathon_site/dashboard/frontend/src/components/general/SnackbarNotifier/SnackbarNotifier.js
YuYing-Liang: yes, in unit tests

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,358 @@
...
+    );
+    const hardware = useSelector(hardwareBeingViewedSelector);
+    const categories = useSelector((state: RootState) =>
+        selectCategoriesByIds(state, hardware?.categories || [])
+    );
+    if (categories.length > 0) {
+        categoryNames = categories
+            .filter((category): category is Category => !!category)
+            .map((category) => category.name);
+        constraints = hardware?.max_per_team
```
grahamhoyes:
```suggestion
        constraints = hardware?.max_per_team !== undefined
```

Minor change, but it allows us to set max per team to 0 in the event we want to stop allowing an item to be checked out for some reason. No idea why we'd do that, but now we can!

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -0,0 +1,358 @@
...
+            ? [`Max ${hardware.max_per_team} of this item`]
+            : [];
+        maxPerTeam = hardware?.max_per_team ?? null;
+        for (const category of categories) {
+            if (category?.max_per_team) {
```
grahamhoyes:
```suggestion
            if (category?.max_per_team !== undefined) {
```

hackathon_site/event/admin.py
```diff
@@ -1,6 +1,12 @@
 from django.contrib import admin
```
grahamhoyes: How'd this sneak in here :eyes:

grahamhoyes (approved):

PR #327, opened 2021-11-13
IEEE-39 extend the team serializer to include team members

Leo6Leo:
## Overview

TeamSerializer is used for the current team view, and will be used for list views on the tech team side. Currently, it doesn’t contain the team members, but it should (with their first name, last name, and email at least).

I would first look at removing the team serializer from the profile serializer if it’s still there (but leave it in the fields list, so it stays as just an ID). Then ProfileSerializercan be included in the team serializer. Add user fields to the profile serializer (by creating a UserInProfile serializer, since we can’t use the current UserSerializer). Update tests.

If that breaks too many tests, or seems like a bad idea (I can’t think of anywhere we’d need to get the full team from a profiles API endpoint, other than for the current user which we can get from the current team endpoint), then create a new ProfileInTeam serializer that has everything but the team in

## Unit Tests Created

Test whether the output of the new team serializer whether matches the expected result or not.

### DIFF
--- a/hackathon_site/event/serializers.py
+++ b/hackathon_site/event/serializers.py
@@ -16,7 +16,22 @@ class UserInSerializer(serializers.ModelSerializer):
         fields = ("id", "first_name", "last_name", "email")
 
 
-class ProfileSerializer(serializers.ModelSerializer):
+class ProfileInUserSerializer(serializers.ModelSerializer):
+    user = UserInSerializer(read_only=True)
+
+    class Meta:
+        model = Profile
+        fields = (
+            "id",
+            "id_provided",
+            "attended",
+            "acknowledge_rules",
+            "e_signature",
+            "user",
+        )
+
+
+class ProfileInTeamSerializer(serializers.ModelSerializer):
     user = UserInSerializer(read_only=True)
 
     class Meta:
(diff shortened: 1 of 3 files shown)


hackathon_site/event/serializers.py
```diff
@@ -10,18 +10,28 @@ class Meta:
         fields = ("id", "name")
 
 
-class TeamSerializer(serializers.ModelSerializer):
+class UserInSerializer(serializers.ModelSerializer):
```
grahamhoyes: `UserInProfileSerializer`

hackathon_site/event/tests.py
```diff
@@ -1,6 +1,8 @@
 import re
 from unittest.mock import patch
 from datetime import datetime, timedelta
+import json
```
grahamhoyes: Not used anymore

hackathon_site/event/tests.py
```diff
@@ -17,6 +19,9 @@
     UserSerializer,
     GroupSerializer,
     ProfileSerializer,
+    ProfileInUserSerializer,
+    ProfileInTeamSerializer,
+    UserInSerializer,
```
grahamhoyes: Still need a test for this one! (And to rename it to `UserInProfileSerializer`)

grahamhoyes (commented):
Great work, super close!

Serializers look good, just need to rename one of them and add one more test

grahamhoyes (approved):

PR #330, opened 2021-11-17
Ieee 110 remove from cart action

Mustaballer:
## Overview

- Resolves IEEE-110
- Created removeFromCart action
- Updated the “X” button in CartCard to dispatch this action.

## Unit Tests Created

- tested removeFromCart action by adding an item to the cart and then removing it, and checked if length of the store's state was zero
- added multiple items in cart and then dispatched removeFromCart and checked if the correct item was removed and what remained

## Steps to QA

- Go to cart page, once you click "add item to Cart", you can then click the "X" button to remove it from the cart.

### DIFF
--- a/hackathon_site/dashboard/frontend/src/api/types.ts
+++ b/hackathon_site/dashboard/frontend/src/api/types.ts
@@ -84,3 +84,9 @@ export interface Order {
     created_at: string;
     updated_at: string;
 }
+
+/** Cart API */
+export interface CartItem {
+    hardware_id: number;
+    quantity: number;
+}
--- a/hackathon_site/dashboard/frontend/src/pages/Cart/Cart.test.tsx
+++ b/hackathon_site/dashboard/frontend/src/pages/Cart/Cart.test.tsx
@@ -12,15 +12,17 @@ import { get } from "api/api";
 
 import Cart from "pages/Cart/Cart";
 import { Hardware } from "api/types";
+import { addToCart } from "slices/hardware/cartSlice";
 
 jest.mock("api/api");
 const mockedGet = get as jest.MockedFunction<typeof get>;
 
 describe("Cart Page", () => {
-    test("Cart items and Cart Summary card appears", async () => {
-        // TODO: Cart page currently loads all of mockCartItems. Update this test once
-        // the page pulls items from redux.
+    test("Cart items from store and Cart Summary card appears", async () => {
         const store = makeStoreWithEntities({ hardware: mockHardware });
+        mockCartItems.forEach((item) => {
+            store.dispatch(addToCart(item));
+        });
 
         const { getByText } = render(<Cart />, { store });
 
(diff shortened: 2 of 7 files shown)


YuYing-Liang:
@Mustaballer I think you branched off from your addToCart branch, could you try syncing with develop?

hackathon_site/dashboard/frontend/src/components/cart/CartCard/CartCard.test.tsx
```diff
@@ -3,11 +3,12 @@ import CartCard from "components/cart/CartCard/CartCard";
 
 import { render, makeStoreWithEntities, waitFor } from "testing/utils";
 import { mockHardware } from "testing/mockData";
+import { addToCart, cartSelectors } from "slices/hardware/cartSlice";
+import { fireEvent, getByRole } from "@testing-library/react";
```
YuYing-Liang: minor: can you import `fireEvent `and `getByRole `from `testing/utils` (on line 4)

hackathon_site/dashboard/frontend/src/components/cart/CartCard/CartCard.test.tsx
```diff
@@ -64,4 +65,42 @@ describe("<CartCard />", () => {
             expect(getByText(error)).toBeInTheDocument();
         });
     });
+
+    test("removeFromCart button", async () => {
```
YuYing-Liang: this is a good test!
However, you're testing a UI component but you are making assertions on the cart store.
When we test react components, we are looking for noticeable changes in the UI(e.g. is a text being shown, is a new ui element being displayed/hidden) so in your case, when we remove a cart item, we want to check that it's actually being removed from the page. 
Also, this kind of test is probably better in Cart.test.tsx because your testing to see if the cart card of the item you want to remove is not there anymore

hackathon_site/dashboard/frontend/src/components/cart/CartCard/CartCard.test.tsx
```diff
@@ -64,4 +65,42 @@ describe("<CartCard />", () => {
...
+        const store = makeStoreWithEntities({ hardware: mockHardware });
+
+        store.dispatch(
```
YuYing-Liang: It's good you are using the reducer you made! But it's not needed for this test since you're trying to test the remove from cart functionality. What you can do instead is populate the store with initial cart items, you'll have to modify the makeStoreWithEntities function (in utils.tsx line 100) to accept cart items.

This is not required for this ticket, but it might help having a pre-populated store for future tickets, reach out if you want help doing this!

YuYing-Liang (commented):
Nice start!

grahamhoyes:
> @Mustaballer I think you branched off from your addToCart branch, could you try syncing with develop?

Ideall, do a `git rebase --onto` and force push. Look for `--onto` here for instructions:  https://git-scm.com/docs/git-rebase

hackathon_site/dashboard/frontend/src/components/cart/CartCard/CartCard.test.tsx
grahamhoyes: An example of how you might check that it's removed from the page: `expect(queryByText(mockHardware[0].name)).not.toBeInTheDocument()`, assuming that hardware name is something that gets rendered as part of the cart card.

PR #332, opened 2021-11-20
IEEE-116 json order model

Leo6Leo:
## Overview

Add a request = models.JSONField(null=False) to the Order model. 

Update the Order create view to save validated_data into this field (this will be inside OrderCreateSerializer.create method, nothing needs changing in the actual view)

Update the Order serializer to return this field.

This will be used by the frontend so we can distinguish quantity requested and quantity granted for each order.

## Unit Tests Created

- Under Progress


Leo6Leo:
That's quite weird. After I changed the serializer, 47 tests failed. 
image

hackathon_site/hardware/migrations/0007_order_request.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ("hardware", "0006_auto_20210918_1440"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="order",
+            name="request",
+            field=models.JSONField(default=123),
```
grahamhoyes:
```suggestion
            field=models.JSONField(default={hardware: []}),
```

Leo6Leo:
1. need to change the previous test cases for Order seriliazer because we add the new field: "request" to that
2. modify all the old tests whose initialization of Order object doesn't have field ”request“

hackathon_site/hardware/serializers.py
```diff
@@ -1,9 +1,12 @@
+import json
```
grahamhoyes: Unused import

hackathon_site/hardware/serializers.py
```diff
@@ -1,9 +1,12 @@
...
 from collections import Counter
 import functools
 from django.db.models import Count, Q
 from django.core.exceptions import ObjectDoesNotExist
 from rest_framework import serializers
 
+from collections import OrderedDict
```
grahamhoyes: Unused

hackathon_site/hardware/serializers.py
```diff
@@ -204,6 +208,16 @@ def create(self, validated_data):
...
         new_order = None
         response_data = {"order_id": None, "hardware": [], "errors": []}
+
+        # The reason why doing this is because the id field stores the hardware object, django cannot translate hardware object into JSON. Therefore, loop has been used to call Profile seriliazer and assign the JSON file to the validated_data filed.
+        counter = 0
+        for item in validated_data["hardware"]:
+
+            temp = HardwareSerializer(item["id"]).data
+
+            validated_data["hardware"][counter] = temp
```
grahamhoyes: Overwriting `validated_data` isn't a great idea, in case we try to use it for another purpose later on.

Also, the `request` field should also be storing the actual request that was made, it doesn't need to contain a fully serialized hardware instance - in other words, it should contain the data from `OrderCreateSerializer`, which you can get from `self.data`

hackathon_site/hardware/serializers.py
```diff
@@ -220,7 +234,9 @@ def create(self, validated_data):
                 continue
             if new_order is None:
                 new_order = Order.objects.create(
-                    team=self.context["request"].user.profile.team, status="Submitted"
+                    team=self.context["request"].user.profile.team,
+                    status="Submitted",
+                    request=validated_data,
```
grahamhoyes: Per above:
```suggestion
                    request=self.data,
```

hackathon_site/hardware/serializers.py
```diff
@@ -204,6 +208,16 @@ def create(self, validated_data):
         )
         new_order = None
         response_data = {"order_id": None, "hardware": [], "errors": []}
+
+        # The reason why doing this is because the id field stores the hardware object, django cannot translate hardware object into JSON. Therefore, loop has been used to call Profile seriliazer and assign the JSON file to the validated_data filed.
+        counter = 0
+        for item in validated_data["hardware"]:
```
grahamhoyes: This block is unnecessary since you can just use `self.data`, but for future reference a cleaner way to write it would have been:

```python
for i, item in enumerate(validated_data["hardware"]):
    validated_data["hardware"][i] = HardwareSerializer(item["id"]).data
```

hackathon_site/hardware/serializers.py
grahamhoyes: You'll also need to add tests to `OrderListViewPostTestCase` to make sure the request field is being built properly, since there aren't any direct serializer test for this - that would have caught that the request field is being built wrong (submit an order request and look at the order in the admin site to confirm this).

It looks like in most of the tests there, there is a `request_data` variable - just do `self.assertEqual(order.request, request_data)` in the test casses where the order is made successfully.

Leo6Leo: Replying to the first comment: but this is written within `OrderCreateSerializer`. The serializer will call itself, doesn't it cause recursion?

Leo6Leo: image
 If we use self.data, we would need to also fetch the object every time, will this be not efficient?

Leo6Leo: image
image

If we use `self.data`, it will throw out this error. I tried to look into this problem, but I have no idea how this happend....

hackathon_site/hardware/serializers.py
```diff
@@ -207,6 +208,16 @@ def create(self, validated_data):
         )
         new_order = None
         response_data = {"order_id": None, "hardware": [], "errors": []}
+
+        # The reason why doing this is because the id field stores the hardware object, django cannot translate hardware object into JSON. Therefore, loop has been used to call Profile seriliazer and assign the JSON file to the validated_data filed.
```
grahamhoyes: I'm not exactly sure where the error you're getting is coming from. Delete this block of code, and everything should work - it does on my machine

hackathon_site/hardware/serializers.py
grahamhoyes: I would need to see the stack trace to see where the error is coming from.

However, I can't reproduce the issue. I deleted the block of code on my end that uses the `counter` variable and such, made a post request, and everything worked as expected.

Leo6Leo: If we remove that counter block,  I will get this error. That's the reason why I just overwritten the validated_data to make it become JSON able
image

hackathon_site/hardware/serializers.py
Leo6Leo: See above

hackathon_site/hardware/serializers.py
```diff
@@ -211,12 +212,21 @@ def validate(self, data):
         return data
 
     def create(self, validated_data):
+
+        # make a copy of validated_data
+        validated_data_copy = validated_data
```
YuYing-Liang: might be better to use validated_data.copy() to ensure it's an actual copy of the variable

YuYing-Liang: might be better to use validated_data.copy() to ensure it's an actual copy of the variable

hackathon_site/hardware/serializers.py
YuYing-Liang: Okay, this makes sense, could you change "Profile serializer" to "Hardware serializer" in the comment

YuYing-Liang (commented):
Looks good!

YuYing-Liang (approved):

PR #334, opened 2021-11-24
Convert Dashboard and ItemTables to typescript

YuYing-Liang:
## Overview

- Resolves #IEEE-135
- converted Dashboard.js, ItemTables.js and their respective test files to typescript
- created mock orders (from order type) to mock whats being returned from the api
- pending orders table can now display multiple pending orders
- removed broken table and any buttons that have to do with incident reports
- removed return table (temporarily) until we have more clarity on how to handle those
- requires completion of https://github.com/ieeeuoft/hackathon-template/pull/326 

## Unit Tests Created

- changed itemtables from the old way of using redux (connected  & unconnected components) to the new way with actions & selectors --> changed the respective tests to reflect that
- commented out tests for broken and returned tables

## Steps to QA

- Go to dashboard page of hardware signout site


Mustaballer (approved):
Great Job!

PR #335, opened 2021-11-28
IEEE-147 and IEEE-148: Admin Hardware Pages

grahamhoyes:
## Overview

Makes a bunch of improvements for handling hardware from the admin site:
- Proper searchable list views for Hardware, Categories, Orders, Order Items, Event Teams, and Incidents
- Inlines for displaying related information. The coolest one is probably to display order iteams and incidents (in a rather hacky way) on the order page:
image
  - Due to the somewhat hacky way that incidents are displayed on the order page, they can't actually be created from there. This would require a custom form and a decent amount of work that isn't worth it. They can still be created directly from the admin incident page
  - The way that `quantity_remaining` works is very inefficent, as it makes a separate database query for every item in the order. If we expect that the admin site will actually see heavy use, then this queryset should be cached in redis at the beginning of the request.
- Adds [client side image cropping](https://github.com/koendewit/django-client-side-image-cropping/) for uploading hardware photos to the admin site, to make sure they are cropped to 600x600.

## Unit Tests Created

- None, we don't write tests for most of the admin site

## Steps to QA

- Play around with the different admin pages under the Hardware app


YuYing-Liang (dismissed):
very useful, looks good

Mustaballer (approved):
I was messing around with this PR on my side, and I didn't pick up on any issues!

PR #338, opened 2021-12-03
IEEE-121: Create an admin export resource for Review model

arshinmar:
## Overview

- Resolves #IEEE-121. 
- Created a review resource.
- Updated the review admin class to set the resource class to the above newly created review resource.

## Unit Tests Created

- None (admin portion not testing)

## Steps to QA

- Not 100% sure.


grahamhoyes:
Can you add a quick definition of the new environment variables to the main readme?

hackathon_site/review/admin.py
```diff
@@ -15,8 +17,52 @@
...
+            "review_status",
+            "rsvp",
+            "created_at",
+            "updated_at",
+        ]
+        return export_headers
+
+
 @admin.register(Review)
 class ReviewAdmin(admin.ModelAdmin):
```
grahamhoyes: Also needs the `ExportMixin`, like the application admin has

hackathon_site/review/admin.py
```diff
@@ -15,8 +17,52 @@
...
+class ReviewResource(resources.ModelResource):
+    class Meta:
+        model = Review
+        fields = (
+            "application__user__first_name",
+            "application__user__last_name",
+            "application__user__email",
+            "application__user__username",
+            "application__team__team_code",
+            "birthday",
```
grahamhoyes: These are fields on the applications, not the reviews.

The fields should be:
- All the application user ones you have, those are good
- `reviewer__email`,
- `interest`
- `experience`
- `quality`
- `reviewer_comments`
- `status`
- `decision_sent_date`
- `created_at`
- `updated_at`

Set the export headers accordingly as well. "reviewer_email" (one underscore instead of two) etc.

Leo6Leo (approved):
I didn't see any issues!!

Also, I have taken a look with this similar pr mentioned on Jira:
[Link](https://github.com/ieeeuoft/hackathon-template/blob/develop/hackathon_site/registration/admin.py#L29)

PR #341, opened 2021-12-04
Include Order Items in the Order API

grahamhoyes:
## Overview

The Order list API previously didn't give the `part_returned_health` of the order items, since it just included nested hardware.

Now it includes `items` directly, but no longer includes nested hardware - only hardware IDs. This is because the queryset annotation for `quantity_remaining` wouldn't work properly by including order items. Even before that it wasn't working with a prefetch related, so this view is now much more efficient.

On the frontend when fetching orders, we'll just need a separate request to fetch the associated hardware.

The response now looks like: 

```json
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 5,
            "items": [
                {
                    "id": 8,
                    "hardware_id": 1,
                    "part_returned_health": null
                },
                {
                    "id": 9,
                    "hardware_id": 2,
                    "part_returned_health": "Healthy"
                },
                {
                    "id": 10,
                    "hardware_id": 1,
                    "part_returned_health": null
                }
            ],
            "team_id": 4,
            "team_code": "781B5",
            "status": "Submitted",
            "created_at": "2021-11-27T12:22:57.241253-05:00",
            "updated_at": "2021-11-27T13:33:44.277462-05:00"
        }
    ]
}
```

## Unit Tests Created

- Updated the unit tests for the order serializer.

## Steps to QA

- Visit http://localhost:8000/api/hardware/orders/, make sure all is in order. Nothing on the frontend uses this API yet.

### DIFF
--- a/hackathon_site/dashboard/frontend/src/api/types.ts
+++ b/hackathon_site/dashboard/frontend/src/api/types.ts
@@ -78,8 +78,12 @@ export type OrderStatus = "Submitted" | "Ready for Pickup" | "Picked Up" | "Canc
 
 export interface Order {
     id: number;
-    hardware: Hardware[];
-    team: number;
+    items: {
+        id: number;
+        hardware_id: number;
+        part_returned_health: string | null;
+    }[];
+    team_id: number;
     team_code: string;
     status: OrderStatus;
     created_at: string;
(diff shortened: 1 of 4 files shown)


YuYing-Liang (approved):
Looks good! @Leo6Leo I think you might have some conflicts when this gets merged

PR #345, opened 2022-01-01
update the slices and selector

Leo6Leo:
## Overview

Make it under slices/event/teamSlice.ts 

The store should have keys for:

Loading, error, whatever else we have on the other ones

currentTeam, which uses the Team interface from the API types file

Create a getCurrentTeam thunk, and add reducers to the slice to store the current team. Use the /event/teams/team/ endpoint.

Write a selector for the current team and isLoading

Simple unit tests for the thunk (sets the loading flag while loading, errors properly, updates on success, etc).
## Unit Tests Created

- Not created yet


YuYing-Liang (approved):
Looks good!!

PR #354, opened 2022-03-05
IEEE-151: Get Hardware Order Items (Admin Side)

arshinmar:
## Overview

- Resolves #354 
- Created a new order item serializer, and wrote a base api function that filters by team code and order id

## Unit Tests Created

- Wrote tests that check team code/order id filtering, along with permissions.

## Steps to QA

- Review tests to ensure they tests different scenarios (although I already have imo)
- Verify once more on swagger.

### DIFF
--- a/hackathon_site/hardware/api_filters.py
+++ b/hackathon_site/hardware/api_filters.py
@@ -6,7 +6,7 @@ from hardware.serializers import (
     HardwareSerializer,
     OrderListSerializer,
     IncidentListSerializer,
-    OrderItemListSerializer
+    OrderItemListSerializer,
 )
 
 
@@ -25,13 +25,16 @@ class IncidentFilter(filters.FilterSet):
     hardware_id = filters.NumberFilter(field_name="order_item__hardware__id")
     team_id = filters.NumberFilter(field_name="order_item__order__team__id")
 
+
 class OrderItemFilter(filters.FilterSet):
     queryset = OrderItem
     serializer_class = OrderItemListSerializer
 
-    order_id = IntegerCSVFilter(field_name="order__id",
-        label="Comma separated list of team IDs",
-        help_text="Comma separated list of team IDs",)
+    order_id = IntegerCSVFilter(
+        field_name="order__id",
+        label="Comma separated list of order IDs",
+        help_text="Comma separated list of order IDs",
+    )
     team_code = filters.CharFilter(field_name="order__team__team_code")
 
 
--- a/hackathon_site/hardware/api_filters.py
+++ b/hackathon_site/hardware/api_filters.py
@@ -29,8 +29,10 @@ class OrderItemFilter(filters.FilterSet):
     queryset = OrderItem
     serializer_class = OrderItemListSerializer
 
-    hardware_id = filters.NumberFilter(field_name="order__id")
-    team_id = filters.NumberFilter(field_name="order__team_code")
+    order_id = IntegerCSVFilter(field_name="order__id",
+        label="Comma separated list of team IDs",
+        help_text="Comma separated list of team IDs",)
+    team_code = filters.NumberFilter(field_name="order__team__team_code")
 
 
 class HardwareFilter(filters.FilterSet):
--- a/hackathon_site/hardware/api_filters.py
+++ b/hackathon_site/hardware/api_filters.py
@@ -32,7 +32,7 @@ class OrderItemFilter(filters.FilterSet):
     order_id = IntegerCSVFilter(field_name="order__id",
         label="Comma separated list of team IDs",
         help_text="Comma separated list of team IDs",)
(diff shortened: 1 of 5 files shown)


hackathon_site/hardware/api_urls.py
```diff
@@ -15,4 +15,5 @@
         name="hardware-detail",
     ),
     path("orders/<int:pk>/", views.OrderDetailView.as_view(), name="order-detail",),
+    path("order_items/", views.OrderItemListView.as_view(), name="order-detail",),
```
YuYing-Liang: should be called "order-list" instead of "order-detail"

hackathon_site/hardware/api_filters.py
```diff
@@ -24,6 +25,13 @@ class IncidentFilter(filters.FilterSet):
     hardware_id = filters.NumberFilter(field_name="order_item__hardware__id")
     team_id = filters.NumberFilter(field_name="order_item__order__team__id")
 
+class OrderItemFilter(filters.FilterSet):
+    queryset = OrderItem
+    serializer_class = OrderItemListSerializer
+
+    hardware_id = filters.NumberFilter(field_name="order__id")
```
YuYing-Liang: change this variable name to order_id and we would want to be able to put multiple order ids so IntegerCSVFilter might be a be a better option, look at OrderFilter for example.

hackathon_site/hardware/api_filters.py
```diff
@@ -24,6 +25,13 @@ class IncidentFilter(filters.FilterSet):
...
+    team_id = filters.NumberFilter(field_name="order__team_code")
```
YuYing-Liang: call this team_code instead of team_id and team_code is a char field so use CharFilter instead of NumberFilter. Also, team is nested in order, not team_code so change the field name to "order__team__team_code"

hackathon_site/hardware/serializers.py
```diff
@@ -80,6 +80,25 @@ class Meta:
...
 
+class OrderItemListSerializer(serializers.ModelSerializer):
+    team_code = serializers.SerializerMethodField()
+    order_id = serializers.SerializerMethodField()
+
+    class Meta:
+        model = OrderItem
+        fields = (
+            "order_id"
+            "team_code"
```
YuYing-Liang: Looking at the Figma, you should also add created_at, updated_at , part_returned_health and hardware to the fields list

hackathon_site/hardware/api_filters.py
```diff
@@ -63,4 +71,4 @@ class OrderFilter(filters.FilterSet):
         help_text="Comma separated list of team IDs",
     )
     team_code = filters.CharFilter(field_name="team__team_code")
-    status = filters.CharFilter(field_name="status")
+    status = filters.CharFilter(field_name="status")
```
YuYing-Liang: nitpick: could you add a newline at the end of the file (same goes for all the files you changed)

hackathon_site/hardware/views.py
```diff
@@ -76,6 +77,36 @@ def post(self, request, *args, **kwargs):
         return self.create(request, *args, **kwargs)
 
 
+class IncidentListView(
```
YuYing-Liang: why are the incident changes here? You need to sync with develop I think

hackathon_site/hardware/views.py
```diff
@@ -76,6 +77,36 @@ def post(self, request, *args, **kwargs):
...
+    permission_classes = [FullDjangoModelPermissions]
+    queryset = Incident.objects.all().select_related("order_item__order__team")
+    serializer_class = IncidentListSerializer
+
+    def get(self, request, *args, **kwargs):
+        return self.list(request, *args, **kwargs)
+
+
+class OrderItemListView(mixins.ListModelMixin, generics.GenericAPIView):
+    search_fields = ("order__team", "order__id")
```
YuYing-Liang: should be order__team__team_code I'm pretty sure

YuYing-Liang (commented):
Just some initial thoughts, I haven't tested this with swagger locally yet but there are some things that are inconsistent in this PR which hopefully was the reason for the errors you were facing, lmk if there are still any errors after these changes tho

hackathon_site/hardware/api_urls.py
arshinmar: yep forgot, thanks

hackathon_site/hardware/views.py
arshinmar: yeah makes sense

hackathon_site/hardware/api_filters.py
arshinmar: ok sure

YuYing-Liang (approved):
looks good to me!

PR #355, opened 2022-03-06
remove entity adapter from team slice

YuYing-Liang:
## Overview

- removed entity adapter from team slice because not needed
- added selector for getting the team

## Unit Tests Created

- changed team slice tests

## Steps to QA

- no need to QA

### DIFF
--- a/hackathon_site/dashboard/frontend/src/slices/event/teamSlice.test.ts
+++ b/hackathon_site/dashboard/frontend/src/slices/event/teamSlice.test.ts
@@ -1,11 +1,11 @@
 import store, { makeStore, RootState } from "slices/store";
 import {
     teamReducerName,
-    teamSelectors,
     teamSliceSelector,
     isLoadingSelector,
     initialState,
     getCurrentTeam,
+    teamSelector,
 } from "slices/event/teamSlice";
 import { waitFor } from "testing/utils";
 import { mockTeam } from "testing/mockData";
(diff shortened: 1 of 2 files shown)


Mustaballer (commented):
Looks great! I’m just curious how I would be able to use the team selector to retrieve the names of the current team, and the team code additionally.

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.test.ts
```diff
@@ -47,6 +47,18 @@ describe("Selectors", () => {
         expect(isLoadingSelector(loadingTrueState)).toEqual(true);
         expect(isLoadingSelector(loadingFalseState)).toEqual(false);
     });
+
+    test("teamSelector returns the team", () => {
+        const teamState = {
```
YuYing-Liang, line 52: @Mustaballer the team is directly in the teamState. Take a look at this unit test, to get the current team do something like `const currentTeam = useSelector(teamSelector)` which gets the entire team object. Take a look at mockTeam object in mockData.tsx and the Team type in types.ts to see what currentTeam would look like. The users in the team should be nested under the users property which is an array of their data that you can loop through.

Mustaballer: Thanks for the clarification! Everything looks good to me

Mustaballer (approved):

PR #356, opened 2022-03-07
added redux persist to application

YuYing-Liang:
## Overview

- added redux persist library

## Unit Tests Created

- I don't think this library can be tested

## Steps to QA

- add items to your cart
- refresh the page, those items should still be there

### DIFF
--- a/hackathon_site/dashboard/frontend/package.json
+++ b/hackathon_site/dashboard/frontend/package.json
@@ -27,6 +27,7 @@
     "react-slick": "^0.26.1",
     "redux": "^4.0.5",
     "redux-mock-store": "^1.5.4",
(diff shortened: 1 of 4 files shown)


Mustaballer (commented):
Looks cool. So my understanding of redux persists is that it allows it to render certain parts of the website after the state has been retrieved and saved to the redux store.

YuYing-Liang:
@Mustaballer kind of but not quite, redux persist uses local storage (which is persistent browser storage) and updates it on every state change. It's then able to rehydrate the store (retrieve the data from local storage) when the web app loads. I'll do a demo during standup this week!

Mustaballer (approved):

PR #358, opened 2022-03-10
IEEE-128: updated categoryslice and test

YuYing-Liang:
## Overview

- Resolves #IEEE-128
- updated getCategories thunk to exhaustively get pages of categories successfully

## Unit Tests Created

- updated categorySlice.test.ts 

## Steps to QA

- in hackathon_site/settings/__init__.py set PAGE_SIZE (line 197) to something small, so it forces multiple category pages to load
- go to /inventory on hardware signout site
- categories should load in filters section without a problem


arshinmar (approved):
I can verify that the expected changes work on my end!

PR #361, opened 2022-03-20
IEEE-122: added export resource to custom user model

YuYing-Liang:
## Overview

- Resolves #IEEE-122
- made new custom user model
- added export resource to user model

image

## Unit Tests Created

- no tests for admin site :///

## Steps to QA

- Have a bunch of users
- go to the admin site
- you should see the user's active status, application status ("Applied" if you've applied or a dash if you haven't) and review status (dash if not reviewed yet) immediately in the table of users
- there should be an export button for all users, click that and it should export the right users with the right info to whatever file you want (please verify this)


Mustaballer (approved):
I tried different export formats, and it worked great!

PR #371, opened 2022-05-01
IEEE-129: update order slice

YuYing-Liang:
## Unit Tests Created

## Steps to QA


hackathon_site/dashboard/frontend/src/api/helpers.ts
```diff
@@ -0,0 +1,79 @@
+import {
+    Order,
+    OrderInTable,
+    OrderItemTableRow,
+    ReturnedItem,
+    ReturnOrderInTable,
+} from "./types";
+
+export const teamOrderListSerialization = (
```
Mustaballer, line 9: So the teamOrderListSerizalization essentially categorizes the orders?

hackathon_site/dashboard/frontend/src/api/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+    hardwareIdsToFetch: number[];
+} => {
+    const pendingOrders: OrderInTable[] = [];
+    const checkedOutOrders: OrderInTable[] = [];
+    const returnedOrders: ReturnOrderInTable[] = [];
+    const hardwareIdsToFetch: Record<number, number> = {};
+    orders.forEach((order) => {
+        if (order.status !== "Cancelled") {
+            const hardwareItems: Record<number, OrderItemTableRow> = {};
+            const hardwareRequested: Record<number, number> = {};
```
Mustaballer, line 24: What's the advantage of using Record type in these cases?

hackathon_site/dashboard/frontend/src/api/types.ts
```diff
@@ -99,10 +99,33 @@ export interface Order {
...
     status: OrderStatus;
+    request: {
+        id: number;
+        requested_quantity: number;
+    }[];
     created_at: string;
     updated_at: string;
 }
 
+/** Sanitized Orders */
```
Mustaballer, line 110: 'Sanitized" meaning that it will remove the unnecessary attributes correct?

hackathon_site/dashboard/frontend/src/components/general/AlertBox/AlertBox.test.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import AlertBox from "components/general/AlertBox/AlertBox";
+import { render } from "testing/utils";
+
+const mockErrorList = [
```
Mustaballer, line 5: Are these errors supposed to mock what a user might potentially have as reason for faulty hardware?

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.test.tsx
```diff
@@ -32,30 +32,43 @@ const mockedGet = get as jest.MockedFunction<typeof get>;
...
+        .mockResolvedValue(
+            withDelay ? promiseResolveWithDelay(teamAPIResponse, 500) : teamAPIResponse
+        );
 
 describe("Dashboard Page", () => {
     it("Renders correctly when the dashboard appears 4 cards and 3 tables", async () => {
-        const response = { data: mockTeam } as AxiosResponse<Team>;
-        when(mockedGet).calledWith(teamUri).mockResolvedValue(response);
-
+        mockTeamAPI();
```
Mustaballer, line 52: What does calling the function do?

hackathon_site/dashboard/frontend/src/slices/order/orderSlice.ts
```diff
@@ -1,32 +1,95 @@
...
 interface RejectValue {
     status: number;
     message: any;
 }
 
+export const getTeamOrders = createAsyncThunk<
+    APIListResponse<Order>,
+    undefined,
+    { state: RootState; rejectValue: RejectValue; dispatch: AppDispatch }
+>(`${orderReducerName}/geTeamOrders`, async (_, { rejectWithValue }) => {
```
Mustaballer: this should be getTeamOrders on line 45 right?

Mustaballer (commented):
When there are no orders, '00' appears under the team card. 
image

hackathon_site/dashboard/frontend/src/api/helpers.ts
YuYing-Liang: yes it serializes the backend data into a format that's usable on the frontend (in this case, the data we want to see in the order tables)

hackathon_site/dashboard/frontend/src/api/helpers.ts
YuYing-Liang: Record is like a generic javascript object where the first argument is the key type and the second is the value type, e.g. `Record<type1, type2>` creates the type
`{ [key: type1]: type2 }` and `Record<number, string>` creates the type `{ [number]: "string value"}`

YuYing-Liang: you can learn more from typescript documentation here: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

hackathon_site/dashboard/frontend/src/api/types.ts
YuYing-Liang: Sanitized in this case means your taking one object structure and changing it to another to fit your needs, so it's not "removing" unneccesary attributes but reorganizing them or using them to calculate other attributes.

hackathon_site/dashboard/frontend/src/components/general/AlertBox/AlertBox.test.tsx
YuYing-Liang: No, these are just random strings, they're used for making sure they display properly in the AlertBox component, take a look at the tests to see how they're used.

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.test.tsx
YuYing-Liang: This questions makes me think you browsed over the code without trying to understand it 🤔. Try to find where the mockTeamAPI() function is and look at the function body, you've definitely wrote the code in this function before so I think you'll know what it does. I put that code inside a function so I could reuse it in other unit tests which is a better practice than copy + pasting it over and over again.

hackathon_site/dashboard/frontend/src/slices/order/orderSlice.ts
YuYing-Liang: yes, thanks for catching!

YuYing-Liang:
> When there are no orders, '00' appears under the team card. image

Thank you! I will look into this

jameskokoska:
Looks great!

jameskokoska (dismissed):

hackathon_site/hardware/serializers.py
```diff
@@ -250,7 +250,8 @@ def validate(self, data):
     def create(self, validated_data):
 
         # make a copy of validated_data
-        validated_data_copy = validated_data.copy()
+        # validated_data_copy = validated_data.copy()
```
Mustaballer: Should't we remove this comment, since it's not in use

jameskokoska: Yea I was thinking the same thing! I think it will also be beneficial to comment that you removed the comment, in case one forgets in the future. It's always good to comment changes that become invisible like ghosts 👻

hackathon_site/dashboard/frontend/src/api/helpers.ts
Leo6Leo: so basically is this creating a single dict like what we did in python?

hackathon_site/dashboard/frontend/src/components/general/AlertBox/AlertBox.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+    "This particular reason",
+    "Another horrible excuse",
+    "As well as this unfortunate reason",
+];
+const mockSingleError = "Error is here for this single reason.";
+const mockErrorTitle = "An error has occurred due to the following reasons:";
+
+describe("Error Box displays correctly with the right information", () => {
+    it("Shows title and a list if error prop is an array", async () => {
+        const { getByText, getByTestId } = render(
```
Leo6Leo, line 15: I have seen many times the function `async ()` appears. I did some researches on that, it says The async and await keywords enable asynchronous, promise-based behavior to be written in a cleaner style, avoiding the need to explicitly configure promise chains. 

It is quite hard to understand. From my understanding, I think async means that the code written within the async will be executed at the same time, like verilog right?

hackathon_site/dashboard/frontend/src/components/general/AlertBox/AlertBox.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { Alert, AlertTitle } from "@material-ui/lab";
+
+interface ErrorBoxProps {
+    error: string[] | string;
+    type?: "error" | "info" | "success" | "warning";
+    title?: string;
+}
```
Leo6Leo, line 8: Note: We define the property for this component here

hackathon_site/dashboard/frontend/src/api/types.ts
Leo6Leo: its more like parsing the JSON data to other data type you can use in other part of the code I think. Data is the same, just the form is changed.

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
```diff
@@ -13,22 +13,11 @@ const CartErrorBox = () => {
...
-                                    ))}
-                                </ul>
-                            </>
-                        ) : (
-                            orderSubmissionError
-                        )}
-                    </Alert>
+                    <AlertBox
+                        error={orderSubmissionError}
+                        type="info"
```
Leo6Leo, line 18: we have pre-define the type as type?: "error" | "info" | "success" | "warning";

what if we pass the different value to here. Will this through the error? Or we can set a default value to handle this?

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.tsx
```diff
@@ -33,12 +36,13 @@ describe("<ChipStatus />", () => {
...
-        );
+        const store = makeStoreWithEntities({
+            pendingOrders: mockPendingOrdersInTable,
+        });
+        const { getByText, queryByText } = render(<PendingTable />, { store });
         expect(getByText(/pending orders/i)).toBeInTheDocument();
         expect(queryByText("In progress")).toBeInTheDocument();
-        mockPendingOrders.map(({ id }) => {
+        mockPendingOrdersInTable.map(({ id }) => {
             expect(getByText(`Order #${id}`)).toBeInTheDocument();
```
Leo6Leo, line 45: how react test works: test one component, and change the store and render the component with that store.

We read the text on the component, and compare it with the expected result.

hackathon_site/dashboard/frontend/src/api/helpers.ts
YuYing-Liang: @Leo6Leo it's not creating a dict it's just **defining the type** of an object (which is like a dict in python)

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
YuYing-Liang: good question! There are pre-defined types because this is an MUI component, if you do some digging into the source for for AlertBox, you'll see that the type is exactly as you guessed: `"error" | "info" | "success" | "warning"`
There will be an error if you pass in anything different because we are using typescript.

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.tsx
YuYing-Liang: Yes exactly in this test:
1. Render the component with the store with mock data
2. make assertions on the component, here I use getByText() which grabs text from the UI but you can make other assertions here as well, check out the jest documentation to learn more:

https://jestjs.io/docs/expect 
https://testing-library.com/docs/react-testing-library/cheatsheet

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
jameskokoska: TypeScript is always handy to have around for structured code! It feels like everything fits together in the end with TypeScript 🚀

hackathon_site/dashboard/frontend/src/components/general/AlertBox/AlertBox.test.tsx
YuYing-Liang: uhhh i don't think so but tbh i have forgotten all my verilog knowledge so idk what you're referring to 😅

You are correct, async await is used for promises. Promises are pieces of code that run asynchronously, so they don't run in the same timeline as other pieces of code. This can make it difficult to work with so async await is a solution to use promises (and other asynchronous entities) in a synchronous way to make it easier to program with. It's a difficult concept to understand but you'll get used to it! I think this article explains it pretty well: https://www.freecodecamp.org/news/async-await-in-javascript/ 

Also, this test is a bad example, I don't need async await! I'll remove it.

hackathon_site/hardware/serializers.py
YuYing-Liang: I'll remove the comment, thanks for catching!

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.test.tsx
```diff
@@ -7,12 +7,15 @@ import {
...
-    itemsCheckedOut,
     mockPendingOrders,
     mockCheckedOutOrders,
     mockHardware,
+    mockPendingOrdersInTable,
+    mockCheckedOutOrdersInTable,
+    mockReturnedOrdersInTable,
 } from "testing/mockData";
 import { RootStore } from "slices/store";
+import { OrderInTable, ReturnOrderInTable } from "../../../api/types";
```
Mustaballer: change to absolute import

hackathon_site/dashboard/frontend/src/testing/utils.tsx
```diff
@@ -30,6 +37,11 @@ import {
 } from "slices/hardware/cartSlice";
 
 import { teamReducerName, TeamState } from "slices/event/teamSlice";
+import {
+    initialState as orderInitialState,
+    orderReducerName,
+    OrderState,
+} from "../slices/order/orderSlice";
```
Mustaballer: change to absolute import

Mustaballer (commented):
Just a few changes to absolute imports

Mustaballer (approved):
Looks great!

PR #385, opened 2022-05-24
IEEE-13: GET: /api/hardware/incidents/<id>/

Mustaballer:
## Overview

- Resolves #IEEE-13
- created IncidentDetailView for an incident on participant side
-  added new URL for incident/<id>

## Unit Tests Created

- IncidentDetailViewGetTestCase


hackathon_site/hardware/views.py
```diff
@@ -82,6 +82,16 @@ def post(self, request, *args, **kwargs):
         return self.create(request, *args, **kwargs)
 
 
+class IncidentDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
+    queryset = Incident.objects.all()
+
+    serializer_class = IncidentListSerializer
```
YuYing-Liang: why are you using the incident list serializer for a detail view?

Mustaballer: I believe it still worked, do you think I should create a new serializer for the detail view?

YuYing-Liang:
@Mustaballer the serializer is fine, that was my bad I forgot what it looked like. But the name is misleading, could you refactor IncidentListSerializer to IncidentSerializer everywhere.

Also there seems to be a bit of confusion with the permissions on this endpoint. If you're ever curious on how this api will be used, look at the Figma! It will give you the best idea on whether it's more important on the participant or the admin side. In the case for this ticket, this api may be used on BOTH sides, so you have to create 2 endpoints:
- one for the admin side: can access all incidents accross all teams
- one for the participant side: can only access incidents for the team they're in

You can do both in this PR but its not required, if you only want to address one scenario, then make a ticket to complete the other.

Based on the title of this ticket `/api/hardware/incidents/<id>/` this PR is suppose to handle the admin side (which you've done so far) except the name of the api has the wrong naming convention.
You named it `/hardware/teams/team/incidents/{id}/`, there are a couple things to note:
- teams is an event app property so it should not be in a hardware api
- if you put teams in the url you are implying that this is supposed to be used on the participant side, because only participants have teams, admins do not have teams

So please change the url to `/hardware/incidents/<id>/` as written in the ticket.

You were kind of close here, for the participant side, the api would be `/api/event/teams/team/incidents/{id}` but the view would be placed in the event app folder.

Another small thing: in the future, don't put slashes in your branch names, only because usually if there's a slash it means a different branch folder, better not to confuse github 😅

hackathon_site/event/views.py
```diff
@@ -203,3 +205,13 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
 
     def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
+
+
+class TeamIncidentDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
+    queryset = Incident.objects.all()
+
+    serializer_class = IncidentListSerializer
+    permission_classes = [FullDjangoModelPermissions]
```
YuYing-Liang: Since this is a participant, you should check that they have logged in with a profile, since participants won't be given view permissions like this. 

You can use UserHasProfile permission class (the other views in api_views.py do the same)

hackathon_site/event/views.py
```diff
@@ -203,3 +205,13 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
...
+
+    def get(self, request, *args, **kwargs):
+        return self.retrieve(request, *args, **kwargs)
```
YuYing-Liang: You need to check whether the incident being queried belongs to the user's team, if it doesn't, then throw an unauthorized error.

Also, update the unit tests for both these comments.

hackathon_site/hardware/api_urls.py
```diff
@@ -1,7 +1,6 @@
 from django.urls import path
 from hardware import views
 
-
```
YuYing-Liang: Can you add this line back lol

hackathon_site/event/views.py
```diff
@@ -203,3 +205,13 @@ class TeamListView(mixins.ListModelMixin, generics.GenericAPIView):
 
     def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
+
+
+class TeamIncidentDetailView(mixins.RetrieveModelMixin, generics.GenericAPIView):
```
YuYing-Liang: You put this view in the wrong file (nw tho it's confusing) this file is for the jinja views on the hackathon site. Please move this to api_views.py so it's with the other api views.

YuYing-Liang (changes requested):
Also Mustafa, please refactor IncidentListSerializer to IncidentSerializer, we need the refactor now bc the serializer is being used in the list and detail view

hackathon_site/event/views.py
Leo6Leo: - [ ] Following up

hackathon_site/hardware/api_urls.py
```diff
@@ -1,14 +1,18 @@
...
 app_name = "hardware"
 
 urlpatterns = [
-    path("hardware/", views.HardwareListView.as_view(), name="hardware-list",),
+    path("hardware/", views.HardwareListView.as_view(), name="hardware-list"),
     path("orders/", views.OrderListView.as_view(), name="order-list"),
     path("categories/", views.CategoryListView.as_view(), name="category-list"),
     path("incidents/", views.IncidentListView.as_view(), name="incident-list"),
+    path(
+        "hardware/incidents/<int:pk>/",
```
YuYing-Liang: This will create an api called `"api/hardware/hardware/incidents/<int:pk>"` so remove the hardware part, the url should be `"incidents/<int:pk>/"` 
always check swagger to verify the api is correct before pushing code

hackathon_site/event/test_api.py
```diff
@@ -1,19 +1,21 @@
...
-
 from event.models import Profile, User, Team
 from event.serializers import (
     UserSerializer,
     TeamSerializer,
 )
 
-from hardware.serializers import OrderListSerializer
-from hardware.models import Hardware, Order, OrderItem
+from hardware.serializers import OrderListSerializer, IncidentSerializer
```
YuYing-Liang: unused imports here, but this file should not have changes, you can revert this file to match what's on develop by running
`git checkout develop -- hackathon_site/event/test_api.py` in the hackathon-template directory

hackathon_site/event/api_views.py
```diff
@@ -5,14 +5,13 @@
 from rest_framework.exceptions import ValidationError, PermissionDenied
 from rest_framework.response import Response
```
YuYing-Liang: this file shouldn't have changes, could you revert this file to what's on develop

hackathon_site/hardware/test_api.py
```diff
@@ -864,6 +864,89 @@ def test_successful_post(self):
             self.assertEqual(final_response[attribute], self.request_data[attribute])
 
 
+class IncidentDetailViewGetTestCase(SetupUserMixin, APITestCase):
+    def setUp(self):
+        super().setUp()
+        self.profile = Profile.objects.create(user=self.user)
```
YuYing-Liang: you're testing an admin endpoint, so a user will not have a profile, you can remove this.

hackathon_site/hardware/test_api.py
```diff
@@ -864,6 +864,89 @@ def test_successful_post(self):
...
+            content_type__app_label="hardware", codename="view_incident"
+        )
+
+        self.view = reverse("api:hardware:incident-detail", args=[self.incident.id])
+
+    def test_user_not_logged_in(self):
+        response = self.client.get(self.view)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_user_has_no_profile(self):
```
YuYing-Liang: here instead, you should test that a user should not be able to use this api if they have a profile but no view permissions.
you can use this line in your setup to create a profile in this function only: 
`Profile.objects.create(user=self.user)`

hackathon_site/hardware/test_api.py
YuYing-Liang: similarly, admins will not have teams so you can remove line 871 as well

hackathon_site/hardware/test_api.py
```diff
@@ -864,6 +864,89 @@ def test_successful_post(self):
             self.assertEqual(final_response[attribute], self.request_data[attribute])
 
 
+class IncidentDetailViewGetTestCase(SetupUserMixin, APITestCase):
+    def setUp(self):
+        super().setUp()
+        self.profile = Profile.objects.create(user=self.user)
+        self.team = Team.objects.create()
```
YuYing-Liang, line 870: Just a side note, your database setup here is fine but you have some objects that are unnecessary because you never user/reference them. This is a problem in other parts of our code too but we should aim for minimal code in the future.

1. you don't need to assign objects to "self" if you never reference them, e.g. self.order, self.hardware etc. you can simply call the objects.create() function and the data will be created.
2. you don't need to create so many objects. you never query the hardware or order items tables so you don't need to create multiple of them, always try to do the minimum amount when you create your setup data. So here you only need 1 hardware, 1 order_item and 1 incident
3. your data needs to add up. you first create an order with:
       2 items with hardware id 1
       3 items with hardware id 2
But you only create 1 order item with hardware id 1 and 1 order item with hardware id 2, meanwhile you should create as many order items as there are in the order you specify. Just something to keep in mind. Going off of point 2 though, you should change your order request to only request 1 item for minimal db setup.

YuYing-Liang (commented):
Just some minor changes to tests & unused files

hackathon_site/hardware/test_api.py
Mustaballer: Doesn't the Order object require a team, so wouldn't I keep line 871? With the team it passes the test cases.

YuYing-Liang: Ah true, keep it then.

YuYing-Liang (approved):
Looks good to me!

PR #386, opened 2022-05-26
IEEE-180: Added a check in cart summary for invalid team sizes

EthanHugh:
…team size is less than 2 or greater than 4, the button will be disabled.

## Overview

- Resolves #IEEE-180
- Added teamSizeSelector in teamSlice.ts that gets the number of team members in the current user's team.
- Added logic in CartSummary.tsx that checks your teamSize, and disables the button if this value is either less than 2 or greater than 4.

## Unit Tests Created

- None for now, currently creating one with Mustafa's help, as existing test cases failed

## Steps to QA

- Change team size to 1 or 5 and above to see if the submit order button becomes disabled

### DIFF
--- a/hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
+++ b/hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
@@ -10,6 +10,18 @@ const CartErrorBox = () => {
     const cartQuantity = useSelector(cartTotalSelector);
     const orderSubmissionError = useSelector(errorSelector);
     const teamSize = useSelector(teamSizeSelector);
+    const teamSizeTooLarge = teamSize > maxTeamSize;
+    const teamSizeTooSmall = teamSize < minTeamSize;
+    let errorMessage;
+    let errorTitle;
+
+    if (teamSizeTooSmall) {
+        errorMessage = "few";
+        errorTitle = "Small";
+    } else if (teamSizeTooLarge) {
+        errorMessage = "many";
+        errorTitle = "Large";
+    }
 
     return (
         <Grid direction="row" container>
@@ -33,16 +45,14 @@ const CartErrorBox = () => {
                         )}
                     </Alert>
                 )}
-                {teamSize < minTeamSize && (
+                {(teamSizeTooLarge || teamSizeTooSmall) && (
                     <Alert severity="warning" style={{ margin: "15px 0px" }}>
-                        <AlertTitle>{`Team Size Too Small!`}</AlertTitle>
-                        Sorry, you need to add more team members to place an order.
-                    </Alert>
-                )}
-                {teamSize > maxTeamSize && (
-                    <Alert severity="warning" style={{ margin: "15px 0px" }}>
-                        <AlertTitle>{`Team Size Too Large!`}</AlertTitle>
-                        Sorry, you have too many team members to place an order.
+                        <AlertTitle>{`Team Size Too ${errorTitle}!`}</AlertTitle>
+                        There are too {errorMessage} people on your team to place an
+                        order. We only allow teams between {minTeamSize} to{" "}
+                        {maxTeamSize} to checkout items. To join or leave a team please
+                        go to your Dashboard (link to dashboard page) and click the EDIT
+                        button under your team.
                     </Alert>
                 )}
             </Grid>
--- a/hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
+++ b/hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
@@ -3,8 +3,8 @@ import { useSelector } from "react-redux";
 import { cartTotalSelector, errorSelector } from "slices/hardware/cartSlice";
 import Grid from "@material-ui/core/Grid";
 import { Alert, AlertTitle } from "@material-ui/lab";
-import { teamSizeSelector } from "../../../slices/event/teamSlice";
-import { maxTeamSize, minTeamSize } from "../../../constants";
+import { teamSizeSelector } from "slices/event/teamSlice";
+import { maxTeamSize, minTeamSize } from "constants.js";
 
 const CartErrorBox = () => {
     const cartQuantity = useSelector(cartTotalSelector);
@@ -12,16 +12,8 @@ const CartErrorBox = () => {
     const teamSize = useSelector(teamSizeSelector);
     const teamSizeTooLarge = teamSize > maxTeamSize;
     const teamSizeTooSmall = teamSize < minTeamSize;
-    let errorMessage;
-    let errorTitle;
-
-    if (teamSizeTooSmall) {
-        errorMessage = "few";
-        errorTitle = "Small";
-    } else if (teamSizeTooLarge) {
-        errorMessage = "many";
-        errorTitle = "Large";
-    }
+    const errorMessage = teamSizeTooLarge ? "many" : teamSizeTooSmall ? "few" : "";
+    const errorTitle = teamSizeTooLarge ? "Large" : teamSizeTooSmall ? "Small" : "";
 
     return (
         <Grid direction="row" container>
@@ -48,11 +40,13 @@ const CartErrorBox = () => {
                 {(teamSizeTooLarge || teamSizeTooSmall) && (
                     <Alert severity="warning" style={{ margin: "15px 0px" }}>
                         <AlertTitle>{`Team Size Too ${errorTitle}!`}</AlertTitle>
-                        There are too {errorMessage} people on your team to place an
-                        order. We only allow teams between {minTeamSize} to{" "}
-                        {maxTeamSize} to checkout items. To join or leave a team please
-                        go to your Dashboard (link to dashboard page) and click the EDIT
-                        button under your team.
+                        {`There are too ${errorMessage} people on your team to place an
+                        order. We only allow teams between ${minTeamSize} to ${maxTeamSize} 
+                        to checkout items. To join or leave a team please
+                        go to your `}
+                        <a href="/">Dashboard </a>
+                        {` and click the EDIT
+                        button under your team.`}
                     </Alert>
                 )}
             </Grid>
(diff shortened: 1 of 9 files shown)


EthanHugh:
The test fails in CartSummary.test.tsx line 79. If i comment that line out, the test passes, so my guess is that the order-loading-icon does not appear when there is an invalid team size, however I do not know how to change the unit test to account for this.

hackathon_site/dashboard/frontend/src/components/cart/CartSummary/CartSummary.tsx
```diff
@@ -37,7 +39,12 @@ const CartSummary = () => {
                 color="primary"
                 variant="contained"
                 className={styles.btn}
-                disabled={cartQuantity === 0 || cartOrderLoading}
+                disabled={
+                    cartQuantity === 0 ||
+                    cartOrderLoading ||
+                    teamSize < 2 ||
```
YuYing-Liang: the team size constraints are constant but they may change based on what logistics decides, so instead of hardcoding the values here, can you make 2 constants in the src/constants.js file that have these constraints and reference them here.

YuYing-Liang: For some reason, this looks a bit ... messy (not ur fault tho, i would've done the same)
looking at the designs, I don't think we'll need to use the team size anywhere else so can you change your `teamSizeSelector ` to something like `teamMeetsSizeConstraintsSelector` that returns a boolean and do the min max logic there.

YuYing-Liang (changes requested):
Great start Ethan!

The reason why the test is failing is because you need to mock a team the redux store for that test. Because you don't mock the team the code thinks there are 0 members (the "else" case in your selector) which means the button is disabled so the test can't submit an order.

You'll need to mock the team in that unit test *and* create a unit test that makes sure the button is disabled when the team is not within the size constraints. Look around in the code for examples but if you need extra help just let mustafa or I know :)

YuYing-Liang:
I know this was put as optional in your ticket but we do need an error/warning message for the user explaining why the button is disabled (or else they'll be very confused). If you don't do it in this PR we'll create another ticket to address it and probably assign it to you, it needs to get done eventually, so feel free to implement that in this PR too if you want!

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -107,3 +108,10 @@ export const teamMemberNamesSelector = createSelector(
             (member) => member.user.first_name + " " + member.user.last_name
         )
 );
+
+export const teamSizeValidSelector = createSelector([teamSliceSelector], (teamSlice) =>
```
YuYing-Liang: Hi Ethan, so so sorry about this, but I realized we do need the team size somewhere else, could you change this selector back to how you had it before? 🙏🙏🙏

And in CartSummary.tsx could you do the minmax logic in a constant instead of directly in the jsx

YuYing-Liang, line 112: Hi Ethan, so so sorry about this, but I realized we do need the team size somewhere else, could you change this selector back to how you had it before? 🙏🙏🙏

And in CartSummary.tsx could you do the minmax logic in a constant instead of directly in the jsx

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.test.ts
```diff
@@ -74,6 +75,18 @@ describe("Selectors", () => {
         expect(teamCodeSelector(teamState)).toEqual(mockTeam.team_code);
     });
 
+    test("teamSizeSelector returns false", () => {
```
Mustaballer: In this case, the description should be "returning true", since you use a valid team

hackathon_site/dashboard/frontend/src/components/cart/CartSummary/CartSummary.test.tsx
```diff
@@ -11,7 +17,8 @@ import {
     promiseResolveWithDelay,
     when,
 } from "testing/utils";
-import { post } from "api/api";
+import { get, post } from "api/api";
+import Cart from "../../../pages/Cart/Cart";
```
Mustaballer: change to import Card from "pages/Cart/Cart";

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
```diff
@@ -3,10 +3,25 @@ import { useSelector } from "react-redux";
 import { cartTotalSelector, errorSelector } from "slices/hardware/cartSlice";
 import Grid from "@material-ui/core/Grid";
 import { Alert, AlertTitle } from "@material-ui/lab";
+import { teamSizeSelector } from "../../../slices/event/teamSlice";
```
YuYing-Liang: Absolute imports! It should be 
`"slices/event/teamSlice"` if you want pycharm to do this automatically go to Settings -> Editor -> Code Style -> JavaScript and you should be able to uncheck relative path imports

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
```diff
@@ -3,10 +3,25 @@ import { useSelector } from "react-redux";
...
+import { maxTeamSize, minTeamSize } from "../../../constants";
```
YuYing-Liang: Absolute imports here too, since constants is in the root file it'll be `"constants.js"`

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
```diff
@@ -3,10 +3,25 @@ import { useSelector } from "react-redux";
...
 
 const CartErrorBox = () => {
     const cartQuantity = useSelector(cartTotalSelector);
     const orderSubmissionError = useSelector(errorSelector);
+    const teamSize = useSelector(teamSizeSelector);
+    const teamSizeTooLarge = teamSize > maxTeamSize;
+    const teamSizeTooSmall = teamSize < minTeamSize;
+    let errorMessage;
```
YuYing-Liang: We try not to use if statements in JavaScript as much as possible, let's use ternary statements instead!

`const errorMessage = teamSizeTooLarge ? "many" : teamSizeTooSmall ? "few" : ""`

And do the same for errorTitle

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
```diff
@@ -30,6 +45,16 @@ const CartErrorBox = () => {
                         )}
                     </Alert>
                 )}
+                {(teamSizeTooLarge || teamSizeTooSmall) && (
+                    <Alert severity="warning" style={{ margin: "15px 0px" }}>
+                        <AlertTitle>{`Team Size Too ${errorTitle}!`}</AlertTitle>
+                        There are too {errorMessage} people on your team to place an
```
YuYing-Liang: For this error description, can you wrap this entire string in `${}` like you did with the AlertTitle. This non breaking string will detect white space so you don't need the `{" "}` in the middle of your sentence.

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -107,3 +107,7 @@ export const teamMemberNamesSelector = createSelector(
             (member) => member.user.first_name + " " + member.user.last_name
         )
 );
+
+export const teamSizeSelector = createSelector([teamSliceSelector], (teamSlice) =>
```
YuYing-Liang: Add a test where teamSlice.team doesn't exist and assert that teamSizeSelector returns 0

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -107,3 +107,7 @@ export const teamMemberNamesSelector = createSelector(
...
+    teamSlice.team ? teamSlice.team.profiles.length : 0
```
YuYing-Liang: Also, we can use nullish chaining & coalescing here for cleaner code

`teamSlice.team?.profiles?.length ?? 0`

YuYing-Liang (commented):
Nice job! 
Could you add two tests in CartErrorBox to test that the error message shows correctly for teams that are too large and too small. Add to the file index.test.tsx 

Also in the warning description, add a link to the dashboard page with `<a href="/dashboard">`

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
```diff
@@ -19,6 +27,18 @@ const CartErrorBox = () => {
                         title="Unable to submit your order because:"
                     />
                 )}
+                {(teamSizeTooLarge || teamSizeTooSmall) && (
+                    <Alert severity="warning" style={{ margin: "15px 0px" }}>
+                        <AlertTitle>{`Team Size Too ${errorTitle}!`}</AlertTitle>
```
Mustaballer, line 33: Is there a reason why 'Team Size Too' has capital letters instead of 'Team size too'?

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
```diff
@@ -19,6 +27,18 @@ const CartErrorBox = () => {
...
+                        {`There are too ${errorMessage} people on your team to place an
+                        order. We only allow teams between ${minTeamSize} to ${maxTeamSize}
```
YuYing-Liang, line 35: Sorry me english is bad, this sentence should be "We only allow teams between {min} to {max} **people** to checkout items"

hackathon_site/dashboard/frontend/src/components/cart/CartErrorBox/index.tsx
```diff
@@ -19,6 +27,18 @@ const CartErrorBox = () => {
...
                     />
                 )}
+                {(teamSizeTooLarge || teamSizeTooSmall) && (
+                    <Alert severity="warning" style={{ margin: "15px 0px" }}>
+                        <AlertTitle>{`Team Size Too ${errorTitle}!`}</AlertTitle>
+                        {`There are too ${errorMessage} people on your team to place an
+                        order. We only allow teams between ${minTeamSize} to ${maxTeamSize} 
+                        to checkout items. To join or leave a team please
+                        go to your `}
+                        <a href="/">Dashboard </a>
```
YuYing-Liang: image

here, "Dashboard" doesn't look very noticeable, could you use MUI's link component instead https://v4.mui.com/components/links/ to give it an underline

YuYing-Liang (commented):
Looks good! Just some minor comments

YuYing-Liang (approved):
great job!

PR #389, opened 2022-05-29
#IEEE-186: Added a Jinja filter to change system specific formatters on failure

terryluan12:
## Overview

- Resolves #186
- Added filter to change from window's datetime format to linux/mac and vice versa

## Unit Tests Created

- 

## Steps to QA

- Open http://localhost:8000/ (or wherever the server is hosted), and look at location date.

### DIFF
--- a/hackathon_site/event/jinja2/event/landing.html
+++ b/hackathon_site/event/jinja2/event/landing.html
@@ -32,7 +32,7 @@
                     <h5>Location</h5>
                     <h5>
                         {# Update this logic if your event ends in a different month or year #}
-                        {{  event_start_date | formatDateTime("%B %-d") }}-{{ event_end_date | formatDateTime("%B %-d") }}
+                        {{  event_start_date | formatDateTime("%B %-d") }} - {{ event_end_date | formatDateTime("%-d %Y") }}
                     </h5>
                 </div>
                 <div class="row">
--- a/hackathon_site/event/jinja2/event/landing.html
+++ b/hackathon_site/event/jinja2/event/landing.html
@@ -2,9 +2,9 @@
 
 {% block head_extends %}
 <script>
-    const registrationOpenDate = new Date("{{ registration_open_date | getDateTime }}");
-    const registrationCloseDate = new Date("{{ registration_close_date | getDateTime }}");
-    const eventStartDate = new Date("{{ event_start_date | getDateTime }}");
+    const registrationOpenDate = new Date("{{ registration_open_date | formatDateTime('%B %-d, %Y, %H:%M:%S') }}");
+    const registrationCloseDate = new Date("{{ registration_close_date | formatDateTime('%B %-d, %Y, %H:%M:%S') }}");
+    const eventStartDate = new Date("{{ event_start_date | formatDateTime('%B %-d, %Y, %H:%M:%S') }}");
 </script>
 {% endblock %}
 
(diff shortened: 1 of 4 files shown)


hackathon_site/event/jinja2/event/landing.html
```diff
@@ -32,7 +32,7 @@ <h4 class="header s12">Subtitle</h4>
                     <h5>Location</h5>
                     <h5>
                         {# Update this logic if your event ends in a different month or year #}
-                        {{ localtime(event_start_date).strftime("%B %-d") }}-{{ localtime(event_end_date).strftime("%-d, %Y") }}
+                        {{  event_start_date | formatDateTime("%B %-d") }}-{{ event_end_date | formatDateTime("%B %-d") }}
```
Leo6Leo: Add a space between the '-'. Currently it looks like this on my end. (MacOS)
CleanShot 2022-05-29 at 15 12 13@2x

Leo6Leo:
Great Job Terry. But I have a question here: why we are removing the year?

Or will that show different format on the different operating system? I am currently on MacOS.

This is the one before your change:
CleanShot 2022-05-29 at 15 14 34@2x


This is your change:
CleanShot 2022-05-29 at 15 12 13@2x

hackathon_site/hackathon_site/utils.py
```diff
@@ -15,3 +18,18 @@ def is_registration_open():
...
+
+
+def formatDateTime(datetime, format):
+    """
+    Format a datetime using strftime. 
+    """
+    local_zone = tz.tzlocal()
+    # Replaces incompatible identifiers.
+    # Add to this function if there are any datetime format issues
+    if platform == "win32":
```
Leo6Leo, line 30: We don't need to set it for MacOS right? because it is UNIX base. I am not sure.

Leo6Leo (changes requested):

hackathon_site/event/jinja2/event/landing.html
```diff
@@ -2,9 +2,9 @@
 
 {% block head_extends %}
```
Leo6Leo, line 3: Change the title of the ticket to the template I sent in the Slack Webmaster channel. 

And the reason why you are failing the Github check is because you didn't run black (the python file formatted) before you push your code. You should always run that before you commit/push.

Leo6Leo (changes requested):

hackathon_site/event/jinja2/event/landing.html
terryluan12: Thank you! The title has been changed, and the checks are all passing now

hackathon_site/event/jinja2/event/landing.html
terryluan12: Whoops! Thanks for the catch, I accidentally put in the wrong format after making the filter. It's been fixed!

hackathon_site/hackathon_site/utils.py
terryluan12: Nope! That's why it's only checking for windows.

Leo6Leo:
@terryluan12 Hey Terry if you resolved the issues, you can click on "Resolve conversation", and request the review again.

YuYing-Liang:
Will not use for now. Branch will still remain if needed.

PR #395, opened 2022-06-05
IEEE 63: Minimum and Maximum team size for orders

dalshekerchi:
## Overview

- Resolves #[IEEE 63](https://ieeeuoft.atlassian.net/jira/software/projects/IEEE/boards/1?selectedIssue=IEEE-63)
- Enforce a minimum and maximum team size to be able to place orders

## Unit Tests Created

- Tested case where the team was smaller than min
- Tested case where the team was larger than max
- Tested case where team was between 2-4

## Steps to QA

- Using the front end, one can add item to cart
- Go to cart and try to checkout
- If team has less than 1 or more than 4 a message will appear


hackathon_site/hackathon_site/__init__.py
```diff
@@ -0,0 +1,2 @@
+MIN_MEMBERS = 2
+MAX_MEMBERS = 4
```
YuYing-Liang: There might be some confusion here, could you put these constants in `hackathon_site/settings/__init__.py` we don't put anything in hackathon_site/__init__.py because our setup goes in the settings folder

in `hackathon_site/settings/__init__.py` could you put the constants here:
image

you can also put a small comment above it e.g. `# Team requirements` to follow our current format

You'll have to change your imports everywhere too. Wherever you use these constants you should be importing settings from django (not the hackathon_site folder) like this `from django.conf import settings` and then you should be able to do `settings.MIN_MEMBERS` or `settings.MAX_MEMBERS` to get the value of these constants.

hackathon_site/hardware/serializers.py
```diff
@@ -198,6 +200,9 @@ def validate(self, data):
             user_profile = self.context["request"].user.profile
         except ObjectDoesNotExist:
             raise serializers.ValidationError("User does not have profile")
+        team_size = Profile.objects.filter(team__exact=user_profile.team).count()
+        if team_size < MIN_MEMBERS or team_size > MAX_MEMBERS:
+            raise serializers.ValidationError("Does not match team size criteria")
```
YuYing-Liang: normally, we relate error messages to the user (because the user will be submitting the order, not the team), could you change this message to `"User's team does not meet team size criteria"`

hackathon_site/hardware/test_api.py
```diff
@@ -954,7 +963,9 @@ def test_user_has_no_profile(self):
 
     def test_create_simple_order(self):
         self._login()
-        profile = self._make_event_profile()
+        self.profile = Profile.objects.create(user=self.user, team=self.team)
```
YuYing-Liang: Nice job on the tests!
Looking at your changes, you're repeating creating 2 profiles in almost all the tests. Since this is repeated code, you can put it in a helper function in this class just to create 2 profiles.

under `setUp()`
```
def create_min_number_of_profiles(self):
   Profile.objects.create(user=self.user, team=self.team)
   Profile.objects.create(user=self.user2, team=self.team)
```
and call that function in your tests. Since you don't use `self.profile` or `self.profile2` anywhere, you don't need to assign the profiles to a variable.

hackathon_site/hardware/test_api.py
```diff
@@ -1708,3 +1743,108 @@ def test_failed_beginning_status(self):
             response.json(), {"status": ["Cannot change the status for this order."]},
         )
         self.assertFalse(request_data["status"] == Order.objects.get(id=self.pk).status)
+
+
+class MinMaxTeamOrderTestCase(SetupUserMixin, APITestCase):
```
YuYing-Liang, line 1809: Nice job for creating a new class! But we should put tests that belong to the same view in the same class. In your case you changed the OrderCreateSerializer which belongs to the OrderListViewPostTestCase class, could you move all your tests there. 

If it helps you can create more helper functions and create users/profiles in a for loop if the code gets too bloated.

hackathon_site/hardware/test_api.py
```diff
@@ -1708,3 +1743,108 @@ def test_failed_beginning_status(self):
...
+        self.profile5 = Profile.objects.create(user=self.user5, team=self.team)
+
+        request_data = {"hardware": [{"id": self.hardware1.id, "quantity": 2}]}
+        response = self.client.post(self.view, request_data, format="json")
+        self.assertEqual(
+            response.json(),
+            {"non_field_errors": ["Does not match team size criteria"]},
+        )
+
+    def test_team_correct_size_order(self):
```
YuYing-Liang: you can remove this test because almost all the tests in OrderListViewPostTestCase class test the correct team order, you only need to test if the constraints are not met (your 2 previous tests are sufficient)

YuYing-Liang (changes requested):
Really good job Dana! No significant code changes requested, just tests & some organizational stuff

Leo6Leo (dismissed):
It looks great to me! But don't merge it yet, wait for Yuying's response. Nice job Dana!

hackathon_site/hardware/test_api.py
```diff
@@ -1631,6 +1677,56 @@ def test_no_remaining_quantities(self):
         }
         self.assertEqual(response.json(), expected_response)
 
+    def create_order(self):
```
YuYing-Liang: could you move this helper function to the top of the class, under create_min_number_of_profiles, we try to keep them near the top so they're all in one place

YuYing-Liang (changes requested):
One last small thing Dana, otherwise your code looks great!

YuYing-Liang (approved):

PR #396, opened 2022-06-08
Ieee 152 move hardware in view to hardware slice

YuYing-Liang:
## Overview

- I moved the hardware entity in view to the hardware slice (it used to be in the ui slice) because it's easier to manage that way
- also moved the call for getUpdatedHardwareDetails out of product overview, this saves us a small step of not having to pass the hardware id into product overview but @KarandeepLubana if you don't like this I'll add the useEffect back, I'm still kind of debating which method is better

## Unit Tests Created

- unit tests for product overview & dashboard
- unit tests for hardwareslice for new state

## Steps to QA

- open product overview from dashbaord page (click info icon in checked out items table)
- open product overview from inventory page (click any hardware item)


KarandeepLubana:
I think either method is fine. We can keep it as is.

KarandeepLubana:
LGTM!

KarandeepLubana (approved):

PR #398, opened 2022-06-11
IEEE-179: Added Serializer for creation of profile, and post API endpoint to cr…

terryluan12:
...eate profile

## Overview

- Resolves #IEEE-179
- Added Serializer + POST API to allow for creation of a Profile/Team.

## Unit Tests Created

- Correctly creates profile when called
- If a user is already assigned to a profile, it is not assigned twice

## Steps to QA

- Test in Swagger


hackathon_site/event/serializers.py
```diff
@@ -86,6 +86,29 @@ def update(self, instance: Profile, validated_data):
         return serializers.ModelSerializer.update(self, instance, validated_data)
 
 
+class CreateProfileSerializer(ProfileSerializer):
+    class Meta(ProfileSerializer.Meta):
+        read_only_fields = ("id", "team", "attended")
+
+    def update(self, instance: Profile, validated_data):
```
Leo6Leo: Why would you want to use update method?

hackathon_site/event/test_api.py
```diff
@@ -55,6 +55,16 @@ def test_user_has_profile(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.json(), serializer.data)
 
+    def test_create_profile_for_new_user(self):
+        response = self.client.post(self.view, self.request_body)
+        data = response.json()
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(self.expected_response, data)
```
Leo6Leo: where did you define your expected_response?

hackathon_site/event/api_views.py
```diff
@@ -163,6 +164,23 @@ def patch(self, request, *args, **kwargs):
...
+    generics.GenericAPIView, mixins.CreateModelMixin,
+):
+    serializer_class = CreateProfileSerializer
+    queryset = Profile.objects.all()
+
+    def post(self, request, *args, **kwargs):
+        # create a new profile for the user
+        profile = Profile.objects.create(user=self.request.user)
+        team = profile.team
+
```
Leo6Leo: @YuYing-Liang Yuying, do we need to check whether the profile exists or not here before creating the profile object?

hackathon_site/event/serializers.py
Leo6Leo: Take a look at OrderCreateSerilizer.

Leo6Leo (dismissed):

YuYing-Liang:
@Leo6Leo please hold off from reviewing Terry's PR, I need to explain some things to him

hackathon_site/event/api_views.py
```diff
@@ -166,16 +167,33 @@ def patch(self, request, *args, **kwargs):
...
+            return CurrentProfileSerializer
+        elif self.request.method == "POST":
+            return CreateProfileSerializer
 
     def get_object(self):
         return self.request.user.profile
+    def get_permissions(self):
+        if self.request.method == "PATCH":
+            return [UserHasProfile()]
+        elif self.request.method == "POST":
```
YuYing-Liang, line 209: you need to add GET to the if statement.
and the permissions for the post method is whether the user is authenticated, you can use `[permissions.IsAuthenticated()]` here

hackathon_site/event/serializers.py
```diff
@@ -86,6 +86,22 @@ def update(self, instance: Profile, validated_data):
         return serializers.ModelSerializer.update(self, instance, validated_data)
 
 
+class CreateProfileSerializer(ProfileSerializer):
```
YuYing-Liang: I may have missed this from our conversation but I don't think you need a new serializer here. In CurrentProfileSerializer you can add your create function in there, no need to make a new serializer.

That also means that you don't need get_serializer_class method in your view and revert back to `serializer_class = CurrentProfileSerializer`

hackathon_site/event/serializers.py
```diff
@@ -86,6 +86,22 @@ def update(self, instance: Profile, validated_data):
...
+    class Meta(ProfileSerializer.Meta):
+        read_only_fields = ("id", "team", "id_provided", "attended")
+
+    def create(self, validated_data):
+        response_data = {"attended": True}
+        response_data["user"] = self.context["request"].user
+        response_data["id_provided"] = True
```
YuYing-Liang: this should be initialized to False

hackathon_site/event/serializers.py
```diff
@@ -86,6 +86,22 @@ def update(self, instance: Profile, validated_data):
         return serializers.ModelSerializer.update(self, instance, validated_data)
 
 
+class CreateProfileSerializer(ProfileSerializer):
+    class Meta(ProfileSerializer.Meta):
+        read_only_fields = ("id", "team", "id_provided", "attended")
+
+    def create(self, validated_data):
+        response_data = {"attended": True}
```
YuYing-Liang: you need to do some validation here with the user

if `self.context["request"].user.profile" exists then throw a ValidationError with a "user already has profile" message

hackathon_site/event/serializers.py
```diff
@@ -86,6 +86,22 @@ def update(self, instance: Profile, validated_data):
...
+        response_data["user"] = self.context["request"].user
```
YuYing-Liang: you don't need this since your serializer doesn't include user

YuYing-Liang (dismissed):
Left some initial comments. I remember in standup you said you wrote some tests? It would be great if you included those, and remember to run black!

hackathon_site/event/serializers.py
terryluan12: Yep! I have resolved most of the issues brought up, when I try to remove user though, it complains that it is missing a mandatory user_id within the request.

hackathon_site/event/serializers.py
```diff
@@ -85,6 +85,26 @@ def update(self, instance: Profile, validated_data):
         # ProfileSerializer update function and we want to keep the logic between these two update functions separate.
         return serializers.ModelSerializer.update(self, instance, validated_data)
 
+    def create(self, validated_data):
+        try:
+            self.context["request"].user.profile
```
YuYing-Liang: is this suppose to be an if statement?

YuYing-Liang: performing a try catch like this is a bit problematic, I'd prefer if you'd just use an if statement: if user.profile == None then raise exception

hackathon_site/event/serializers.py
```diff
@@ -85,6 +85,26 @@ def update(self, instance: Profile, validated_data):
...
+            raise serializers.ValidationError("User already has profile")
+        except:
+
+            response_data = {"attended": True}
```
YuYing-Liang: Just so you know, if this code did work, it would enter the except clause if an error is raised, which is when a user already has a profile so this logic doesn't make sense

hackathon_site/event/serializers.py
```diff
@@ -85,6 +85,26 @@ def update(self, instance: Profile, validated_data):
...
+            response_data["id_provided"] = False
```
YuYing-Liang: also, to write cleaner code (just a standard, doesn't affect functionality) you can populate the response_data object when you create it:
```
response_data = {
   "attended": True,
   "id_provided": False,
    etc.
}
```

hackathon_site/event/serializers.py
```diff
@@ -85,6 +85,26 @@ def update(self, instance: Profile, validated_data):
...
+            response_data["acknowledge_rules"] = validated_data.pop(
+                "acknowledge_rules", False
+            )
+            response_data["e_signature"] = validated_data.pop("e_signature", "NULL")
+            response_data["user"] = self.context["request"].user
+            response_data["id"] = self.context["request"].user.id
```
YuYing-Liang: the id field is a primary key so its automatically generated by django when the profile is created, you don't need to set it here.

hackathon_site/event/serializers.py
```diff
@@ -85,6 +85,26 @@ def update(self, instance: Profile, validated_data):
...
+
+            profile = Profile.objects.create(**response_data)
+            response_data["team"] = self.context["request"].user.profile.team
```
YuYing-Liang: why are you setting this after you create the profile?? What's your intention of doing this?

If you look into the Profile model in event/models.py you'll see that if not given a team when created, the Profile model will generate a new team for the profile. Meaning, you don't need it to create the profile but you should set it to `profile.team` not `self.context["request"].user.profile.team`

hackathon_site/event/test_api.py
```diff
@@ -461,10 +461,27 @@ def setUp(self):
...
+        # When posting, the user is assumed to be attending the hackathon
+        # The team should be increased by 1 since, the profile will be deleted
+        self.expected_response["attended"] = True
+        self.expected_response["team"] = 2
+
+        self.profile.delete()
+        self._login()
+        response = self.client.post(self.view, self.request_body)
+        data = response.json()
+
```
YuYing-Liang: here, you're only testing the api response but not actually the profile that's created. You need to query the profile model, `Profile.objects.get(user=self.user)` to make sure its created correctly and  use that profile's team in the response json

YuYing-Liang (changes requested):
Just curious, do you know how to test your API @terryluan12 ? Becuase just from running your code cold, I got a 500 error when trying to use the API, which tells me you didn't test your code before asking for a review .... 🤔

hackathon_site/event/serializers.py
terryluan12: Hey Yuying! I didn't notice the test wasn't passing, sorry! I reran the code after talking to Leo and Mustafa, and apparently if you run the tests after midnight, then an error occurs (for the email test). I just wanted clarification if the test that you got the 500 error on was different or if I just don't know how to run tests😓

terryluan12: In addition, with regards to the if else statement, I tried replacing it with `self.context["request"].user.profile == None`, but the issue was that if the profile isn't created, the "profile" attribute doesn't exist, and thus the code throws a  "RelatedObjectDoesNotExist" error. Similarly if I try `if user.profile == None`, I get a NameError, because the name 'user' doesn't exist.

Leo6Leo:
Hey Terry @terryluan12 , here is the update from me as I promised during our meeting:

Correct me if I am wrong Yuying @YuYing-Liang .
1. Cold code running
I have manually called your API  /event​/profiles​/profile​/ on swagger, and it shows me the Internal Server Error 500, which is the same case that Yuying commented on your PR. And I have added the screenshot for your convenience.

The potential reason that might cause this is that you didn't do the check to see whether the profile exists or not in the database. Because regarding your API, your job is to create the profile for the users. If the profile exists for that user, then there is no need to create one. I got this conclusion by reading the error message: " Key (id)=(1) already exists."

Or there might be some logical error in your API.

And from this 500 Error, we can see that your Django tests might not be working properly or your test cases are not covering all the corner cases or it is not testing the real API at all.

2. The "User has no profile" Bug you showed us during the meeting
I cannot replicate the error. When I run your Django tests, all of them(6 out of 6) passed without showing errors. 

3. Not NULL constraint failing bug you showed us during the meeting
Same as above. I cannot replicate the error. 
I have asked Mustafa to run the Django tests on his laptop and let's see what will happen.

Main take-home message: I would recommend you solve the 500 Internal Error first by going through your code in detail, creating a new account, and applying for the hackathon on your own server. Or reset your current database. 

This is usually how I debug with the API ticket: use PyCharm's debugger to set the breakpoint, and manually call the API. PyCharm will execute your code line by line, and until it hit the breakpoint. You will also be able to see the values of all variables. VSCode can also do that but I have no idea on how to set it up.

lmk if you have any other questions!

hackathon_site/event/serializers.py
YuYing-Liang: sorry @terryluan12, my python not so good atm the correct condition is `hasattr(self.context["request"].user, 'profile')` this will not throw any errors.

YuYing-Liang: Could you address all the PR comments first and I'll check the 500 error problem again

hackathon_site/event/serializers.py
```diff
@@ -85,6 +85,25 @@ def update(self, instance: Profile, validated_data):
         # ProfileSerializer update function and we want to keep the logic between these two update functions separate.
         return serializers.ModelSerializer.update(self, instance, validated_data)
 
+    def create(self, validated_data):
+        if hasattr(self.context["request"].user, "profile"):
+            raise serializers.ValidationError("User already has profile")
+
```
YuYing-Liang, line 94: This slipped my mind earlier, acknowledge rules and e-signature should be required in the request body. Could you check validated_data that both these properties are there and raise validation errors if they aren't.

The other easier thing you could do is remove the default value when you perform `validated_data.pop()`, without the default value django will assume it's required and automatically throw an error.

hackathon_site/event/test_api.py
```diff
@@ -461,6 +461,9 @@ def setUp(self):
         }
         self.view = reverse("api:event:current-profile")
 
+    def _build_view(self, acknowledge_rules, e_signature):
```
YuYing-Liang: remove, not used.

hackathon_site/event/test_api.py
```diff
@@ -522,6 +525,65 @@ def test_partially_modifying_profile(self):
...
+        self.expected_response = {
+            "attended": True,
+            "id_provided": False,
+            "acknowledge_rules": True,
+            "e_signature": "user signature",
+            "team": self.profile.team.team_code,
+        }
+        self.view = reverse("api:event:current-profile")
+
+    def _build_view(self, acknowledge_rules, e_signature):
```
YuYing-Liang: remove unused parameters

hackathon_site/event/test_api.py
```diff
@@ -522,6 +525,65 @@ def test_partially_modifying_profile(self):
...
+        return reverse("api:event:current-profile", kwargs=self.request_body)
+
+    def test_user_not_logged_in(self):
+        # Test that permissions work
+        response = self.client.post(self.view, self.request_body)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_user_can_has_profile(self):
```
YuYing-Liang: nit pick: name of function should be grammatically correct: `test_user_can_have_profile`

YuYing-Liang: also, you should combine this and the test below together, it doesn't make sense that they're separate because they do the same thing

YuYing-Liang: Also, remove comments in this class when you're finished implementing.

hackathon_site/event/test_api.py
```diff
@@ -522,6 +525,65 @@ def test_partially_modifying_profile(self):
...
+        # When posting, the user is assumed to be attending the hackathon
+        # The team should be increased by 1 since, the profile will be deleted
+
+        self.profile.delete()
+        self._login()
+        response = self.client.post(self.view, self.request_body)
+        data = response.json()
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(self.expected_response, data)
```
YuYing-Liang, line 559: the reason why this test is failing is because the expected response is incorrect. You can't create a profile in the setup and expect a newly created profile to have the same team code. Instead, the expected response should contain the team of the profile you just created from the database.

Also, I changed your response serializer so you'll have to add the profile id in the expected response as well.

hackathon_site/event/test_api.py
```diff
@@ -522,6 +525,65 @@ def test_partially_modifying_profile(self):
...
+    def test_user_can_actually_has_profile(self):
+        # Queries the database to test that the profile has actually been created properly
+        # When posting, the user is assumed to be attending the hackathon
+        # The team should be increased by 1 since, the profile will be deleted
+        self.profile.delete()
+        self._login()
+        response = self.client.post(self.view, self.request_body)
+        data = response.json()
+
+        database_profile = Profile.objects.get(user=self.user)
```
YuYing-Liang: for verifying that the profile in the database is correct you need to verify:
- the team is not null (the asserIsNotNone() below works for this)
- the attended field has the correct value, based on request
- the id_provided field has the correct value, based on request
- the e_signature field has the correct value, based on request
- the user field is correct based on the currently logged in user

hackathon_site/event/test_api.py
```diff
@@ -522,6 +525,65 @@ def test_partially_modifying_profile(self):
...
+
+        self.assertIsNotNone(database_profile)
+
+    def test_user_already_has_profile(self):
+        # "The bad path". The user attempts to create a new profile,
+        # when they already have a profile
+        self._login()
+        response = self.client.post(self.view, self.request_body)
```
YuYing-Liang, line 638: check the response body for an error message as well

hackathon_site/event/test_api.py
```diff
@@ -522,6 +525,65 @@ def test_partially_modifying_profile(self):
...
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+
```
YuYing-Liang, line 645: FYI, you need to test the new serializer you created, please do so in events/tests.py, you'll see similar test examples there.

hackathon_site/event/serializers.py
YuYing-Liang: you'll need to write tests verifying errors were being thrown.

YuYing-Liang (commented):
Nice effort @terryluan12, this PR is moving along. I left some comments on your serializers and tests. Also, I made a commit to change the response serializer you created to minimize code additions, the serializer is essentially the same though, lmk if you have any confusion about it!

hackathon_site/event/serializers.py
```diff
@@ -107,3 +123,7 @@ class Meta:
             "updated_at",
             "profiles",
         )
+
+
+class ProfileCreateResponseSerializer(ProfileSerializer):
```
YuYing-Liang, line 159: changed your serializer here to re-use existing code.

hackathon_site/event/tests.py
```diff
@@ -786,12 +786,6 @@ def test_serializer(self):
         }
         self.assertEqual(profile_expected, profile_serialized)
 
-    def test_readonly_serializer_fields(self):
```
YuYing-Liang, line 791: i removed these tests to accomodate the changed serializer, other functionality not affected.

hackathon_site/event/test_api.py
terryluan12: What do you mean by this? I have removed _build_view() as per the last comment, so I don't think there should be any more parameters at all?

YuYing-Liang: In this commit, you had 2 _build_view functions

hackathon_site/event/serializers.py
terryluan12: Unfortunately, even when removing the default values, they were still creating it for some reason (I think acknowledge_rules was automatically defaulting to false), so I just did some validation checking

hackathon_site/event/serializers.py
```diff
@@ -87,6 +86,40 @@ def update(self, instance: Profile, validated_data):
...
         return serializers.ModelSerializer.update(self, instance, validated_data)
 
+    def create(self, validated_data):
+        current_user = self.context["request"].user
+        if hasattr(current_user, "profile"):
+            raise serializers.ValidationError("User already has profile")
+
+        try:
+            review_status = Review.objects.get(application__user=current_user).status
+            if review_status != "Accepted":
```
Leo6Leo, line 108: Do we also need to check the RSVP status?

Leo6Leo (dismissed):
Great job! I have tested on my local device with all the possible corner cases, and everything works as expected.

See my one comment above.

hackathon_site/event/serializers.py
YuYing-Liang: good catch, yes we do!

YuYing-Liang (approved):

Leo6Leo (approved):
Other than the problem I sent you on Slack, the rest of the code looks good to me.

PR #406, opened 2022-06-25
IEEE-141: added import resources for hardware and category models

YuYing-Liang:
## Overview

- Resolves #IEEE-141
- Now we can import many hardware & categories from csv or excel files

## Unit Tests Created

- None bc django admin site

## Steps to QA

- run django migration `python manage.py migrate` to add new image_url field to hardware model
- go to admin site (http://localhost:8000/admin/hardware) and click "Import" on hardware page, do the same for category
- DM me for sample csv files to import


Mustaballer (commented):
Awesome work! I tested categories with a xlsx file and it imported correctly. I tested one hardware, and it also worked, however when I tried importing this csv, it kept loading indefinitely. I have the categories used in the image, and they are all unique and I don't have these items already, so I'm not sure what the issue is. I added this to the testing page in https://docs.google.com/spreadsheets/d/1MOzNtOBf4SXS2H3VUZskG5Dg-O1AIHyyuaK3tfjlQiI/edit#gid=8788834
image

Mustaballer (approved):
Great job! I tested the import hardware with this table of info, and it was successful. 
image

PR #409, opened 2022-06-26
IEEE-193: pending order table formatting fixes

YuYing-Liang:
## Overview

- Resolves #IEEE-193 (partially)
- Fixed Issues from IEEE-193-create-pending-order-table:
The table width is different for each table ✅
The cells are all fixed widths (I did use styles.nowrap from a different table css file but i removed that for now) ✅
The buttons at the bottom are meant to be right aligned but it does not appear to be so, it is currently hovering weirdly in the center ✅
The Qty Granted field needs two elements, but their vertical alignments are not matching as I could not find a way to center align items vertically ✅

image

## Unit Tests Created

- none, purely formatting PR

## Steps to QA

- go to localhost:3000/teams/<any team code> and look at the order tables
- play around with different screen sizes and see if anything looks funky


YuYing-Liang:
@EthanHugh here are some things I did:
- removed the Grid surrounding the table because there wasn't much use for it except for the buttons under the table
- I used flexbox for aligning the elements in the Quantity Granted column
- the buttons at the bottom were actually aligned fine, they didn't look like they were aligned because you added sizing to the previously surrounding grid container

lmk if you have any questions!
I see there's a lot of overlap between this table and the Pending Orders table from ItemTables.tsx so I'm going to generalize the css and jsx in another commit.

EthanHugh (approved):

PR #412, opened 2022-07-04
IEEE-208: added new api for user application status

YuYing-Liang:
## Overview

- Resolves #IEEE-208
- gets user's application status

## Unit Tests Created

- when user has a full application & review
- when user has a full application but no review
- when user has no application or review

## Steps to QA

- /api/event/users/user/review_status
- acceptance is None if application or review are incomplete and "Rejected", "Waitlisted" or "Accepted" if all are complete


hackathon_site/event/api_urls.py
```diff
@@ -6,6 +6,11 @@
 
 urlpatterns = [
     path("users/user/", api_views.CurrentUserAPIView.as_view(), name="current-user"),
+    path(
+        "users/user/acceptance",
```
Leo6Leo: just wondering if this name is a bit unclear, I think it would be better to change to "review_status" or something relate to application status.

hackathon_site/event/serializers.py
```diff
@@ -107,3 +109,18 @@ class Meta:
...
+
+
+class UserAcceptanceSerializer(serializers.ModelSerializer):
+    acceptance = serializers.SerializerMethodField()
+
+    class Meta:
+        model = User
+        fields = ("id", "first_name", "last_name", "email", "acceptance")
+
+    @staticmethod
```
Leo6Leo, line 121: Just want to confirm with you Yuying. Me and Mustafa saw this @static method in his ticket, and we did some research on that. @staticmethod means to make this function become static, which means it can be called without a instance. Normally we will do
`haha = User()
haha_response = haha.get_acceptance()
`
If we do static method, does that mean we can directly do like this?
`
get_acceptance()
`

hackathon_site/event/test_api.py
```diff
@@ -810,3 +810,46 @@ def test_cannot_change_other_team_order(self):
...
+
+
+class CurrentUserAcceptanceTestCase(SetupUserMixin, APITestCase):
+    def setUp(self):
+        super().setUp()
+        self.view = reverse("api:event:current-user-acceptance")
+
+    def get_expected_response(self, acceptance):
+        return {
+            **{
```
Leo6Leo, line 822: why there is `**` here?

Leo6Leo (commented):
Yuying @YuYing-Liang  ! I have performed the following test on my local server, and all of them passed. There are just some minor comments and potential changes I leave as the comment. I will approve your PR after you have replied to those conversations.

Test performed:
1. Create the account and try to call the API without submitting the application.
2. Use the account created in step 1 and submit the application but the application hasn't been reviewed by the admin account. Then try to call the API.
3. Open the incognito mode and login with my Admin account, and review the application I created in step 2. Assign the status as "Accepted". Then try to call the API.
4. With the application's status as "Accepted", use applicant account to **accept** the offer. And then call the API.
5. With the application's status as "Accepted", use applicant account to **decline** the offer. And then call the API.
6. Change the status to "Waitlisted". And then call the API.
7. Change the status to "Rejected". And then call the API.
8. Make the applicant account become "Inactive", and remain the status of the application as "Accepted", and then call the API.
9. Make the applicant account remain "Inactive", and change the status of the application as "Waitlisted", and then call the API.
10.  Make the applicant account remain "Inactive", and change the status of the application as "Rejected", and then call the API.
11. Change the applicant's account to "Active", and stay on the swagger page. Log out the applicant account, and call the API without refreshing the swagger page.
12. Perform automated Django tests. (Mailer test failed, but I believe it will pass on GitHub Action)

hackathon_site/event/api_urls.py
YuYing-Liang: sure, i can change it to review_status

hackathon_site/event/test_api.py
YuYing-Liang: this is a spread operator so it copies the entire object after the `**`
It works the same as javascript's `...`

hackathon_site/event/serializers.py
YuYing-Liang: Yeah good question! You are correct, you can use the function without declaring an instance of UserAcceptanceSerializer. You'd use it this way: 
```
from event.serializers import UserAcceptanceSerializer
UserAcceptanceSerializer.get_acceptance(user=<a user object>)
```
the static method decorator isn't entirely needed here because the function just needs to get the acceptance for the serializer. But since the function doesn't rely on the class, it's a static method anyways.

hackathon_site/event/test_api.py
YuYing-Liang: https://sparrow.dev/object-spread-operator-python/

hackathon_site/event/serializers.py
Leo6Leo: But why we still need `UserAcceptanceSerializer` in front of `get_acceptance(user=<a user object>)` ?

hackathon_site/event/test_api.py
Leo6Leo: I am still quite confused on this function.

`attr` has the same effect is . (i.e self.user.id == getattr(self.user, "id"))

I am confused that why the for statement is behind the attr? And what is the reason why we need to copy that?

Leo6Leo (approved):
Nice job Yuying! I didn't see any problems, but I left several questions here.

hackathon_site/event/serializers.py
YuYing-Liang: because that's what the @staticmethod decorator does for us. if we wanted to make a static method that didn't need the class, i'd put it outside the class but we still need it for the serializer to get the acceptance field.

hackathon_site/event/test_api.py
YuYing-Liang:
```
{
   attr: getattr(self.user, attr) for attr in ("id", "first_name", "last_name", "email")
}
```
this line here takes the properties in self.user and converts them to a dictionary.  The for loop is meant to iterate through all the attributes in the tuple. the result is:
```
{
    "id": 304958,
    "first_name": "blah blah",
     etc.
}
```
Then the `**` copies that ^ dictionary into the `expected_response` dict.

YuYing-Liang:
@Leo6Leo I'm going to merge and close the PR but you'll still be able to view it! Just save the URL.

hackathon_site/event/test_api.py
Leo6Leo: ohh that makes a lot of sense to me rn

hackathon_site/event/serializers.py
Leo6Leo: got it!!

PR #416, opened 2022-07-08
IEEE-206 leave join team thunk

Leo6Leo:
## Overview

- Resolves #IEEE-206
When a participant user goes to the Dashboard page on the HSS, they will see their team info. After they clicked on the Edit button, the Edit Team drawer will appear. The UI has been completed. So it is the time to make it functional!

Design: https://www.figma.com/file/1bYQy6ioaZwxPD9w6D47jW/Hardware-signout-site?node-id=3116%3A5303 

Ticket Type: Redux, React, API

Acceptance Criteria:

Users will be able to leave the team or switch between different teams

When the user is the only member in the team, they are not allowed to leave the team

if the team has made the order, then they are not allowed to switch teams.

Technical Help/Notes:

No

## Steps to QA

- Log in with your participant site account, and click on "Edit" on the dashboard to evoke the edit team drawer.


Leo6Leo:
@YuYing-Liang  Yuying I might need your help on determining whether the team has processed orders, and written the unit test. Other parts have been completed.

hackathon_site/dashboard/frontend/src/components/dashboard/EditTeam/EditTeam.tsx
```diff
@@ -63,8 +75,16 @@ const TeamChangeForm = ({ canChangeTeam }: TeamModalProps) => {
...
+                            disabled={
+                                !canChangeTeam
+                                    ? true
+                                    : isJoinTeamLoading
+                                    ? true
+                                    : value == ""
+                                    ? true
+                                    : value == teamCode
+                            }
+                            onClick={handleJoinTeam}
```
YuYing-Liang: you should use formik here to get the value of the team in the text field instead of the React useState hook

hackathon_site/dashboard/frontend/src/components/dashboard/EditTeam/EditTeam.tsx
```diff
@@ -109,8 +138,15 @@ export const EditTeam = ({ teamCode, canChangeTeam, teamSize }: TeamModalProps)
                     <TeamChangeForm
                         teamCode={teamCode}
                         canChangeTeam={canChangeTeam}
+                        canLeaveTeam={false}
                         teamSize={teamSize}
                     />
+                    {isJoinTeamLoading && (
```
YuYing-Liang: Instead of the loading bar being here, I think it would make more sense to put a circular loading icon on the submit or leave team buttons since that is where the action is initiated

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -69,6 +69,7 @@ const Dashboard = () => {
             <EditTeam
                 teamCode={team_code == null ? "None" : team_code}
                 canChangeTeam={true}
+                canLeaveTeam={team_size !== 1}
```
YuYing-Liang: I think this is a time where we leave the constraint handling to the backend because making all the warning/info messages on why someone can't leave the team is too much work & its not the main focus of the application. Focus on making the error handling really good and it will simplify this ticket.
In this case, you can remove the `canLeaveTeam` prop

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -52,6 +58,84 @@ export const getCurrentTeam = createAsyncThunk<
...
+                team_code: teamCode,
+            });
+
+            dispatch(
+                displaySnackbar({
+                    message: `You have left the team.`,
+                    options: { variant: "success" },
+                })
+            );
+
```
YuYing-Liang, line 84: here you can also dispatch the `closeTeamModalItem` action you created previously to close the side drawer

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -77,6 +161,47 @@ const teamSlice = createSlice({
...
+                if (payload) {
+                    state.isLeaveTeamLoading = false;
+                    state.error = null;
+                    state.team = payload;
+                }
+            }
+        );
+
+        builder.addCase(leaveTeam.rejected, (state, { payload }) => {
+            state.error = payload?.message || "Something went wrong";
```
YuYing-Liang, line 181: set the loading state to false here

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -52,6 +58,84 @@ export const getCurrentTeam = createAsyncThunk<
...
+    }
+);
+
+export const joinTeam = createAsyncThunk<
+    Team,
+    string,
+    { state: RootState; rejectValue: RejectValue; dispatch: AppDispatch }
+>(
+    `${teamReducerName}/joinTeam`,
+    async (teamCode, { dispatch, getState, rejectWithValue }) => {
```
YuYing-Liang: remove getState, its not being used

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -77,6 +161,47 @@ const teamSlice = createSlice({
...
+
+        builder.addCase(joinTeam.pending, (state) => {
+            state.isJoinTeamLoading = true;
+            state.error = null;
+        });
+
+        builder.addCase(
+            joinTeam.fulfilled,
+            (state, { payload }: PayloadAction<Team>) => {
+                if (payload) {
```
YuYing-Liang: don't need this condition here because payload will always be a non-null value

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -77,6 +161,47 @@ const teamSlice = createSlice({
...
+
+        builder.addCase(leaveTeam.pending, (state) => {
+            state.isLeaveTeamLoading = true;
+            state.error = null;
+        });
+
+        builder.addCase(
+            leaveTeam.fulfilled,
+            (state, { payload }: PayloadAction<Team>) => {
+                if (payload) {
```
YuYing-Liang: don't need this condition here because payload will always be a non-null value

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
YuYing-Liang: same for the joinTeam thunk

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -52,6 +58,84 @@ export const getCurrentTeam = createAsyncThunk<
...
+                displaySnackbar({
+                    message: `You have left the team.`,
+                    options: { variant: "success" },
+                })
+            );
+
+            return response.data;
+        } catch (e: any) {
+            // order reached quantity limits
+            const errorData = e.response?.data?.non_field_errors;
```
YuYing-Liang, line 88: Note that this is an array, so we may need to handle displaying multiple errors. Alternatively you could always take the first error in the list (assume it's the most important) and display it only.

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
```diff
@@ -91,6 +216,16 @@ export const isLoadingSelector = createSelector(
     (teamSlice) => teamSlice.isLoading
 );
```
YuYing-Liang, line 216: You should also create selectors for the error state. Because we're leaving all the error & constraint handling to the backend, you'll need a clearer way to display errors other than the snackbar.
Consider adding an error box in the modal if an error occurs:

YuYing-Liang (commented):
Nice start Leo!

hackathon_site/dashboard/frontend/src/components/dashboard/EditTeam/EditTeam.tsx
Leo6Leo: I would say I prefer the progress bar, here are the reasons:
1. The MUI version we are using currently is V4, which doesn't have a component called `<LoadingButton>`. Which means we need to write our own CSS to achieve the loading animation. On MUI's document, they give a demo, which looks really ugly from my perspective.
CleanShot 2022-08-21 at 16 44 14@2x

2. I love the animation 👻

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.tsx
Leo6Leo: By having this condition checked, we can avoid some meaningless API requests. And it is intuitive because we know if we only have 1 ppl in the team, we are not allowed to leave the team.

hackathon_site/dashboard/frontend/src/components/dashboard/EditTeam/EditTeam.tsx
Leo6Leo: done!

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
Leo6Leo: done

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
Leo6Leo: you are right yuying

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
Leo6Leo: you are right yuying

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
Leo6Leo: can you elaborate more on this Yuying? I tried to console.log the error, but I didn't find any array.

CleanShot 2022-08-21 at 17 00 28@2x

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
Leo6Leo: Currently, we use a notification box(snackbar) to display the error msg. You would like me to make an error box in the modal, and show the error in the box, instead of the notification box, is that correct?

CleanShot 2022-08-21 at 17 01 47@2x

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.tsx
YuYing-Liang: If you want to do that, you should also disable that button if the team has any orders. You will have to do the same for the join team button because users can't leave/join a team if they have any active orders. You can use the orderslice to get this information.

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
YuYing-Liang: Since you want to catch errors in the frontend before they occur, we don't need the extra error box the snackbar is fine as is.

hackathon_site/dashboard/frontend/src/pages/Dashboard/Dashboard.tsx
Leo6Leo: then I agree that handle this from the backend will be much easier! I will remove the "canLeaveTeam” check

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
Leo6Leo: So just to confirm that I don't need to make any changes to my current code right now right?

hackathon_site/dashboard/frontend/src/components/dashboard/EditTeam/EditTeam.tsx
```diff
@@ -36,41 +45,72 @@ const TeamInfoBlock = ({ teamCode, canChangeTeam, teamSize }: TeamModalProps) =>
...
     const [value, setValue] = React.useState("");
-
+    const dispatch = useDispatch();
+    const isJoinTeamLoading = useSelector(isJoinLoadingSelector);
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         setValue(event.target.value);
     };
+
+    const handleSubmitExternal = async (values: FormikValues) => {
+        console.log(values);
```
YuYing-Liang: remove the log

hackathon_site/dashboard/frontend/src/components/dashboard/EditTeam/EditTeam.tsx
```diff
@@ -36,41 +45,72 @@ const TeamInfoBlock = ({ teamCode, canChangeTeam, teamSize }: TeamModalProps) =>
...
+                                    disabled={
+                                        !canChangeTeam
+                                            ? true
+                                            : isJoinTeamLoading
+                                            ? true
+                                            : values.teamCode == ""
+                                            ? true
+                                            : values.teamCode == teamCode
+                                    }
+                                    // onClick={handleJoinTeam}
```
YuYing-Liang: remove comment

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
YuYing-Liang: I was probably thinking about something else, leave it as is. Could you remove the console log though

hackathon_site/dashboard/frontend/src/slices/event/teamSlice.ts
YuYing-Liang: yeah no changes needed

hackathon_site/dashboard/frontend/src/components/dashboard/EditTeam/EditTeam.tsx
```diff
@@ -36,41 +45,72 @@ const TeamInfoBlock = ({ teamCode, canChangeTeam, teamSize }: TeamModalProps) =>
...
-
+    const dispatch = useDispatch();
+    const isJoinTeamLoading = useSelector(isJoinLoadingSelector);
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         setValue(event.target.value);
     };
+
+    const handleSubmitExternal = async (values: FormikValues) => {
+        console.log(values);
+        dispatch(joinTeam(values.teamCode));
```
YuYing-Liang, line 52: The redux stuff looks great! the only minor thing is that when a user joins a new team, the new team's orders are not displayed. So could you dispatch `getTeamOrders` from orderSlice.ts right after you dispatch the joinTeam thunk

YuYing-Liang (approved):

PR #417, opened 2022-07-11
IEEE 149: POST: /hardware/orders/returns/ - return items

Mustaballer:
## Overview

- Resolves #IEEE-149
- request body =
```
{
  "hardware": [
    {
      "id": 0,
      "quantity": 0,
      "part_returned_health": [Healthy, Heavily Used, Broken, Lost]
    }
  ],
  "order": <order id> ,
}
```
- validate hardware exists in inventory
- update order item part_returned_health 

## Unit Tests Created

- in progress

## Steps to QA

- check endpoint on swagger

## Questions/Issues
- [x] I was able to replicate the request on swagger, is there anything I need to change?
- [x] To validate if hardware exists in the inventory how would I update the view to check?
- [x] I copied code from this serializer OrderCreateSerializer, what should I keep or remove?
- [x] We update the part_returned_health but what would the response look like? I imagine that would be handled in the view
- [x] Is there any issues you see rn with the code, and some resources I should look more into to better understand the ticket? 
- [x] Suggestions to tackle this ticket?


hackathon_site/hardware/serializers.py
```diff
@@ -343,3 +342,117 @@ class OrderCreateResponseErrorSerializer(serializers.Serializer):
...
+    order = serializers.PrimaryKeyRelatedField(
+        queryset=Order.objects.all(), many=False, required=True
+    )
+
+    def validate(self, data):
+        # get array of hardware and order id from data parameter
+        hardware_array = data["hardware"]
+        if len(hardware_array) < 1:
+            raise ValidationError(
+                "Hardware array size must be greater or equal to one."
```
YuYing-Liang: This is not very helpful to the user, what if they don't know what an array is? 
A better error message could be "No hardware specified in return request"

hackathon_site/hardware/serializers.py
```diff
@@ -343,3 +342,117 @@ class OrderCreateResponseErrorSerializer(serializers.Serializer):
...
+                )
+                # continue
+
+            max_available_quantity = hardware_item["quantity"]
+            if len(order_items_with_hardware) < hardware_item["quantity"]:
+                max_available_quantity = len(order_items_with_hardware)
+                response_data["errors"].append(
+                    {
+                        "hardware_id": hardware_item["id"],
+                        "message": "Requested quantity was {} higher than available. {} was returned ".format(
```
YuYing-Liang: code suggestion:
Instead of using .format() you can use the cool new python 3.6+ way of formatting strings like this (I changed up the first sentence bc of grammar)
```
f"Requested quantity of { hardware_item["quantity"] - len(order_items_with_hardware) } was higher than available. { max_available_quantity } was returned."
```
Also be careful of grammatical errors e.g. "1 was returned" vs. "2 **were** returned"

hackathon_site/hardware/serializers.py
```diff
@@ -343,3 +342,117 @@ class OrderCreateResponseErrorSerializer(serializers.Serializer):
...
+                order_items_in_order.filter(
+                    hardware=hardware_item["id"], part_returned_health__isnull=True
+                )
+            )
+
+            if len(order_items_with_hardware) == 0 and hardware_item["quantity"] > 1:
+                response_data["errors"].append(
+                    {
+                        "hardware_id": hardware_item.id,
+                        "message": "There are 0 order items for this hardware. No hardware items were returned",
```
YuYing-Liang: error message suggestion:
"There are no checked out items for this hardware, all have already been returned."

hackathon_site/hardware/serializers.py
```diff
@@ -343,3 +342,117 @@ class OrderCreateResponseErrorSerializer(serializers.Serializer):
...
+                                - len(order_items_with_hardware)
+                            ),
+                            max_available_quantity,
+                        ),
+                    }
+                )
+
+            for quantity_idx in range(max_available_quantity):
+                valid_status = False
+                for status in hardware_item["HEALTH_CHOICES"]:
```
YuYing-Liang: instead you can use this condition in python to eliminate the for loop:
```
if hardware_item["part_returned_health"] in HardwareItemReturnSerializer.HEALTH_CHOICES:
    < add error to array >
    continue
```
By placing the `continue` keyword in this if statement means you don't need the else block, it just makes your code look cleaner.

hackathon_site/hardware/serializers.py
```diff
@@ -343,3 +342,117 @@ class OrderCreateResponseErrorSerializer(serializers.Serializer):
...
+        return response_data
+
+
+class OrderItemReturnResponseSerializer(serializers.Serializer):
+    order_id = serializers.PrimaryKeyRelatedField(
+        queryset=Order.objects.all(), many=False, required=True
+    )
+    checked_out_items = OrderItemInOrderSerializer(many=True, required=True)
+    returned_items = OrderItemInOrderSerializer(many=True, required=True)
+    errors = OrderItemReturnSerializer(many=True, required=True)
```
YuYing-Liang: Why does the error property have this serializer?

hackathon_site/hardware/serializers.py
```diff
@@ -343,3 +342,117 @@ class OrderCreateResponseErrorSerializer(serializers.Serializer):
...
+                    order_items_with_hardware[quantity_idx].save()
+
+        currently_checked_out_items = OrderItem.objects.filter(order=order)
+        for order_item in currently_checked_out_items:
+            serialized_order_item = OrderItemInOrderSerializer(order_item).data
+            if order_item.part_returned_health is None:
+                response_data["checked_out_items"].append(serialized_order_item)
+            else:
+                response_data["returned_items"].append(serialized_order_item)
+
```
YuYing-Liang, line 432: General comment here:
I realized there could be multiple errors per hardware item and a better way to group these errors is by hardware id instead of a flat array, could you change the error structure to be:
```
errors: {
    <hardware id>: [ <array of strings (error messages)>],
}
```

YuYing-Liang (changes requested):
Nice start Mustafa!

hackathon_site/hardware/serializers.py
Mustaballer: Yeah I realized I should make a custome serializer. I mimicked OrderCreateResponseErrorSerializer
image

hackathon_site/hardware/serializers.py
Mustaballer: I haven't implemented the error structure yet but when I try to test hardware with an incorrect health status on swagger I get this error
`{'required': 'This field is required.', 'null': 'This field may not be null.', 'invalid': 'Invalid data. Expected a dictionary, but got {datatype}.'}
Internal Server Error: /api/hardware/orders/returns/`

Mustaballer: @YuYing-Liang I also noticed, that there isn't a returned status for the order which is why the order table still exists in the pending order section. Is that something I need to handle

image

YuYing-Liang (dismissed):

hackathon_site/dashboard/frontend/src/components/general/OrderTables/OrderTables.tsx
```diff
@@ -195,6 +196,7 @@ export const GeneralReturnTable = ({
 }) => {
     const hardware = useSelector(hardwareSelectors.selectEntities);
 
+    console.log(orders);
```
Mustaballer: remove log comment and everything should be good

YuYing-Liang (approved):

PR #423, opened 2022-07-28
IEEE 213: load user data into team detail page

lukecheseldine:
## Overview

- Resolves #IEEE-213
- loads user data into teamInfoTable on teamDetailPage
- i don't know how to write redux tests but will add them later

## Unit Tests Created

- none

## Steps to QA

- /teams/{team_id}
- make sure you enter a team id, not a team code. they are different and the page expects an id. check django admin site to find team ids


hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
```diff
@@ -7,21 +7,29 @@ import { RouteComponentProps } from "react-router-dom";
...
+import { getTeamInfoData } from "slices/event/teamDetailSlice";
 
 export interface PageParams {
     id: string;
 }
 
 const TeamDetail = ({ match }: RouteComponentProps<PageParams>) => {
+    // TODO: change api to use team_code instead of team_id
+    const teamCode = match.params.id;
+    // dispatch thunk here, it will populate store, then call the selector in teamInfoComponent to get team info data
```
YuYing-Liang: remove comment bc you already implemented it!

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
```diff
@@ -7,21 +7,29 @@ import { RouteComponentProps } from "react-router-dom";
...
+    useEffect(() => {
+        dispatch(getTeamInfoData(teamCode));
+    });
+
     return (
         <>
             <Header />
 
-            <Grid container direction="column" spacing={6} xs={12}>
+            <Grid container item direction="column" spacing={6} xs={12}>
```
YuYing-Liang: curious, why did you change this to a grid item? there's not parent Grid container for it.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -0,0 +1,98 @@
...
+    createSlice,
+} from "@reduxjs/toolkit";
+import { AppDispatch, RootState } from "slices/store";
+import { displaySnackbar } from "slices/ui/uiSlice";
+import { get } from "api/api";
+import { Team } from "api/types";
+
+const teamDetailAdapter = createEntityAdapter();
+
+const simpleState = {
```
YuYing-Liang: could you rename this to extraState, all our other slices name it that way (I only named it differently for the tutorial - forgot to mention that).

Also you need to make a typescript interface for this extraState so the typing for your selectors lines up.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -0,0 +1,98 @@
...
+    createEntityAdapter,
+    createSelector,
+    createSlice,
+} from "@reduxjs/toolkit";
+import { AppDispatch, RootState } from "slices/store";
+import { displaySnackbar } from "slices/ui/uiSlice";
+import { get } from "api/api";
+import { Team } from "api/types";
+
+const teamDetailAdapter = createEntityAdapter();
```
YuYing-Liang: add a type for this adapter (this will help your selector types line up), you are populating it with profiles so: `createEntityAdapter<Profile>()` will work

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+);
+
+const teamDetailSlice = createSlice({
+    name: teamDetailReducerName,
+    initialState,
+    reducers: {},
+    extraReducers: (builder) => {
+        builder.addCase(getTeamInfoData.pending, (state) => {
```
YuYing-Liang, line 69: nie job with the reducers, you can remove the comments though

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -0,0 +1,98 @@
...
+        } catch (e: any) {
+            dispatch(
+                displaySnackbar({
+                    message: `Failed to retrieve team info: Error ${e.response.status}`,
+                    options: {
+                        variant: "error",
+                    },
+                })
+            );
+            return rejectWithValue({
```
YuYing-Liang, line 56: **Error Handling**:
This api call is very important because it determines whether the team the user queried in the URL `/teams/{team code}` exists. Play around with the api on swagger and try to query an invalid team id. Look at the error response and make sure the message you're passing in here is correct.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -0,0 +1,98 @@
...
+            // when the api call completes
+            state.isLoading = false;
+            state.error = null;
+
+            teamDetailAdapter.setMany(state, payload.profiles);
+        });
+        builder.addCase(getTeamInfoData.rejected, (state, { payload }) => {
+            // when the api call fails
+            state.isLoading = false;
+            state.error = payload?.message ?? "something went wrong";
```
YuYing-Liang: nit pick: capitalize Something

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
```diff
@@ -7,21 +7,29 @@ import { RouteComponentProps } from "react-router-dom";
...
 import Typography from "@material-ui/core/Typography";
+import { useDispatch } from "react-redux";
+import { getTeamInfoData } from "slices/event/teamDetailSlice";
 
 export interface PageParams {
     id: string;
 }
 
 const TeamDetail = ({ match }: RouteComponentProps<PageParams>) => {
+    // TODO: change api to use team_code instead of team_id
```
YuYing-Liang, line 22: **Error Handling**:
Use the error selector you created here and in the JSX display an error message if the api returns an error.  If there is an error you shouldn't show anything else on this page.

Use component from components/general/AlertBox for the error message.

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
```diff
@@ -10,44 +11,56 @@ import {
...
 import React from "react";
+import { useSelector } from "react-redux";
+import {
+    isLoadingSelector,
+    teamDetailAdapterSelector,
+} from "slices/event/teamDetailSlice";
 
 export const TeamInfoTable = () => {
+    const isTeamInfoLoading = useSelector(isLoadingSelector);
+    const teamInfo = useSelector(teamDetailAdapterSelector.selectAll) as any[];
```
YuYing-Liang: after you type your adapter in the team detail slice, you won't need to do this anymore

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    it("Updates error status on API failure", async () => {
+        const mockResponse = makeMockApiResponse(mockTeam);
+        mockedGet.mockRejectedValueOnce(mockResponse);
+
+        const store = makeStore();
+        await store.dispatch(getTeamInfoData("1"));
+
+        await waitFor(() => {
+            expect(mockedGet).toHaveBeenCalledWith("/api/event/teams/1/");
+            expect(errorSelector(store.getState())).toBeTruthy();
```
YuYing-Liang: check that a snackbar has been dispatched, something like:
```
const actions = store.getActions();

expect(actions).toContainEqual(
       displaySnackbar({
            message: <message>,
            options: { variant: "error" },
       })
);
```

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
YuYing-Liang: You'll have to add a test in TeamDetail.test.tsx for this as well.

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.test.tsx
```diff
@@ -1,41 +1,66 @@
...
+describe("Team info table", () => {
+    it("Renders team info table", async () => {
+        const teamInfoApiResponse = makeMockApiResponse(mockTeam);
+
+        when(mockedGet)
+            .calledWith("/api/event/teams/1/")
+            .mockResolvedValue(teamInfoApiResponse);
+
+        const { getByText, queryByTestId, getAllByRole } = render(<TeamInfoTable />);
+
```
YuYing-Liang, line 27: **Code Style Comment**:
There is a lottt of iteration here, lines 26-35 seem unnecessary. You could combine lines 54-63 with a single loop:
```
mockTeam.profiles.forEach((profile) => {
   expect(profile.user.first_name).toBeInTheDocument();
   etc. 
})
```

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.test.tsx
```diff
@@ -1,41 +1,66 @@
...
+        const firstNames = mockTeam.profiles.map((user) => {
+            return user.user.first_name;
+        });
+        const lastNames = mockTeam.profiles.map((user) => {
+            return user.user.last_name;
+        });
+        const emails = mockTeam.profiles.map((user) => {
+            return user.user.email;
+        });
+        let numIdsProvided = 0;
```
YuYing-Liang: I see what you're doing here with the checkboxes but it doesn't test whether the correct row has a checked/unchecked box. I'd suggest using test-id on the checkbox to identify it but the way you did it before is also pretty good.

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.test.tsx
```diff
@@ -1,41 +1,66 @@
...
+        mockTeam.profiles.forEach((user) => {
+            if (user.id_provided) numIdsProvided++;
+        });
+
+        await waitFor(() => {
+            expect(queryByTestId("team-info-linear-progress")).toBeInTheDocument();
```
YuYing-Liang: your tests are failing because of this assertion. You don't call the api in this component so the loading spinner and the store will never be populated. 
You should split up this test: populate the store in the table test (you can use mockStoreWithEntities function from utils.ts) and verify the content here, then in TeamDetail.test.tsx mock the API and verify that it's being called when the page renders.

YuYing-Liang (changes requested):
Really great start Luke!

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
lukecheseldine: i found it resolved this warning i was getting in my browser dev tools console:
Screen Shot 2022-07-31 at 5 52 19 PM
but i realised that the xs prop isn't even needed, so i've removed both. leftovers from when i was trying to fix the styling in mobile view.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: i've changed the message to be "Team code not found". this is hardcoded bc i wanted to have a nice string to display. the api returned message, which i can get with e.response.data.detail, is "Not found." which doesn't look nice

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: are you sure that `Profile` is the correct type? when i use it i get this error when i attempt to populate the adapter:
Screen Shot 2022-07-31 at 6 08 36 PM
i believe payload.profiles which i am feeding into the adapter is an array of  `ProfileWithoutTeamNumber & UserWithoutProfile` objects, but i don't know how to declare this union of types in the annotation

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
lukecheseldine: i'm having trouble using the error selector. this is what my code looks like: 
Screen Shot 2022-07-31 at 7 25 53 PM
when i try to load team info for an invalid team id, like abc, the error is briefly displayed, and then the page just repeatedly refreshes. i think it is because the combination of the dispatch and useSelector calls force the page to reload, i guess because the state changes, and then the page gets stuck in a cycle. i don't know how to prevent this constant reloading from happening.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.test.ts
lukecheseldine: i'm having trouble getting the actions from the store. the `getActions()` function does not seem to exist:
Screen Shot 2022-07-31 at 7 40 19 PM
Screen Shot 2022-07-31 at 7 41 31 PM

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
YuYing-Liang: Yes you're correct that was my mistake. If you look in types.ts lines 85-87 that is definition of the profile array type. This union is also used on lines 73-75 so we should have made a type for it a long time ago lol.
What you can do is create a new interface in types.ts as follows (place it on line 70):
```
export interface ProfileWithUser extends ProfileWithoutTeamNumber {
  user: UserWithoutProfile;
}
```
now you can use `ProfileWithUser` on lines 85 and 73 instead of what's there now and use the interface to type your adapter.

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
YuYing-Liang: I think I need to see more of your render logic, we'll set up a call for this!

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.test.ts
YuYing-Liang: yes, I forgot to mention in my last comment that you'll need to write a new test for it and mock the store in a different way. This is the full implementation:
```
Outside your test block:
type DispatchExts = ThunkDispatch<RootState, void, AnyAction>;
const mockStore = configureStore<RootState, DispatchExts>([thunk]);

const mockInitialState: RootState = {
    ...store.getState(),
    [teamOrderReducerName]: {
        ...initialState,
    },
};

In your test block:
     <mock failure response here>

        const store = mockStore(mockInitialState);
        await store.dispatch(<your thunk here>);
        const actions = store.getActions();

        expect(actions).toContainEqual(
            displaySnackbar({
                message: <your error message>,
                options: { variant: "error" },
            })
        );
```

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
```diff
@@ -10,44 +11,56 @@ import {
...
 import React from "react";
+import { useSelector } from "react-redux";
+import {
+    isLoadingSelector,
+    teamDetailAdapterSelector,
+} from "slices/event/teamDetailSlice";
 
 export const TeamInfoTable = () => {
+    const isTeamInfoLoading = useSelector(isLoadingSelector);
+    const teamInfo = useSelector(teamDetailAdapterSelector.selectAll);
```
YuYing-Liang: Nit pick: a better name for this const would be "profiles"

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.test.tsx
```diff
@@ -15,7 +24,40 @@ test("renders without crashing", () => {
...
+        response: {
+            status: 404,
+            statusText: "Not Found",
+            message: "Could not find team code: Error 404",
+        },
+    };
+
+    mockedGet.mockRejectedValueOnce(failureResponse);
+
+    const teamDetailProps = {
```
YuYing-Liang: Since you're using teamDetailProps in both unit tests, you can turn this into a global constant.

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.test.tsx
```diff
@@ -1,12 +1,21 @@
...
+import { AxiosResponse } from "axios";
 
-test("renders without crashing", () => {
+jest.mock("api/api", () => ({
+    ...jest.requireActual("api/api"),
+    get: jest.fn(),
+}));
+
+const mockedGet = get as jest.MockedFunction<typeof get>;
+
```
YuYing-Liang, line 17: Could you wrap both these tests in a `describe("<TeamDetail />")` block

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
```diff
@@ -7,41 +7,56 @@ import { RouteComponentProps } from "react-router-dom";
...
+    const teamCode = match.params.id;
+    const dispatch = useDispatch();
+    useEffect(() => {
+        dispatch(getTeamInfoData(teamCode));
+    }, [dispatch, teamCode]);
+    const error = useSelector(errorSelector);
```
YuYing-Liang: Nit pick: declare error const before dispatch const (or at least before useEffect hook)

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.test.ts
```diff
@@ -0,0 +1,115 @@
...
+        );
+    });
+
+    it("Updates error state on API failure", async () => {
+        mockedGet.mockRejectedValueOnce(failureResponse);
+
+        const store = makeStore();
+        await store.dispatch(getTeamInfoData("abc"));
+
+        expect(errorSelector(store.getState())).toBeTruthy();
```
YuYing-Liang: Instead of just verifying the error selector is not null, you should verify that it has the correct error message

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -0,0 +1,99 @@
...
+    createEntityAdapter,
+    createSelector,
+    createSlice,
+} from "@reduxjs/toolkit";
+import { AppDispatch, RootState } from "slices/store";
+import { displaySnackbar } from "slices/ui/uiSlice";
+import { get } from "api/api";
+import { ProfileWithUser, Team } from "api/types";
+
+const extraState = {
```
YuYing-Liang: You need an interface to type your extra state

YuYing-Liang (commented):
Some small changes requested, but overall looks good!

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
lukecheseldine: k changed

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.test.tsx
lukecheseldine: yee

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.test.tsx
lukecheseldine: done

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
lukecheseldine: yup

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.test.ts
lukecheseldine: ok

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: addeddd :)

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.test.tsx
```diff
@@ -1,21 +1,60 @@
...
+    test("displays 404 error when the requested team id is not found", async () => {
+        const failureResponse = {
+            response: {
+                status: 404,
+                statusText: "Not Found",
+                message: "Could not find team code: Error 404",
+            },
+        };
+
+        mockedGet.mockRejectedValueOnce(failureResponse);
```
YuYing-Liang: I didn't catch this earlier but here you are mocking the entire get function. In the future we will need to use the get function for other things so this might cause some problems.
Instead you should use something like:

`when(mockedGet).calledWith(..).mockRejectedValue(failureResponse)`

^^ This might not be completely accurate but take a look at some other unit tests in the repo for some inspiration.

YuYing-Liang (commented):
Just 1 small comment about a test, everything else looks great 🙂

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.test.tsx
lukecheseldine: i used exactly what you suggested and the test still passes, so i think you're right!

YuYing-Liang (approved):
Amazing job!

PR #425, opened 2022-08-03
IEEE-216 replace <int:pk> with regex and update unit tests

KarandeepLubana:
## Overview

- Resolves [#IEEE-216](https://ieeeuoft.atlassian.net/jira/software/projects/IEEE/boards/1/backlog?selectedIssue=IEEE-216)

## Unit Tests Created

- N/A


hackathon_site/event/api_urls.py
```diff
@@ -39,7 +39,7 @@
         api_views.CurrentTeamOrderListView.as_view(),
         name="team-orders",
     ),
-    path("teams/<int:pk>/", api_views.TeamDetailView.as_view(), name="team-detail"),
+    re_path("teams/(?P<team_code>[A-Z0-9]{5})/", api_views.TeamDetailView.as_view(), name="team_code"),
```
YuYing-Liang: Don't need to change the name here, leave it as team-detail

hackathon_site/event/test_api.py
```diff
@@ -701,7 +701,7 @@ def setUp(self, **kwargs):
             content_type__app_label="event", codename="view_team"
         )
         super().setUp()
-        self.view = reverse("api:event:team-detail", args=[self.team.pk])
```
YuYing-Liang, line 704: Change this back

YuYing-Liang (commented):
What you also need to do is in TeamDetailView (in events/api_views.py) you need to tell Django which property of the team you need to query. You should do this with the lookup field property.

`lookup_field = "team_code"`

Then test it out on localhost:8000/swagger with team codes of the teams you have in your database. You can create more teams on the Django admin site.

YuYing-Liang (dismissed):
looks good to me!

YuYing-Liang:
@KarandeepLubana one thing you should also do is comment out the TODO in TeamDetail.tsx since that's what you implemented

YuYing-Liang (approved):
great job!

PR #426, opened 2022-08-07
Ieee 193 simple order fulfillment

YuYing-Liang:
## Overview

- Resolves IEEE-195 and IEEE-193
- The Table in IEEE-193 is not ready to be implemented with the current state of our API
- for NewHacks we'll stick with simple order fulfillment
- this PR also adds the return order table (pretty much copied from participant dashboard)

## Unit Tests Created

- Checkbox tests for the table

## Steps to QA

- Go to /teams/<team code> and play around with the order tables
- check that the tables on the participant dashboard are unaffected


YuYing-Liang:
@EthanHugh I'm setting you as a reviewer because this PR affects the pending order table you created

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
```diff
@@ -1,13 +1,16 @@
 import React from "react";
-
+import styles from "./TeamDetail.module.scss";
```
EthanHugh, line 2: Is this a relative import?

hackathon_site/dashboard/frontend/src/components/teamDetail/SimpleOrderTables/SimpleOrderTables.tsx
```diff
@@ -0,0 +1,233 @@
...
+const TableCheckbox = ({
+    field,
+    option,
+    order,
+    ...props
+}: FieldProps & { option: string; order: OrderInTable }) => {
+    const { values: formikValues, setFieldValue } =
+        useFormikContext<SimpleOrderFormValues>();
+
+    const recheckAllItems = (checked: boolean) => {
```
EthanHugh, line 41: This was actually such a smart way of doing this, I was confused for like 15 minutes till I realised what the purpose was LOL

hackathon_site/dashboard/frontend/src/components/teamDetail/SimpleOrderTables/SimpleOrderTables.tsx
```diff
@@ -0,0 +1,233 @@
...
+                }}
+                control={<Checkbox color="primary" />}
+                label=""
+                checked={field.value}
+            />
+        </FormGroup>
+    );
+};
+
+const getExtraColumn = (order: OrderInTable) => ({
```
EthanHugh: I don't think this name is descriptive enough because it makes it sound like the checkbox column is just an extra irrelevant column. Maybe call it something like enableCheckboxColumn for others to understand what column is being added while also putting importance to it

hackathon_site/dashboard/frontend/src/components/general/OrderTables/OrderTables.tsx
```diff
@@ -88,3 +102,206 @@ export const GeneralOrderTableTitle = ({
...
+    pendingOrder,
+    extraColumn,
+}: {
+    pendingOrder: OrderInTable;
+    extraColumn?: {
+        header: ReactElement;
+        body: (id: number) => ReactElement;
+    };
+}) => {
+    const hardware = useSelector(hardwareSelectors.selectEntities);
```
EthanHugh, line 116: It seems like the tables rendered don't have anything filled in the rows, is it due to this line here (instead of using ```hardware = mockHardware```) ?

hackathon_site/dashboard/frontend/src/components/teamDetail/SimpleOrderTables/SimpleOrderTables.test.tsx
```diff
@@ -0,0 +1,132 @@
+import { fireEvent, render, within, act } from "testing/utils";
+import { mockPendingOrdersInTable, mockReturnedOrdersInTable } from "testing/mockData";
+import React from "react";
+import {
+    AdminReturnedItemsTable,
+    SimplePendingOrderFulfillmentTable,
+} from "./SimpleOrderTables";
+import { waitFor } from "@testing-library/react";
+
+describe("<SimplePendingOrderFulfillmentTable />", () => {
```
EthanHugh, line 18: These tests look good! But I think you should add one more test that tests whether the checkboxes are shown only when the order status is ```"Submitted"``` (and otherwise not shown)

EthanHugh (changes requested):
Everything looks good to me but the only issue for me is that the table rows are unpopulated since we aren't using ```mockHardware```, but I think you probably have a reason for that. If not though I think you should then add another test to see if the table rows are rendered correctly.

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
YuYing-Liang: it is but it's okay for the css file since we want the styling to be in the same directory as the react file

hackathon_site/dashboard/frontend/src/components/teamDetail/SimpleOrderTables/SimpleOrderTables.tsx
YuYing-Liang: ty ty <3 this was very sweet

hackathon_site/dashboard/frontend/src/components/teamDetail/SimpleOrderTables/SimpleOrderTables.tsx
YuYing-Liang: ~~I orginally named it this way bc I wanted SimpleOrderTables to be generalized so this extra column could be anything, but now I don't know if that would happen 🤔 I'll change it to a more descriptive name and add a comment but I'd still like to keep it as a general extra column.~~
NVM you are right! I will rename :)

hackathon_site/dashboard/frontend/src/components/general/OrderTables/OrderTables.tsx
YuYing-Liang: Yeah good question! So here I integrated redux to get hardware details because this GeneralPendingTable component is used on the participant side too (where redux  is already fully integrated). This line requires you to have hardware in your database (based on the hardware ids in mockPendingOrders). When the table is tested, the redux store is populated with mockHardware.

YuYing-Liang:
@EthanHugh thanks for the first review, I implemented your feedback and I added a "picked up" button for ready to pick up order tables

hackathon_site/dashboard/frontend/src/components/general/OrderTables/OrderTables.tsx
```diff
@@ -88,3 +102,206 @@ export const GeneralOrderTableTitle = ({
...
+            <GeneralOrderTableTitle
+                orderId={pendingOrder.id}
+                orderStatus={pendingOrder.status}
+            />
+            <TableContainer component={Paper} elevation={2} square={true}>
+                <Table className={styles.table} size="small" aria-label="pending table">
+                    <TableHead>
+                        <TableRow>
+                            <TableCell className={styles.widthFixed} />
+                            <TableCell className={styles.width6} align="left">
```
EthanHugh, line 129: image
I think the table formatting is a bit too differentbetween the ```Ready for pickup``` orders and the ```In progress``` orders, which becomes more apparent on bigger screens. The ```Name``` column is too far to the right for the ```In progress``` order.

EthanHugh (changes requested):
I think this ticket should be almost complete, just for the one thing that I mentioned. Good work!

YuYing-Liang:
@EthanHugh thanks for catching that! I made some adjustments but there is a point where the tables do not look the same on smaller screens: 
image

This would be a bit hard to fix though because it's being caused by the checkboxes in the pending order table. This formatting only occurs on medium-small screens though such as on a tablet/iPad in portrait mode.

EthanHugh (approved):
This looks good now! I've noticed that the discrepancy actually becomes less apparent when the screen is smaller, for example when i zoom in it perfectly aligns like this:
image
But the discrepancy is so small for larger screens (when i zoom out) so I am approving this ticket!

PR #428, opened 2022-08-18
IEEE-200: handle checkbox change in team detail info table

lukecheseldine:
## Overview

- Resolves #IEEE-200
- when you click a checkbox in teamInfoTable, it reloads the table and then the id_provided field is updated for that user in the table, store, and database

## Unit Tests Created

- `teamDetailSlice.test.ts/updateParticipantIdProvided thunk`:
    - thunk updates store entity on success
    - thunk updates store error on failure
    - thunk dispatches snackbar on failure
- `teamInfoTable.test.tsx/Team info table`:
    - when checkbox is clicked, it becomes checked in the table

## Steps to QA

- go to `http://localhost:3000/teams/{team_id}` with the Admin Site Account
- click a checkbox in the ID column of the team info table


lukecheseldine:
right now when a checkbox is clicked, the entire table gets reloaded because there's only one loading state for the slice, which has two different thunks. i can change it to just disabling the checkboxes, i'll just need to create a separate loading state for each of the two thunks. lmk if you want me to do this.

also, i lost braincells writing the checkbox test in `teamInfoTable.test.tsx`, so if you can think of an easier way i'm all ears. react mui just makes it difficult with how input elements are wrapped in spans, and how you have to access properties of child elements to test them.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -30,6 +30,46 @@ interface RejectValue {
     message: any;
 }
 
+interface idProvidedParameters {
```
YuYing-Liang, line 37: I see what you're doing here but good news is patch requests don't require all properties in the request JSON to be there! you don't need the attended property and the request will still work, so you can remove it.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -30,6 +30,46 @@ interface RejectValue {
...
+export const updateParticipantIdProvided = createAsyncThunk<
+    Profile,
+    idProvidedParameters,
+    { state: RootState; rejectValue: RejectValue; dispatch: AppDispatch }
+>(
+    `${teamDetailReducerName}/updateParticipantIdProvided`,
+    async ({ profileId, idProvided, attended }, { rejectWithValue, dispatch }) => {
+        try {
+            const response = await patch<Profile>(`/api/event/profiles/${profileId}/`, {
+                id_provided: !idProvided,
```
YuYing-Liang: I feel like we should handle this logic in the TeamInfoTable component, keep this thunk generalized in case we want to use it elsewhere. 
So just pass in `idProvided` and handle the toggle logic in the checkbox onclick function.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -80,6 +120,26 @@ const teamDetailSlice = createSlice({
             state.isLoading = false;
             state.error = payload?.message ?? "Something went wrong";
         });
+        builder.addCase(updateParticipantIdProvided.pending, (state) => {
+            state.isLoading = true;
```
YuYing-Liang: you are right about the loading states, you should create a new loading state just for clicking the checkbox and update that here instead.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -80,6 +120,26 @@ const teamDetailSlice = createSlice({
...
+            state.error = null;
```
YuYing-Liang: For the error state, technically you should create a new state for the checkbox error too, but I don't think it's necessary here, dispatching the snackbar is enough. So you can remove all the error setting.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -80,6 +120,26 @@ const teamDetailSlice = createSlice({
...
+            state.isLoading = false;
+            state.error = null;
+
+            const updateObject = {
+                id: payload.id,
+                changes: {
+                    id_provided: payload.id_provided,
+                },
+            };
+            teamDetailAdapter.updateOne(state, updateObject);
```
YuYing-Liang, line 139: nice

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -80,6 +120,26 @@ const teamDetailSlice = createSlice({
...
+        });
+        builder.addCase(updateParticipantIdProvided.rejected, (state, { payload }) => {
```
YuYing-Liang, line 141: based on my previous comments, i think this case is not needed anymore

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
```diff
@@ -24,7 +24,7 @@ const TeamDetail = ({ match }: RouteComponentProps<PageParams>) => {
 
     // TODO: change api to use team_code instead of team_id
     const teamCode = match.params.id;
-    const error = useSelector(errorSelector);
+    const error: string | string[] | null = useSelector(errorSelector);
```
YuYing-Liang: I don't think you need these types, redux should automatically type it

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -30,6 +30,46 @@ interface RejectValue {
...
+        try {
+            const response = await patch<Profile>(`/api/event/profiles/${profileId}/`, {
+                id_provided: !idProvided,
+                attended: attended,
+            });
+            return response.data;
+        } catch (e: any) {
+            const message =
+                e.response.statusText === "Not Found"
+                    ? `Could not find participant: Error ${e.response.status}`
```
YuYing-Liang: error message suggestion:
`Could not update participant id status: Error {status code}`

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.test.tsx
```diff
@@ -45,4 +46,41 @@ describe("Team info table", () => {
             }
         });
     });
+    it("updates checkbox on click", async () => {
+        // NOTE: this test relies on at least one checkbox being rendered, so at least one profile in mockTeam
```
YuYing-Liang: I see what you mean about this testcase lol I'll play around with it and lyk the outcome. And reminder to use team_code instead of team id now that karandeep merged his branch to develop

YuYing-Liang (dismissed):
Really great start Luke 💯 you are becoming a true frontend guru.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: oh cool, fixed

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: fixed in teamDetailSlice...

lukecheseldine: fixed

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: removed rejected case for the new thunk

hackathon_site/dashboard/frontend/src/pages/TeamDetail/TeamDetail.tsx
lukecheseldine: ok

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: changed

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.test.tsx
lukecheseldine: okkk

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: ty :)

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: ok, error stuff removed

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: added new loading state

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.test.tsx
lukecheseldine: @YuYing-Liang i think i wrote all the tests before karandeep merged. they still work using team_id's though since there's no database to query against for the testing and i'm mocking the responses. do you want me to change all the tests using team_id, `api/event/teams/1`, to use team_code, `api/event/teams/ABCDE`, or can i just leave them as is?

YuYing-Liang: karandeep changed the tests to use team_code as well so to be consistent I think we should. It also avoids confusion if ppl ever look at this code in the future.

But I'm going to edit the test anyways so i'll change it

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -67,19 +107,35 @@ const teamDetailSlice = createSlice({
...
+            state.isTeamInfoLoading = false;
             state.error = payload?.message ?? "Something went wrong";
         });
+        builder.addCase(updateParticipantIdProvided.pending, (state) => {
+            state.isParticipantIdLoading = true;
+            state.error = null;
+        });
+        builder.addCase(updateParticipantIdProvided.fulfilled, (state, { payload }) => {
+            state.isParticipantIdLoading = false;
+            state.error = null;
```
YuYing-Liang: you aren't setting error to anything useful in these reducers so you can remove those lines.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -67,19 +107,35 @@ const teamDetailSlice = createSlice({
...
+
+            const updateObject = {
+                id: payload.id,
+                changes: {
+                    id_provided: payload.id_provided,
+                },
+            };
+            teamDetailAdapter.updateOne(state, updateObject);
+        });
```
YuYing-Liang, line 140: ok i'm dumb, you do need to handle the error case just to set the loading state to false, but you don't need to set the error state.

YuYing-Liang:
@lukecheseldine i noticed that the hardware site admins user group didn't have permissions to update profiles so i added them, you'll have to run a migration if you want to test it our yourself.
I also fixed the checkbox test :)

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
```diff
@@ -52,7 +56,16 @@ export const TeamInfoTable = () => {
                                         <Checkbox
                                             checked={row.id_provided}
                                             color="primary"
-                                            data-testid="checkbox"
+                                            disabled={isParticipantIdLoading}
+                                            data-testid={`id-provided-check-${row.id}`}
```
Leo6Leo, line 62: what is this data-testid used for? I am assuming it is used only for the jest test, is that correct?

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
```diff
@@ -30,6 +32,44 @@ interface RejectValue {
...
+    { state: RootState; rejectValue: RejectValue; dispatch: AppDispatch }
+>(
+    `${teamDetailReducerName}/updateParticipantIdProvided`,
+    async ({ profileId, idProvided }, { rejectWithValue, dispatch }) => {
+        try {
+            const response = await patch<Profile>(`/api/event/profiles/${profileId}/`, {
+                id_provided: idProvided,
+            });
+            return response.data;
+        } catch (e: any) {
```
Leo6Leo, line 54: If the profile not found, the error msg will show up correctly, but the loading state hasn't been updated. So the table stays in "disabled" state.

Leo6Leo: If we encountered the "Not Found" error, it means there is a profile updated in the database. We would need to get the most up-to-date profile from API. I would suggest querying for the profiles again here. When the most up-to-date profiles from API are retrieved, release the loading state.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
Leo6Leo: if we don't need error in the `fullfilled` state, we don't need to set state.error to null? If we don't, will it be automatically set to null?

Leo6Leo (changes requested):

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: makes sense, ill add it back

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: you're both right. `updateParticipantIdProvided` thunk never changes the error state (has no error state, even), so i don't have to reset it to null. i don't  think doing so will cause bad behavior or interfere with the error state of the other thunk, but it is unnecessary. changed

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
lukecheseldine: yes that is correct

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: you're right about the loading state. i added in a rejected case to set loading to false.

not sure i understand your second comment though @Leo6Leo. if we encounter "Not Found", that means there is no profile in the database matching the id that we queried for. in this case, i don't think anything gets updated in the database, because there is no matching profile to update. could you clarify what you mean and what you'd like me to change?

Leo6Leo: When we encounter "Not Found”, it means the information on the current webpage is not most up-to-date. We need to refresh the table to see the most up-to-date information in the database. Does this clear?

YuYing-Liang: @Leo6Leo not found just means the team doesn't exist like if you go to `/teams/12345` and team 12345 doesn't exist then the Not Fond error will appear

Leo6Leo: The reason why I mentioned this is because I did the following things:
1. open Luke's table, and the profiles displayed
2. Delete the profile in the database
At this time, the profile will still be there on Luke's table because we didn't refresh the page. When we click on the deleted profile's checkbox, it will say "Not Found". 

What I would suggest is to refresh the table after the 'Not Found" error appears.

YuYing-Liang: ah okay makes sense 👍

lukecheseldine: yes i understand now... i'll try just dispatching the thunk again inside the catch block if i get the "Not Found" error message

Leo6Leo: sounds good Luke

lukecheseldine:
@Leo6Leo ok i think i've addressed the bug leo, good catch!

i couldn't update the table in the rejected case itself because i would need to call the `getTeamInfoData` thunk again, but i don't have access to the teamCode within the rejected case, so i can't do that. instead, i created a new error state for the checkboxes. whenever the `updateParticipantIdProvided` thunk fails, it stores the error in the new error state. then i check in the `teamDetail` page if the error was 404. if it is, i can dispatch the `getTeamInfoData` thunk again, since i have access to the teamCode within this page. the table then updates and removes the user that was deleted from the database

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
Leo6Leo: or maybe try to call `getTeamInfoData` everytime when we have executed the `updateParticipantIdProvided` thunk? So that the table will always remain most up-to-date.

Leo6Leo: this could also solve the problem like this:

1. open Luke's table, and the profiles displayed
2. Add one more new team member
At this time, the new team member's profile will not be there on Luke's table because we didn't refresh the page.

Leo6Leo (changes requested):
See my comment in the conversation

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
lukecheseldine: yes that sounds good.

the only problem is that `updateParticipantIdProvided` gets called in the `teamInfoTable` component, but that component doesn't have access to the teamCode, so i can't call the `getTeamInfoData` thunk. i could pass teamCode down from the `TeamDetail` page as props, but idk if this is good practice or not. are you ok with me doing this leo?

lukecheseldine: eh nvm imma just do it 😃

lukecheseldine: @Leo6Leo so i tried calling both thunks every time a checkbox is clicked, but it doesn't work very well because the table reloads (loading bar) every time i click the checkbox. even if the users in the table don't change, it still triggers the loading bar because by calling the `getTeamInfoData` thunk, i'm updating the loading state of the table to true, so it displays the loading bar even though it doesn't have to. i think it's simpler to do it how i had it before, and admins will just have to refresh the page if a new user joins the team

YuYing-Liang:
@Leo6Leo @lukecheseldine if this gets too complicated I don't think we have to implement fetching members before clicking the checkbox because we have no apis to delete a profile and I can't forsee a reason that we'd have to delete a profile manually from the admin site. If the response returns a 404 error just display a different message saying that person isn't part of the team anymore and tell the user to refresh the page.

hackathon_site/dashboard/frontend/src/slices/event/teamDetailSlice.ts
Leo6Leo: sounds good. I agree with u and Yuying that we should leave it as is. And the thing I mentioned could be made as a ticket for the future.

Leo6Leo (approved):
Amazing work Luke! I don't think there are any other problems. Great job buddy.

PR #399, opened 2022-06-12
IEEE-188: Develop Admin Dashboard page

arshinmar:
## Overview

- Resolves #IEEE-188
- Created mock data and files across the tests.

## Unit Tests Created

- N/A

## Steps to QA

- Review the admin dashboard look, along with the mock data structure.

PR #432, opened 2022-09-08
Ieee 188 create dashboard overview

terryluan12:
## Overview

- Resolves #399 
- Added overview cards, and order cards

## Steps to QA

- Open Admin dashboard


hackathon_site/dashboard/frontend/src/components/orders/OrderCard/OrderCard.tsx
```diff
@@ -13,7 +13,10 @@ const OrderCard = ({ teamCode, orderQuantity, time, id }: OrderProps) => {
     const date = new Date(time);
     const month = date.toLocaleString("default", { month: "short" });
     const day = date.getDate();
-    const hoursAndMinutes = date.getHours() + ":" + date.getMinutes();
+    const hours = date.getHours() + ":";
```
YuYing-Liang, line 17: I don't think you're using these constants anywhere, so you can remove them

YuYing-Liang (commented):
Overall, the code is looking pretty good but I have some comments on the design:
image

For the Overview Cards I think it would be better to use shard corners (instead of rounded) which I think the Figma designs have too. The photo here is on my 13.5" laptop so you can see it's a bit squished, maybe try lowering the spacing and increasing the width of the cards?

Also, the "view more" button is purple for me but it should be using the "secondary" button styling e.g. like the "hide all" button for the order tables (team detail page)
image

As well, the pending order cards should be left-aligned (not centered) and could you add more order cards so we can see the horizontal scrolling.

Otherwise, great job! The page is really coming along 👍

hackathon_site/dashboard/frontend/src/components/orders/OrderCard/OrderCard.tsx
YuYing-Liang: bumping this again

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,85 @@ import React from "react";
...
+            flexWrap: "nowrap",
+            transform: "translateZ(0)",
+        },
+    });
+    const classes = useStyles();
     return (
         <>
             <Header />
             <Typography variant="h1">{hackathonName} Admin Dashboard</Typography>
+            <Grid className={styles.dashboard}>
```
YuYing-Liang, line 28: image
The spacing on the overview cards is very wide on large screens, i think a fixed size spacing is better (similar to the order cards). Also the overview cards will have a fixed number (refer to the figma) so there will be at most 5 cards which stack on smaller screens (again, refer to the figma https://www.figma.com/file/1bYQy6ioaZwxPD9w6D47jW/Hardware-signout-site?node-id=2640%3A23301&t=dQj2pNeHVtBsZp9c-0)

YuYing-Liang (changes requested):

hackathon_site/dashboard/frontend/src/components/orders/OrderCard/OrderCard.tsx
terryluan12: Which constants? I'm using the `hours` and `minutes` constant's in the `hoursAndMinutes` constant. The `hoursAndMinutes` constant is being used below as a part of the order card component

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,85 @@ import React from "react";
 import Typography from "@material-ui/core/Typography";
 import Header from "components/general/Header/Header";
 import { hackathonName } from "constants.js";
+import Grid from "@material-ui/core/Grid";
+import ImageList from "@material-ui/core/ImageList";
+import ImageListItem from "@material-ui/core/ImageListItem";
+import styles from "./AdminDashboard.module.scss";
+import { mockPendingOrders, overviewTitles } from "testing/mockData";
+import Paper from "@material-ui/core/Paper";
```
Leo6Leo, line 6: Change the order of these "import". Group all third-party import together, and group the components that made by ourself together.

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,85 @@ import React from "react";
...
+                    <ImageList
+                        className={classes.imageList}
+                        gap={8}
+                        cols={5}
+                        rowHeight={63}
+                    >
+                        {overviewTitles.map((title) => (
+                            <ImageListItem className={styles.overviewCard} key={title}>
+                                <Paper className={styles.overviewCard}>
+                                    <Card className={styles.overviewCard}>
```
Leo6Leo, line 40: add some horizontal offset for the shadow. It looks like "underline" on my end.
CleanShot 2023-02-17 at 01 18 16@2x

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,85 @@ import React from "react";
...
+                                        <Typography variant="body2">
+                                            &nbsp;&nbsp;&nbsp;{title}
```
Leo6Leo: I'm not sure whether this is a good habit or not. Usually we should use CSS to adjust this. But I think it should be good. What do you think? @YuYing-Liang  @Mustaballer

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,85 @@ import React from "react";
...
+                                        </Typography>
+                                    </Card>
+                                </Paper>
+                            </ImageListItem>
+                        ))}
+                    </ImageList>
+                </div>
+                <div className={styles.section}>
+                    <Typography className={styles.title}>
+                        Pending Orders &emsp; &emsp;{" "}
```
Leo6Leo: Same here as I mentioned in the previous comment

Leo6Leo (changes requested):
Keep your good work as you always do Terry!!!

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
Leo6Leo: @YuYing-Liang @Mustaballer

lukecheseldine:
closed bc letting associates do this work instead now that Terry isn't on web

PR #433, opened 2022-09-12
IEEE-221: add phone number to profile model and update all subsequent serializers

lukecheseldine:
## Overview

- added phone_number field to profile model with validation
- updated serializers
- updated unit tests

## Unit Tests Created

- none

## Steps to QA

- apply migrations (you will probably have to provide a default phone number when migrating the profiles that already exist in your local db, since the field is required and has no default)
- check the django admin page and test out some of the relevant event/profile apis on swagger to see that the phone number is now there
- `event.test_api.CreateProfileViewTestCase.test_user_can_have_profile` is still failing and i can't figure out why


YuYing-Liang:
@lukecheseldine the reason the test may be failing is because in the create profile api, you don't copy the phone number from the user's application to the profile, so when you're calling that api the Profile object is incomplete.
To do this look at the `create()` function in the `CurrentProfileSerializer` class. You'll have to add `phone_number` to the `response_data` object. To get the user's phone number try using the django orm query `Application.objects.get(user=self.context["request"].user).phone_number`
Please also manually test the create profile api in swagger before you alter the tests.

Lmk if you have any further questions!

lukecheseldine:
> @lukecheseldine the reason the test may be failing is because in the create profile api, you don't copy the phone number from the user's application to the profile, so when you're calling that api the Profile object is incomplete. To do this look at the `create()` function in the `CurrentProfileSerializer` class. You'll have to add `phone_number` to the `response_data` object. To get the user's phone number try using the django orm query `Application.objects.get(user=self.context["request"].user).phone_number` Please also manually test the create profile api in swagger before you alter the tests.
> 
> Lmk if you have any further questions!

lukecheseldine:
@YuYing-Liang you were right! i updated the create method like you said and all tests pass. i also tested it in swagger by creating a new user and then using the api to create a profile for them, and it seems to have worked.

i'm wondering if i need to change the update method too? i assume this is for the patch endpoint on swagger, but i couldn't actually test it because it says i have an invalid CSRF token when i try to make a call to that endpoint on swagger

hackathon_site/event/models.py
```diff
@@ -27,6 +28,16 @@ class Profile(models.Model):
     team = models.ForeignKey(
         Team, related_name="profiles", on_delete=models.CASCADE, null=False
     )
+    phone_number = models.CharField(
+        max_length=20,
+        null=False,
+        validators=[
```
YuYing-Liang: the validators argument is required in the registration model because we use a django form to create an application. Here, there will be no form so you can remove it.

YuYing-Liang:
> i'm wondering if i need to change the update method too? i assume this is for the patch endpoint on swagger, but i couldn't actually test it because it says i have an invalid CSRF token when i try to make a call to that endpoint on swagger

Ethan also got this CSRF token error but I can't seem to replicate it. Apparently it just fixes itself sometimes 🤷‍♀️
But since the patch requests share the same serializer as the create profile post request, you already added phone number! Just check the expected response under the api in swagger 

<image src="https://user-images.githubusercontent.com/20618460/190316872-7d23b42f-bf29-4391-8f03-ba944800026f.png" width="300" />

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
```diff
@@ -53,7 +51,15 @@ export const TeamInfoTable = () => {
                                         {`${row.user.first_name} ${row.user.last_name}`}
                                     </TableCell>
                                     <TableCell>{row.user.email}</TableCell>
-                                    <TableCell>802-207-9999</TableCell>
+                                    <TableCell>
+                                        {`${row.phone_number.substring(
```
YuYing-Liang: This is fine but some people add their area code in front of their phone number (we allow it in their application) so we have to handle cases for that.

Maybe try implementing something like this: https://learnersbucket.com/examples/javascript/how-to-format-phone-number-in-javascript/ 
but incorporate the regex from our application form

YuYing-Liang (changes requested):
Nice job Luke!

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
Leo6Leo: CleanShot 2022-09-15 at 01 11 26@2x

From the application form

Leo6Leo (commented):
Great works Luke! I have read through your code, and I don't have any further comments right now.

hackathon_site/event/models.py
lukecheseldine: removed

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
lukecheseldine: ok so i added a little phone number formatter thing. you can play with it here https://jsfiddle.net/cst43qku/12/

it's quite simple, but should work for the phone numbers in the application. basically i assume the phone number is valid (it's checked against the regex validator in the registration model) and just get all the digits. i then map the last 10 to (xxx) xxx-xxxx, and anything before that i put a + before and call it the international code.

it won't work for numbers that don't follow the (xxx) xxx-xxxx pattern, but i don't think our registration regex even allows these, and if it does then this problem gets really annoyingly hard without using an external library.

i hate regex, lmk what u think

hackathon_site/event/models.py
```diff
@@ -1,5 +1,6 @@
 from django.db import models
 from django.contrib.auth import get_user_model
+from django.core import validators
```
YuYing-Liang: this import isn't being used so you can remove it

hackathon_site/event/migrations/0006_profile_phone_number.py
```diff
@@ -0,0 +1,29 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="profile",
+            name="phone_number",
+            field=models.CharField(
+                default="8022818076",
+                max_length=20,
+                validators=[
```
YuYing-Liang: I think this migration is not up to date, could you remove the validators array (and the import) and set the default to an empty string so profiles don't accidentally have incorect phone numbers

hackathon_site/event/serializers.py
```diff
@@ -127,6 +130,7 @@ def create(self, validated_data):
             "id_provided": False,
             "acknowledge_rules": acknowledge_rules,
             "e_signature": e_signature,
+            "phone_number": Application.objects.get(user=current_user).phone_number,
```
YuYing-Liang, line 134: in the `read_only_fields` of `CurrentProfileSerializer`, could you add phone_number to that list? because we don't want to change the phone number

YuYing-Liang (commented):
Just some minor things luke but over all looks really good!

hackathon_site/dashboard/frontend/src/components/teamDetail/TeamInfoTable/TeamInfoTable.tsx
YuYing-Liang: honestly, great job on this 👏👏👏

YuYing-Liang (approved):
@lukecheseldine I messed up my database with the phone number migration so I need your code lol. I made & pushed the changes so I'm going to merge now

PR #435, opened 2022-09-21
Email notifs for status changes

YuYing-Liang:
## Overview

- sending notification emails to orders when they are created, updated and when items are returned.
- different emails for participant and admins

## Unit Tests Created

- none, will be created here at a later time: https://ieeeuoft.atlassian.net/browse/IEEE-225 

## Steps to QA

- if you create, update the status or return items from an order, this branch will email HSS admins (the email hardware@newhacks.ca) and all participants in the team corresponding to that order
- please test that the following works ^^ (you should see a log output file in the emails folder)
- look for any edge cases e.g. emails being sent when they aren't suppose to, the content of the email being incorrect (e.g. grammatical error)


Mustaballer (commented):
Awesome job! So far I tested adding an order as a participant, and I got the correct logs. I also was able to view the admin side links as admin. I tested returning orders, and I got the correct output. I tried testing changing status of an order to cancelled, it went through but I did get any email logs(See image below). 
image

Mustaballer:
I tested all the status changes and found no errors with the email logs. This is awesome!

Mustaballer (approved):

PR #448, opened 2022-12-21
fixed react warnings and added feature flags for rsvp and teams

YuYing-Liang:
## Overview

- Resolves: N/A
- fixed react warnings (the ones that appear on npm run build)
- added feature flags for teams and rsvp (this will help us when removing/adding features)
- added npm run build to ci checks to catch warnings on all branches

## Unit Tests Created

- edited tests that have to do with teams and rsvp

## Steps to QA

- disable teams in settings/__init__.py and ensure that anything to do with teams doesn't appear on localhost:8000
- don't need to test RSVP because it requires a migration, we'll test it out when we create the MakeUofT repo

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,3 +78,5 @@ jobs:
         run: yarn run tsc
       - name: Tests
         run: yarn test --watchAll=false
+      - name: Build frontend
+        run: yarn run build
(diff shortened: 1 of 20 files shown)


Mustaballer (approved):
Great job! I set Teams to false, and it removed all 'team' related content on localhost:8000 as expected. The flags definitely will make it easier to work with RSVP/teams if we want them or not.

PR #449, opened 2022-12-22
copied over new product overview features from newhacks and made tests

YuYing-Liang:
## Overview

- New Product Overview Features
- Hid the quantity available in the stock message for hardware items (copied from https://github.com/ieeeuoft/newhacks-site-2022/pull/26/files)

## Unit Tests Created

- Unit tests for new product overview features
- unit tests for new stock messages

## Steps to QA

- test on the admin side that the quantity available shows up but doesn't show up on the participant side
- test that categories and constraints labels don't appear when they don't exist for a hardware item


hackathon_site/dashboard/frontend/src/components/inventory/Item/Item.tsx
```diff
@@ -12,8 +14,11 @@ interface ItemProps {
 }
 
 const Item = ({ image, title, total, currentStock }: ItemProps): ReactElement => {
+    const userType = useSelector(userTypeSelector);
     const stock = !currentStock
         ? "OUT OF STOCK"
+        : userType === "participant"
+        ? currentStock + " IN STOCK"
```
Leo6Leo, line 21: CleanShot 2022-12-23 at 16 13 56@2x

The API will still return the maximum number of items in stock. Do we want to completely prevent participants from seeing that? Or it is OK we just leave like this

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
```diff
@@ -78,6 +85,29 @@ describe("<ProductOverview />", () => {
         expect(getByText("Add to cart")).toBeInTheDocument();
     });
 
+    test("Category label doesn't appear when there are no categories", async () => {
```
Leo6Leo, line 88: CleanShot 2022-12-23 at 16 23 51@2x

I believe if we import from Excel, this won't happen. But if we manually create the hardware item, it will not allow us to leave the category blank. We should change the database schema to make this field become optional.

hackathon_site/dashboard/frontend/src/testing/utils.tsx
```diff
@@ -131,6 +136,7 @@ export interface StoreEntities {
     pendingOrders?: OrderInTable[];
     teamDetailOrders?: OrderInTable[];
     teamDetailOrderState?: Partial<TeamOrderState>;
+    user?: Partial<UserState>;
```
Leo6Leo, line 158: What is Partial?

Leo6Leo (commented):
The manual test passes, and there is no other issues found.

hackathon_site/dashboard/frontend/src/components/inventory/Item/Item.tsx
YuYing-Liang: I think keeping the API response like that is fine because we need the data to show on the admin side. It's also not a big deal if they want to know the total number of items available, omitting it on the front-end is a UX choice.

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
YuYing-Liang: I'm not sure why this happens but the Django admin site puts a lot of fields as mandatory, even tho the database model says optional ,🤷🏻‍♀️ we can look into it more

hackathon_site/dashboard/frontend/src/testing/utils.tsx
YuYing-Liang: It's a partial state type because it means we don't need to input the full state (e.g. the error and loading properties) which gives us more flexibility when testing.

hackathon_site/dashboard/frontend/src/components/inventory/Item/Item.tsx
Leo6Leo: Sounds good!

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.test.tsx
Leo6Leo: I will create a ticket on Jira to keep track of it @YuYing-Liang

hackathon_site/dashboard/frontend/src/testing/utils.tsx
Leo6Leo: makes sense!

Leo6Leo (dismissed):
Looks good to me. I think it is ready to be merged.

Leo6Leo (approved):
No extra comments on this

PR #463, opened 2023-01-12
IEEE-241: Implement the order search function in the admin HSS

### LINKED ISSUE #241: Move the non-React yarn commands up from the event directory

daliaviki:
## Overview

- Resolves #IEEE 241
- Modified getOrdersWithFilters. 
- Redux thunk calls GET /hardware​/orders​/, and the parameter is the keyword from the search textbox.
- Hitting Enter on the keyboard, or clicking on the search button will both trigger the thunk.

## Unit Tests Created

Will be completed soon. 

## Steps to QA

Go to orders page and add orders with different teams and test functionality. Search function queries team code and order id. Please test both.


daliaviki:
@YuYing-Liang What unit tests do you suggest I make?

hackathon_site/dashboard/frontend/src/slices/order/adminOrderSlice.ts
```diff
@@ -46,7 +46,7 @@ interface RejectValue {
 
 export const getOrdersWithFilters = createAsyncThunk<
     APIListResponse<Order>,
-    void,
+    { keepOld?: boolean } | undefined,
```
YuYing-Liang: we don't need keepOld in this case, that was added to the hardware thunk because it required more flexibility. So you can remove this.

hackathon_site/dashboard/frontend/src/slices/order/adminOrderSlice.ts
```diff
@@ -92,19 +92,29 @@ const adminOrderSlice = createSlice({
...
+            }
         },
     },
     extraReducers: (builder) => {
         builder.addCase(getOrdersWithFilters.pending, (state) => {
             state.isLoading = true;
             state.error = null;
         });
-        builder.addCase(getOrdersWithFilters.fulfilled, (state, { payload }) => {
+        builder.addCase(getOrdersWithFilters.fulfilled, (state, { payload, meta }) => {
```
YuYing-Liang, line 118: again, we don't need keepOld so you can remove any instances of `meta`

YuYing-Liang (commented):
> @YuYing-Liang What unit tests do you suggest I make?

Before writing any new tests, make sure that current tests are all passing (you only have 1 that failed so should be a quick fix). You need to test essentially every functionality you've written in this PR. That includes:
- testing the order search bar (make sure that entering an input works, clearing an input, submitting an input with search button and clicking enter)
- you also have to test that you're calling our APIs properly, thus you will need to make a new test/alter a test in adminOrderSlice.test.ts for the thunk you modified

There are some other tests I haven't mentioned but think about what you've written and you'll probably know what to test. You may also want to peek at the Inventory Search Bar and Team Search Bar components for * inspiration * of what kinds of tests to write.

So far great job though!!

hackathon_site/dashboard/frontend/src/slices/order/adminOrderSlice.ts
YuYing-Liang: bumping this comment again

hackathon_site/dashboard/frontend/src/pages/Orders/Orders.tsx
```diff
@@ -27,6 +28,7 @@ const Orders = () => {
     };
 
     useEffect(() => {
+        dispatch(clearFilters());
```
YuYing-Liang, line 31: hmm I'm not so sure we want to clear filters every time we load the page. It might be nice for the user to go back to exactly the same orders they were looking at before, but totally your call.

lukecheseldine (approved):
looks good Dalia, great job! feel free to merge :)

lukecheseldine:
merging this myself since we need this feature in develop

PR #465, opened 2023-01-14
Qr code scanner

YuYing-Liang:
## Overview

- Resolves #<number>
- <Description of all changes made and features implemented>

## Unit Tests Created

## Steps to QA


hackathon_site/event/admin.py
```diff
@@ -96,5 +96,28 @@ def get_members_count(self, obj):
         return obj.members_count
 
 
+@admin.register(UserActivity)
+class UserActivityAdmin(ExportMixin, admin.ModelAdmin):
+    list_display = (
+        "get_user_name",
+        "sign_in",
+        "lunch1",
+        "dinner1",
```
Leo6Leo, line 105: I think we can make this as dynamic by setting it to the variable in the setting file.

hackathon_site/event/jinja2/event/admin_qr_scanner.html
```diff
@@ -0,0 +1,127 @@
...
+
+{% block body %}
+<div class="backgroundColorDashboard">
+    <div class="container">
+        <div class="section">
+            <div class="borderTopDiv z-depth-3">
+
+                {% if get_messages(request) %}
+                    {% for message in get_messages(request) %}
+                        <p id="submitMessage" class="banner banner{{ message.tags }}"> {{ message }} </p>
```
Leo6Leo, line 16: I think we need to change the design a bit. It is not clear right now.
CleanShot 2023-02-03 at 17 26 49@2x

hackathon_site/event/jinja2/event/admin_qr_scanner.html
```diff
@@ -0,0 +1,127 @@
...
+
+                <table>
+                    <tr>
+                        <th>Event</th>
+                        <th>Time</th>
+                        <th>Sign In Interval</th>
+                    </tr>
+                    {% for event in sign_in_times %}
+                        <tr>
+                            <td> {{ event.description }} </td>
```
Leo6Leo, line 79: It would be more intuitive to show which event is happening right now by changing the color of the column or add one more column called status, and show the emoji to indicate the status of the event.

hackathon_site/event/jinja2/event/admin_qr_scanner.html
```diff
@@ -0,0 +1,127 @@
...
+    <script src="{{ static("event/js/qr-scanner.umd.min.js") }}"></script>
+    <script type="text/javascript">
+        const videoElem = document.getElementById("scanner");
+        let oldData = '';
+
+        $("#signInButton").click(() => {
+            console.log(oldData.split(';'))
+        });
+
+        const qrScanner = new QrScanner(
```
Leo6Leo, line 101: Are we able to have a button to disable the camera / QR scanner if we want to use the real scanner instead of the camera.

hackathon_site/event/jinja2/event/admin_qr_scanner.html
```diff
@@ -0,0 +1,127 @@
+{% extends "event/base.html" %}
+
+{% block nav_links %}
+<li><a href="{{ url("event:dashboard") }}" class="active">Dashboard</a></li>
```
Leo6Leo, line 4: Remove the class=active

hackathon_site/event/jinja2/event/dashboard_admin.html
```diff
@@ -0,0 +1,66 @@
...
+                            type="submit">Sign-In Data</button>
+                </div>
+            </div>
+
+            <div class="borderTopDiv z-depth-3">
+                <h1 class="formH1">Analytics</h1>
+                <table>
+                    <tr>
+                        <td>Total number of sign-ups:</td>
+                        <td>345</td>
```
Leo6Leo, line 51: Just don't forget to use the dynamic data.

hackathon_site/event/jinja2/event/dashboard_admin.html
```diff
@@ -0,0 +1,66 @@
...
+            <div class="borderTopDiv z-depth-3">
+                <h1 class="formH1">Admin Dashboard</h1>
+
+                <h2 class="formH2">Administrative Actions</h2>
+                <div class="btn-group-sm">
+                    <a href="{{ url("event:qr-scanner") }}" class="btn-sm btn-small waves-effect waves-light colorBtn">
+                        Scan QR Code
+                    </a>
+                    <a href="" class="btn-sm btn-small waves-effect waves-light colorBtn">
+                        Review Applications
```
Leo6Leo, line 21: TODO: Link waiting to be updated

hackathon_site/event/jinja2/event/dashboard_admin.html
```diff
@@ -0,0 +1,66 @@
...
+                    </a>
+                    <a href="" class="btn-sm btn-small waves-effect waves-light colorBtn">
+                        View Applications
```
Leo6Leo, line 24: TODO: Link waiting to be updated

hackathon_site/event/jinja2/event/dashboard_base.html
```diff
@@ -226,4 +230,15 @@ <h2 class="formH2" id="faq">Application FAQs</h2>
...
     </div>
 </div>
+{% endblock %}
+
+{% block scripts %}
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/qrious/4.0.2/qrious.min.js"></script>
+    <script type="text/javascript">
+        var qrcode = new QRious({
+            element: document.getElementById("qrcode"),
+            value: "{{ user.first_name }};{{ user.last_name }};{{ user.email }}",
```
Leo6Leo, line 240: I think it is unsafe to do so. Here we are exposing user's first name, last name, and their email directly in the QR code. I would recommend we encrypt it?

hackathon_site/event/models.py
```diff
@@ -44,3 +44,12 @@ def save(self, *args, **kwargs):
 
     def __str__(self):
         return f"{self.id} | {self.user.first_name} {self.user.last_name}"
+
+
+class UserActivity(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    sign_in = models.DateTimeField(null=True)
+    lunch1 = models.DateTimeField(null=True)
```
Leo6Leo, line 52: Any ways to make this dynamic? My suggestion on this design would be: We have the table called UserActivity, and the attribute for the table will be user, created_time, event(foreign key to the table Event Arrangement), scanner(Foreign key to the table User). And another table called event, storing all the events that the admin created. And it will potentially have these attributes: event name, event id, start_time, end_time, discord_bot_notify(boolean, if it is true, discord bot will send an announcement in the channel), and more....

hackathon_site/event/views.py
```diff
@@ -188,6 +197,66 @@ def post(self, request, *args, **kwargs):
         return super().post(request, *args, **kwargs)
 
 
+class QRScannerView(LoginRequiredMixin, FormView):
+    success_url = reverse_lazy("event:qr-scanner")
+
+    def get_template_names(self):
+        if self.request.user.is_staff:
+            return "event/admin_qr_scanner.html"
+        return Exception("You do not have permission to view this page.")
```
Leo6Leo, line 206: seems like exception is not working properly.
CleanShot 2023-02-03 at 17 45 03@2x

hackathon_site/event/views.py
```diff
@@ -188,6 +197,66 @@ def post(self, request, *args, **kwargs):
...
+    def form_valid(self, form):
+        if isinstance(form, SignInForm):
+            try:
+                user = User.objects.get(email__exact=form.cleaned_data["email"])
+                sign_in_event = get_curr_sign_in_time()
+                now = datetime.now().replace(tzinfo=settings.TZ_INFO)
+
+                try:
+                    user_activity = UserActivity.objects.get(user__exact=user)
+                    setattr(user_activity, sign_in_event, now)
```
Leo6Leo, line 234: we need to add the check that if the user has already checked in

hackathon_site/event/jinja2/event/dashboard_admin.html
```diff
@@ -0,0 +1,66 @@
...
+                    </a>
+                </div>
+
+                <br/>
+
+                <h2 class="formH2">Export Data to Google Sheets</h2>
```
Leo6Leo, line 30: We should hide these to all our volunteers. We don't want everyone to see our sensitive data.

hackathon_site/event/jinja2/event/dashboard_admin.html
Leo6Leo: We should hide these to all our volunteers. We don't want everyone to see our sensitive data.

hackathon_site/event/jinja2/event/dashboard_admin.html
Leo6Leo: We should hide these to all our volunteers. We don't want everyone to see our sensitive data.

hackathon_site/event/jinja2/event/dashboard_admin.html
```diff
@@ -0,0 +1,66 @@
...
+                            type="submit">Application Data</button>
+                    <button disabled class="btn-sm btn-small waves-effect waves-light colorBtn"
+                            type="submit">Review Data</button>
+                    <button disabled class="btn-sm btn-small waves-effect waves-light colorBtn"
+                            type="submit">Sign-In Data</button>
+                </div>
+            </div>
+
+            <div class="borderTopDiv z-depth-3">
+                <h1 class="formH1">Analytics</h1>
```
Leo6Leo, line 47: We should hide these to all our volunteers. We don't want everyone to see our sensitive data.

hackathon_site/registration/forms.py
```diff
@@ -211,3 +212,50 @@ def clean_team_code(self):
...
+        email = self.cleaned_data["email"]
+
+        try:
+            user = User.objects.get(email__exact=email)
+            application = Application.objects.get(user__exact=user)
+            review = Review.objects.get(application__exact=application)
+            if review.status == "Accepted":
+                if settings.RSVP and application.rsvp is None:
+                    raise forms.ValidationError(
+                        _(f"User {email} has not RSVP'd to the hackathon")
```
Leo6Leo, line 244: And one more thing is that we should disable user's ability to cancel RSVP when the hackathon has started. Otherwise, that will cause many issues.

Leo6Leo (changes requested):
Please see the comments above

hackathon_site/event/jinja2/event/admin_qr_scanner.html
YuYing-Liang: @Leo6Leo you need to compile the scss, run `yarn run scss` in the hackthon_site directory and `python manage.py collectstatic`

hackathon_site/event/jinja2/event/admin_qr_scanner.html
YuYing-Liang: Could you explain this more? What is a "real scanner" I thought QR codes can only be scanned through camera/images.

hackathon_site/event/jinja2/event/dashboard_admin.html
YuYing-Liang: I'll disable those buttons for now (since reviews will be over at that point) and we can figure out later all the volunteer permissions.

hackathon_site/event/jinja2/event/dashboard_admin.html
YuYing-Liang: same comment as above.

hackathon_site/event/jinja2/event/dashboard_base.html
YuYing-Liang: So for newhacks I think data was encoded the same way in the qr code. But also, I don't think this is unsafe because only the participant will have their qr code and first/last name and email is not sensitive data (in my opinion but let me know why you think it is). I don't see how people's information will be exposed in any way. Also encryption on the frontend is not super useful because ppl can find out the encryption algorithm and decrypt it.

hackathon_site/event/models.py
YuYing-Liang: Yeah, I'll think about it. My main idea was because we will have a known finite number of events so in the future we can set it up just like how we set up the application model questions, but if you want to extend it to further uses I can change it.

hackathon_site/event/views.py
YuYing-Liang: No that's right, that's the exception page. But now that I'm thinking about it, I should try to return a 401 unauthorized error (or 403). I didn't want to make a whole new html page for an error so I just went with an Exception.

hackathon_site/event/views.py
YuYing-Liang: If the user has already signed-in it'll just update the time in the table. I didn't see a problem with that so I left it this way.

hackathon_site/event/jinja2/event/dashboard_admin.html
YuYing-Liang: Sure, they're dsiabled for now

hackathon_site/event/jinja2/event/dashboard_admin.html
YuYing-Liang: smae as comment above

hackathon_site/registration/forms.py
YuYing-Liang: That is already implemented in the RSVP feature, but I'll leave this validation check in.

PR #466, opened 2023-02-01
added validation for review decision sent_date

YuYing-Liang:
## Overview

- added a check that only allows users to create profiles after acceptances have been sent out
- fix rsvp bug

## Unit Tests Created

- edited user review and create profile serializers

## Steps to QA

- apply as a (new) user
- accept user to hackathon (but not change decision-sent-date)
- go to HSS, you should be unable to create a profile/view acknowledgements page

### DIFF
--- a/hackathon_site/registration/views.py
+++ b/hackathon_site/registration/views.py
@@ -260,8 +260,9 @@ class RSVPView(LoginRequiredMixin, View):
                 application.rsvp = True
                 application.save()
 
-                profile = Profile(user=user, team=EventTeam.objects.create())
-                profile.save()
+                # TODO: decide whether to create profile when rsvp
+                # profile = Profile(user=user, team=EventTeam.objects.create())
+                # profile.save()
 
             elif decision == "no" and (application.rsvp or application.rsvp is None):
                 application.rsvp = False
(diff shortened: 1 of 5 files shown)


hackathon_site/registration/views.py
```diff
@@ -260,8 +260,9 @@ def get(self, request, *args, **kwargs):
                 application.rsvp = True
                 application.save()
 
-                profile = Profile(user=user, team=EventTeam.objects.create())
-                profile.save()
+                # TODO: decide whether to create profile when rsvp
+                # profile = Profile(user=user, team=EventTeam.objects.create())
+                # profile.save()
```
Leo6Leo, line 265: Any reason for not creating the profile at HSS?

Leo6Leo (approved):
Tested with normal use case and the corner case, it works perfectly on my end.  I left a question, see it before merge the PR.

hackathon_site/registration/views.py
YuYing-Liang: Right now, if you create a profile on RSVP, it doesn't prompt the acknowledgements (we can fix it but that would be more effort than just commenting this out).

PR #467, opened 2023-02-01
added check in order serializer

YuYing-Liang:


Leo6Leo (approved):
Tested, and code checked. No problems found! Great work Yuying!

PR #470, opened 2023-02-13
Ieee 248 datasheet link still appears

terryluan12:


YuYing-Liang:
I think https://github.com/ieeeuoft/hackathon-template/pull/471 already fixes this

PR #471, opened 2023-02-13
Ieee 249 negative stock number

terryluan12:
## Overview

- Resolves #IEE-249
- Removed negative numbered possibilities in stock


hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
```diff
@@ -230,18 +230,29 @@ export const DetailInfoSection = ({
                 Manufacturer
             </Typography>
             <Typography>{manufacturer}</Typography>
-            <Typography variant="body2" className={styles.heading}>
+            <Typography
```
YuYing-Liang: I think it would be better to take advantage of jsx here
```
{
   modelNumber && <>
      <Typography> ... </Typography>
      <Typography> ... </Typography>
   </>
}
```
same for the datasheet prop

YuYing-Liang (commented):
Looks good, try not to use css when you can use javascript though, its more readable and efficient!

hackathon_site/dashboard/frontend/src/components/inventory/ProductOverview/ProductOverview.tsx
terryluan12: Sounds good! Just made the change. Thanks YuYing!

YuYing-Liang (dismissed):

hackathon_site/dashboard/frontend/src/components/inventory/Item/Item.tsx
```diff
@@ -15,11 +15,12 @@ interface ItemProps {
 
 const Item = ({ image, title, total, currentStock }: ItemProps): ReactElement => {
     const userType = useSelector(userTypeSelector);
-    const stock = !currentStock
-        ? "OUT OF STOCK"
-        : userType === "participant"
-        ? currentStock + " IN STOCK"
-        : currentStock + " OF " + total + " IN STOCK";
+    const stock =
+        !currentStock || currentStock < 0
```
lisa-sa-li:
```suggestion
currentStock < 1
```

lisa-sa-li: I think the `!currentStock` existed because we didn't have typing before and the value could have been null?

lisa-sa-li (commented):
@terryluan12 just one minor comment

lisa-sa-li (approved):
lol sorry made the fix myself and approved it cause we need it rn

PR #472, opened 2023-02-14
updated hardware model to accept blank/null fields

YuYing-Liang:
QA Steps:
- run migration
- go to admin site and you should be able to change most fields in your hardware to be empty and save successfully


Mustaballer (approved):
Same as MakeUofT PR

Leo6Leo (approved):
Tested, no problem

PR #485, opened 2023-05-28
Starting work on the Pending Orders cards on the Admin Dashboard

jhimanish:
## Overview

- Resolves 256 Create Pending Order Section on the HSS Admin Dashboard
- Creating pending orders section on the hardware admin dashboard page
- Using MUI grid to layout the OrderCards for pending orders

## Unit Tests Created

- Created AdminDashboard.test.tsx that checks for titles in the Admin dashboard

## Steps to QA

-

## View of Final Page

image

The page only shows up to 6 pending order cards.


Mustaballer:
Hey @jhimanish  there seems to be some typescript errors that the Github Actions caught up on (I suspect its related to files you didn't stage which is fine).
image
 Thomas mentioned that you were having issues with CSS. I suspect its because you didn't compile the scss to css.
 
 Run the following command(more info is found in the README) -->
 `$ yarn run scss`
 
 Optional: To compile the SCSS automatically when you save, run following task running while you work:
```
$ cd hackathon_site
$ yarn run scss-watch
```

jhimanish:
Hmm, I thought I had compiled it correctly when I set it up the first time. I may have forgotten to set up the scss-watch when I initially set up the repo. Thanks for letting me know, I'll try that.

Mustaballer:
To use redux in this example its actually quite convenient that we have the functionality set up to obtain pending orders(pendingOrderSelectors). 
The following code retreives all the pending orders. You can use this instead of the mock data you currently are using, and make changes wherever needed. You'll need to create pending orders either through the Django Admin Site, or as a participant user.
```const orders = useSelector(pendingOrderSelectors.selectAll);```

If you have any questions, let me know :)

**Comment by jhimanish** [comment]
Ahh okay I'll remove the mock data and try that out then. Thanks!

**Comment by Mustaballer** [comment]
Hey Himanish if you need help or suggestions to get started, let me know I'd be happy to help :)

File: hackathon_site/dashboard/frontend/src/components/orders/OrderCard/OrderCard.tsx [hunk 1]
```diff
@@ -14,7 +14,10 @@ const OrderCard = ({ teamCode, orderQuantity, time, id, status }: OrderProps) =>
     const date = new Date(time);
     const month = date.toLocaleString("default", { month: "short" });
     const day = date.getDate();
-    const hoursAndMinutes = date.getHours() + ":" + date.getMinutes();
+    const hoursAndMinutes =
```
Mustaballer, line 17: I like the extra digit formatting :)

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.module.scss
```diff
@@ -0,0 +1,24 @@
+@import "./../../assets/abstracts/mixins";
```
Mustaballer: You should use absolute imports with src as the reference point 
image
For this example it would be `@import "assets/abstracts/mixins"`

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import AdminDashboard from "pages/AdminDashboard/AdminDashboard";
+
+import { render, waitFor } from "testing/utils";
+
+describe("Admin Dashboard Page", () => {
```
Mustaballer, line 9: Can you make a test that checks for Pending order Cards similar to how `Orders.test.tsx` checks for orders. Refer to that file to see how they implement the store
```typescript
test("Display correct order cards", () => {
        const { getByTestId } = render(<Orders />, { store });
        mockPendingOrders.forEach((order) => {
            const orderStatus = order.status;

            if (orderStatus === "Submitted" || orderStatus === "Ready for Pickup") {
                const orderItem = getByTestId(`order-item-${order.id}`);
                const orderDetails = within(orderItem);
                const date = new Date(order.created_at);
                const month = date.toLocaleString("default", { month: "short" });
                const day = date.getDate();
                const hoursAndMinutes = date.getHours() + ":" + date.getMinutes();

                expect(orderDetails.getByText(order.id)).toBeInTheDocument();
                expect(orderDetails.getByText(order.items.length)).toBeInTheDocument();
                expect(orderDetails.getByText(order.team_code)).toBeInTheDocument();
                expect(
                    orderDetails.getByText(`${month} ${day}, ${hoursAndMinutes}`)
                ).toBeInTheDocument();
            }
```

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,69 @@ import React from "react";
...
+import OrderCard from "components/orders/OrderCard/OrderCard";
+import { useDispatch, useSelector } from "react-redux";
+import styles from "./AdminDashboard.module.scss";
+import Grid from "@material-ui/core/Grid";
+import Button from "@material-ui/core/Button";
+import {
+    adminOrderSelectors,
+    getOrdersWithFilters,
+    setFilters,
+} from "../../slices/order/adminOrderSlice";
```
Mustaballer: use absolute import

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,69 @@ import React from "react";
...
+import { OrderFilters } from "../../api/types";
```
Mustaballer: use absolute import

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,69 @@ import React from "react";
...
 
 const AdminDashboard = () => {
+    const dispatch = useDispatch();
+    const orders = useSelector(adminOrderSelectors.selectAll);
+    const pendingFilter: OrderFilters = {
+        status: ["Submitted"],
```
Mustaballer: Is this the only status for pending orders? `Ready for Pickup` I believe is also a pending order

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
```diff
@@ -2,12 +2,69 @@ import React from "react";
...
+    };
+    dispatch(setFilters(pendingFilter));
+    dispatch(getOrdersWithFilters());
+    const ordersLength = orders.length < 7 ? orders.length : 6;
```
Mustaballer: Does this order length cover the whole Size of the page? Ideally it should be as close to the Figma design and I believe that essentially covers the width of the webpage. Please use a variable for the size(for `7` or `6`), we typically don't want to be using numeric literals.

Mustaballer (changes requested):
Overall, great work Himanish! I've left some minor comments, but this is looking close to ready for merge. :)

hackathon_site/dashboard/frontend/src/components/orders/OrderCard/OrderCard.tsx
jhimanish: thanks!

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.module.scss
jhimanish: Sounds good, thanks, fixed

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
jhimanish: Fixed

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
jhimanish: Fixed

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
jhimanish: Ah, yes you're right, I've fixed that as well, thanks

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.tsx
jhimanish: Yes, this does cover the whole size of the page. I've defined the grid items to be 2 for large:
image
So a total of 6 items would fit in the 12 size total. 

This is what it looks like with a max of 6:
image

I think if we try adding another one, each card would be too small. So i think this is a good amount. I'll make a constant instead of the numeric literal

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.test.tsx
jhimanish: I added the test similar to orders.test.tsx. This test uses getByTestId and the data-testid attribute so I implemented it into the admindashboard:
image

But it wouldn't work when I had `id={order.team_id}`. I had to change it to `id={order.id}` for the data-testid to show up. 

What I mean is that when I had the following:

```
id={order.team_id}
status={order.status}
data-testid={`order-item-${order.id}`}
```
The HTML would display `data-testid="order-item-2"` which was the order.team_id rather than the order.id

When I changed it to
```
id={order.id}
status={order.status}
data-testid={`order-item-${order.id}`}
```
Then, it worked properly and the HTML displayed `data-testid="order-item-3"` which is the correct number (order.id). I'm not sure why this happens. Do you know what the issue could be?

Mustaballer: In general, it is not recommended to render components with the same ID. The initial issue might have been caused by rendering multiple components with the same ID, especially considering that teams can have multiple order cards. The second approach would ensure its a unique ID.

Mustaballer:
Great job @jhimanish! One last thing can you add an onClick event when the admin user clicks on the order card it should redirect them to the `/teams/${order.team_code}` page.

See how its done in Orders.tsx:
```typescript
<Grid
    item
    lg={3}
    md={4}
    sm={6}
    xs={12}
    key={idx}
    onClick={() =>
        history.push(`/teams/${order.team_code}`)
    }
</Grid>
```

hackathon_site/dashboard/frontend/src/pages/AdminDashboard/AdminDashboard.test.tsx
jhimanish: Hmm, interesting okay. Yeah I think this way of displaying the ID as the order.id rather than the team_id works better anyways. We already display the team_code so we should display the order id not the team_id.

Mustaballer (approved):

danielqiuu (approved):

PR #486, opened 2023-05-28
[IEEE-245]: Get Review Status By Username

liu-samuel:
## Overview

- Resolves IEEE-245
- Added a new endpoint to get a user's review status based on username

## Unit Tests Created

- User Review Status Unit Tests in test_api.py

## Steps to QA

- N/A


LunarFang416 (dismissed):
LGTM

hackathon_site/event/api_views.py
```diff
@@ -62,6 +62,24 @@ def get(self, request, *args, **kwargs):
         return self.retrieve(request, *args, **kwargs)
 
 
+class UserReviewStatusAPIView(generics.GenericAPIView, mixins.RetrieveModelMixin):
+    """
+    View to handle review status status of a user with a certain username
```
lukecheseldine:
```suggestion
    View to handle review status of a user with a certain username
```

lukecheseldine (changes requested):
great work Sam! this is almost ready, I'd just like you to add two more test cases.
1. the api is hit with an email that doesn't exist (return 400)
2. the api is hit without being logged in (return 401)
in general, we should try to write tests for all possible return codes. write these up and should be good to go!

hackathon_site/event/api_views.py
liu-samuel: Changes are pushed

lukecheseldine (approved):
looks good, thanks for making those changes :)

LunarFang416 (approved):
👍👍

PR #489, opened 2023-05-29
(IEEE-196): Delete Team endpoint added

LunarFang416:
## Overview

- Resolves IEEE-196
- DELETE: /api/event/teams/{team_code} endpoint added; now you can delete teams
- Teams with open orders cannot be deleted

## Unit Tests Created

- Unit tests cover simple and edge cases for team deletion

## Steps to QA

- Test the endpoint on swagger


lukecheseldine:
reallyyy great work!!! check your DMs and see https://ieeeuoft.atlassian.net/browse/IEEE-265. we're going to let this sit in review until that other ticket is completed

hackathon_site/event/signals.py
```diff
@@ -1,10 +1,21 @@
...
 
 @receiver(post_delete, sender=Profile, dispatch_uid="profile_delete_signal")
 def delete_profile(sender, instance, **kwargs):
-    team = instance.team
+    """
+    When Teams are deleted, Profiles trigger a post_delete signal due to the
+    on_delete=CASCADE rule. This may cause errors here when Profiles try
+    to delete a team which has already been deleted due to a previous Profile.
+    Only the first Profile will not raise a Team.DoesNotExist exception and
+    delete the team properly, remaining Profiles do not need to do so
```
lukecheseldine, line 13: great catch!

lukecheseldine (changes requested):
Tested and it works exactly as it should! Great work :)

When I deleted a team though, it deleted all of the profiles, users, and orders associated with that team due to the on_cascade rule you mentioned. This doesn't totally make sense, but at the same time, if we delete the team and not the profiles / users, I don't know how the HSS will behave. I assume the users / profiles not linked to a team will be stranded and unable to do anything @Mustaballer any idea or insight on this?

Anyway, I think not deleting the users / profiles introduces a whole new UI / UX flow where they have to be taken to a new screen to join an existing team, or automatically assigned new teams... not sure how we want to handle this, so I think for now we can just keep the cascade there. 

For the orders, I _would_ like for them to be kept. It's useful to have statistics on total orders, most popular items... so let's keep them even after team deletion. Can we change the deletion behavior just for orders @LunarFang416 so they are kept?

lukecheseldine (approved):
amazing work, thanks for bearing with all my last minute requests! let @liu-samuel review and then it's good to merge!

liu-samuel (approved):
lgtm

PR #491, opened 2023-05-30
IEEE 253 fixing order modal cancel bug

carmen-chau:
## Overview

Description: This PR (for Jira ticket IEEE-253) contains fixes to the order modal cancel bug.

Bug description: Originally, the orders would be canceled chronologically. We don't want that. We want the correct orders to be canceled (corresponding to which cancel button for which order the user presses)

Speculation as to why the bug occurred: The reason why orders were being canceled in chronological order is that in the submitModal() method, the orderID being retrieved is always the most recent orderID (which as a result, would cancel the most recent order, not the order in which the participant originally wanted). I

Summary of changes: Created a new variable called orderId that stores the desired canceled order's id based on the "Cancel order" button press. The state of this variable is updated each time the button is being pressed and should work even when multiple orders are pending. Note: Implementations using event.stopPropagation() did not work, hence why I resorted to this method of change.

## Unit Tests Created

- Did not create nor modify any unit tests

## Steps to QA

- Would appreciate suggestions to improve or rewrite current solution based on best practices. Would also appreciate testing to see whether bug is fully resolved with this solution and/or other ideas as to why the bug originated in the first place.


Mustaballer (changes requested):
It seems like the issue arises due to event bubbling, causing the most recent orderID to always be retrieved. event.stopPropagation() could be a potential solution as it would prevent the onClick event from propagating to outer elements or handlers. Could you look into using that instead of the state variables?

Mustaballer (changes requested):
Could you try using event.stopPropagation() in ItemTables.tsx or whichever file you call the PupupModal when the cancel order button is clicked?

Mustaballer:
Hey @carmen-chau if the `event.propogation`, doesn't work we can try your initial method that does delete order items properly.

danielqiuu (dismissed):

lukecheseldine (dismissed):
LGTM, great job Carmen!

hackathon_site/dashboard/frontend/src/components/dashboard/ItemTable/ItemTable.tsx
```diff
@@ -230,6 +230,7 @@ export const PendingTables = () => {
     const toggleVisibility = () => dispatch(togglePendingTable());
     const cancelOrder = (orderId: number) => dispatch(cancelOrderThunk(orderId));
     const [showCancelOrderModal, setShowCancelOrderModal] = useState(false);
+    const [orderId, setorderId] = useState(0);
```
Mustaballer: Can you replace the `0` with `null`, because setting it to 0 indicates it's given an id, which by default it should not be.

Mustaballer (changes requested):
Had one small comment, besides that looks great

hackathon_site/dashboard/frontend/src/components/general/PopupModal/PopupModal.tsx
```diff
@@ -39,7 +39,7 @@ const PopupModal = ({
                 </Grid>
                 <Grid item container justifyContent={"flex-end"}>
                     <Button onClick={cancelHandler}>{cancelText ?? "Cancel"}</Button>
-                    <Button onClick={submitHandler} color="primary">
+                    <Button id="submitButton" onClick={submitHandler} color="primary">
```
Mustaballer: is there a reason why you gave the Button a `submitButton` id?

Mustaballer (changes requested):
Sorry one last thing, can you change the names of `submitModal` and `setModal`? I think its a bit too general since its being used for cancelling orders and should have a name about that. Thanks :)

danielqiuu (approved):

Mustaballer (approved):
Great work!

Mustaballer:
Please also create a PR in newhacks-site-2023 repo as well :)

PR #492, opened 2023-06-01
(IEEE-265): Add 'Returned' status to orders

LunarFang416:
## Overview

- Resolves IEEE-265
- After all individual hardware items for an order are returned, the status of an order is changes from 'Picked Up' to 'Returned'

## Unit Tests Created

- No new tests
- Updated existing tests to account for new 'Returned' status. Since 'Returned' orders aren't active orders, they are treated like Cancelled orders, hence you can leave a team if you have returned or cancelled orders only

## Steps to QA

- Once all hardware items for an order are returned, you can filter orders by 'Returned' and view all orders with that status


lukecheseldine (approved):
Looks great Abubukker! Only comment is to please name your PR according to the same naming scheme we use for branches like IEEE-123: Did something cool.

Please hold off from merging until @liu-samuel can review it, but this is great work man :)

liu-samuel (approved):
Everything looks good!

PR #501, opened 2023-06-14
Ieee 255 overview section

phamalis:
## Overview

- Resolves #<number>
- <Description of all changes made and features implemented>

## Unit Tests Created

## Steps to QA


Mustaballer (changes requested):
Hey @phamalis we don't want to be deleting the yarn.lock file since it already exists, can you undo deleting it and make sure its the same as before(if it is, git will not include in this PR). Please do the same for the package.json --> we should not be commiting that file at all in this PR. 

You can schedule some time with me this week and we can go over the cool stuff --> using real data instead with redux.

Mustaballer:
Closed in favour of https://github.com/ieeeuoft/hackathon-template/pull/504

PR #503, opened 2023-06-18
Ieee 272 implement item incident form

natapokie:
## Overview

- Resolves [272](https://ieeeuoft.atlassian.net/browse/IEEE-272)

- Created incident form with form validation 
  - displays error messages on invalid submission
  - logs the values on correct submission (later replaced with the actual backend stuff)
    - upon successful submission resets the form
- back button takes you back to the previous page 
- broken link button on checked out tables will navigate user to the incident form page

**Some General Questions...**
- Qty seemed like it was a dropdown, so I just put an arbitrary selection of numbers 1-8 (not sure if there is a specific number?)
- Is the link implementation to the incident form ok? i worked off of a file that was already in the code base
- What should users be submitting in the text fields (what, where when)? Is there an example we can provide as a placeholder instead and use the current placeholder (e.g., "What happened to the hardware?") as a label instead?
- How to handle the submit? for example, display an success message, or disable the submit button after?

**Preview...**

broken link in `CheckedOutTables component` (uncommented previous code),
image

incident form
image

invalid submission
image

## Unit Tests Created

- Unit test ensures that components are rendered properly, `yarn test IncidentForm.test.tsx`
  - note: didn't make any tests to emulate submission since formik in the frontend takes care of it...

## Steps to QA

- navigate to `/incident-form`
- in order to submit, all fields must be entered
- an field that is empty will prompt an error message
- pressing submit will print all values to the console


Mustaballer:
> **Some General Questions...**
> 
> * Qty seemed like it was a dropdown, so I just put an arbitrary selection of numbers 1-8 (not sure if there is a specific number?)
> * Is the link implementation to the incident form ok? i worked off of a file that was already in the code base
> * What should users be submitting in the text fields (what, where when)? Is there an example we can provide as a placeholder instead and use the current placeholder (e.g., "What happened to the hardware?") as a label instead?
> * How to handle the submit? for example, display an success message, or disable the submit button after?
1. For quantity dropdown the number would range from 1 to the total quantity checked out for that hardware.
2. Will need to test(will do soon)
3. For submitting I was thinking of like a redirect to the dashboard page with a success snackbar. You're free to do what you feel makes most sense.

natapokie:
@Mustaballer thanks for clarifying !!

in the link to the incident page, I used query parameters(?) to pass in data -- just wanted to provide this as one of the options because I realize now that having a page for the incident form might not be the best idea since you probably want to pass in the team info too...
currently, i just added the hardware quantity to the link, so the dropdown will show different numbers!

i do also like the idea of a snackbar so i added that in my recent commit,

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,255 @@
...
+                                                                    }
+                                                                    label="Lost"
+                                                                />
+                                                                <FormControlLabel
+                                                                    value="other"
+                                                                    control={
+                                                                        <Radio color="primary" />
+                                                                    }
+                                                                    label="Other"
+                                                                />
```
KarandeepLubana: This can be refactored using the `map()` function which can be used to iterate over an array and generate the corresponding `FormControlLabel `components.

Helpful links:
https://legacy.reactjs.org/docs/lists-and-keys.html 
https://www.pluralsight.com/guides/how-to-use-the-map()-function-to-export-javascript-in-react

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,255 @@
...
+                                                name="where"
+                                                placeholder="Where did this occur?"
+                                                value={values.where}
+                                                error={!!errors?.where}
+                                                helperText={errors?.where}
+                                                onChange={handleChange}
+                                                variant="outlined"
+                                                type="text"
+                                                multiline
+                                            />
```
KarandeepLubana: This can also be refactored using the map() function. Refer to previous comment

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,255 @@
...
+                                                                />
+                                                            </RadioGroup>
+                                                        </>
+                                                    )}
+                                                </Field>
+                                                <FormHelperText>
+                                                    {errors?.state}
+                                                </FormHelperText>
+                                            </FormControl>
+                                            <br></br>
```
KarandeepLubana: Using `<br>` tags to create space between elements is not good practice. 

For virtually anything related to layout, you should use CSS instead. So if you want to create more space between blocks of text or other things, instead of using the `<br>` tag, you should use CSS margin or padding.

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,255 @@
...
+    const initialValues = {
+        state: "",
+        qty: "",
+        what: "",
+        when: "",
+        where: "",
+    };
+
+    const handleSubmit = async (values: FormikValues, { resetForm }: any) => {
+        console.log("submit");
```
KarandeepLubana: Remove all console logs

KarandeepLubana:
image

Empty values are considered invalid. However, you should not display an error in the very beginning, as it is a bad UX (the user clicks a button and sees various errors before having made any action). You should validate a field after its value has changed, or once the submit button is clicked

KarandeepLubana:
Screenshot (761)
image
image


Fix any responsiveness-related issues

KarandeepLubana (changes requested):
Overall great start! There are some issues that need to be addressed before we can merge.

hackathon_site/dashboard/frontend/src/components/general/OrderTables/OrderTables.module.scss
```diff
@@ -71,8 +89,8 @@
 }
 
 .chipPadding {
```
KarandeepLubana: This can be removed

hackathon_site/dashboard/frontend/src/components/general/OrderTables/OrderTables.tsx
```diff
@@ -96,41 +96,33 @@ export const GeneralOrderTableTitle = ({
...
+        {orderStatus && <ChipStatus status={orderStatus} />}
 
         {createdTime && updatedTime ? (
-            <Container>
+            <div style={{ alignSelf: "flex-end" }}>
                 <Chip
                     label={[<b>Created at: </b>, formatDateTime(createdTime)]}
                     icon={<EditIcon />}
                     className={`${styles.chipPurple} ${styles.chip} ${
                         additionalChipFormatting ? styles.chipPadding : ""
```
KarandeepLubana: I believe this can be simplified since `styles.chipPadding` doesn't do anything

hackathon_site/dashboard/frontend/src/components/general/OrderTables/OrderTables.tsx
```diff
@@ -96,41 +96,33 @@ export const GeneralOrderTableTitle = ({
...
                     }`}
                 />
-                {"    "}
+                {/* {"    "} */}
                 <Chip
                     label={[<b>Updated at: </b>, formatDateTime(updatedTime)]}
                     icon={<UpdateIcon />}
                     className={`${styles.chipBlue} ${styles.chip} ${
                         additionalChipFormatting ? styles.chipPadding : ""
```
KarandeepLubana: This can be simplified. (Same as previous comment)

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,404 @@
...
+                                                                                            )}
+                                                                                        </RadioGroup>
+                                                                                    </>
+                                                                                )}
+                                                                            </Field>
+                                                                            <FormHelperText>
+                                                                                {
+                                                                                    item.helperText
+                                                                                }
+                                                                                {/* {errors?.state} */}
```
KarandeepLubana: You can remove this

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,404 @@
...
+    const initialValues = {
+        state: "",
+        qty: "",
+        what: "",
+        when: "",
+        where: "",
+    };
+
+    const handleSubmit = async (values: FormikValues, { resetForm }: any) => {
+        console.log(values);
```
KarandeepLubana: This can be removed

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,404 @@
...
+                                            onSubmit={handleSubmit}
+                                            className={styles.form}
+                                        >
+                                            <div className={styles.formContainer}>
+                                                {incidentFormComponents.map(
+                                                    (item, index) => {
+                                                        return (
+                                                            <div
+                                                                key={`${item.type}-${item.name}`}
+                                                                style={{
```
KarandeepLubana: Avoid using inline styling

KarandeepLubana:
image
We can have a little bit of space between `Qty` and line `Number of Grove temperature and humidity sensor pro affected.
`

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import { fireEvent, render, screen, waitFor } from "testing/utils";
+import IncidentForm from "./IncidentForm";
+import { INCIDENT_ERROR_MSG } from "./IncidentForm";
+
+describe("<Incident Form />", () => {
+    test("Renders without crashing", () => {
+        render(<IncidentForm />);
+        expect(screen.getByText("Item Incident Form")).toBeInTheDocument();
```
KarandeepLubana: You can also test if other components of the form are present

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,419 @@
...
+                options: {
+                    variant: "success",
+                },
+            })
+        );
+        // navigate back to previous page
+        history.goBack();
+    };
+
+    const createQuantityList = (number: number) => {
```
KarandeepLubana: This can be refactored. Both if/else cases do the same thing.

KarandeepLubana: What is supposed to happen if  `hardwareQuantity` is 0?

natapokie: @KarandeepLubana thanks for the feedback!

i kind of see what you mean with the if/else cases... but one of them only displays 0 while the other displays the quantity from 1 to the requested quantity.
just thought that the 0 shouldn't be displayed because they would have some quantity if they're reporting an incident.

hardwareQuantity would never be 0, the 0 state is just used when someone accidentally navigates to the page directly without clicking the incident form link in the table. i just added another commit to prevent submission in case that occurs!

hope that clarifies things!

KarandeepLubana: This is a common scenario where you don’t want users to have access to this route until they’ve added an order. You want to prevent users from directly accessing the route by entering the URL. 

Conditional redirects should be used to redirect users to a different page or display an error message when they try to access routes that should only be reached through specific actions or links. I suggest you implement a solution that uses conditional redirects.

Useful Links:
https://www.toptal.com/react/react-router-tutorial 
https://stackabuse.com/redirects-in-react-router/

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,422 @@
...
+                                                                                    getContentAnchorEl:
+                                                                                        null,
+                                                                                }}
+                                                                                autoWidth
+                                                                                data-testid={
+                                                                                    item.testId
+                                                                                }
+                                                                            >
+                                                                                {createQuantityList(
+                                                                                    Number(
```
KarandeepLubana: `hardwareQuantity` is already of type number. No need to convert it again.

KarandeepLubana:
The "Report Broken/Lost" button has a discoverability issue as users may not realize it exists unless they scroll to the left. Consider redesigning the UI to address this concern, unless the current UI intentionally intends for the button to be less prominent.

image

natapokie:
> The "Report Broken/Lost" button has a discoverability issue as users may not realize it exists unless they scroll to the left. Consider redesigning the UI to address this concern, unless the current UI intentionally intends for the button to be less prominent.
> 
> 
> 
> image
> 
> 

will address this in another issue! 👍

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.test.tsx
```diff
@@ -0,0 +1,92 @@
+import React from "react";
+import { render, screen } from "testing/utils"; // Import your testing utilities here
+import IncidentForm from "./IncidentForm";
+import { BrowserRouter } from "react-router-dom";
+import { Provider } from "react-redux";
+import configureStore, { MockStore } from "redux-mock-store";
+
+// describe("<Incident Form />", () => {
```
KarandeepLubana: The commented code can be removed

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
```diff
@@ -0,0 +1,463 @@
...
+    qty: "",
+    what: "",
+    when: "",
+    where: "",
+};
+
+const createQuantityList = (number: number) => {
+    let entry = [];
+    if (number === 0) {
+        entry.push(
```
KarandeepLubana, line 68: This can be refactored since the 0 case is unnecessary now

KarandeepLubana (dismissed):

hackathon_site/dashboard/frontend/src/pages/IncidentForm/IncidentForm.tsx
natapokie: thanks @KarandeepLubana !

KarandeepLubana (approved):

Mustaballer (approved):

PR #505, opened 2023-06-19
Add popupmodal into TeamActionTable

ishika1202:
## Overview

- Resolves #IEEE-197
- Implemented delete team feature 
-  Added extra warning if the team has team members in it

## Unit Tests Created

- Tests that check UI and redux for delete team feature 

## Steps to QA

- comments


lukecheseldine (dismissed):
hey Ishika, this is really good, you're almost there!

i noticed though that the delete doesn't work if the team has unreturned orders. i tried to delete the team below (with unreturned orders) and it failed. this is intentional design, but your feature doesn't handle the error well. see the screenshot below

Screen Shot 2023-07-22 at 10 23 19 AM

1. can we dispatch a more helpful snackbar message so the user knows why deletion failed? you can use the error message returned in the 400 response.
2. if this error occurs, can we stay on the team detail page of the team that they tried to delete? what happens now is: you're on the team detail page, you press delete, it doesn't delete and you are sent back to the team list page, where the snackbar is displayed. let's keep the user on the team detail page, so that they can easily resolve the unreturned orders and try to delete the team again if that's what they want to do.

please write a test case for the snackbar message too (don't need to test that they stay on the team detail page)

lmk if you need any help, great job! :D

Mustaballer:
@ishika1202 What's latest on this?

Mustaballer (approved):

liu-samuel (approved):

PR #513, opened 2024-01-02
IEEE-287 (adding new order statuses)

carmen-chau:
## Overview

This PR is preliminary work for IEEE-287.

The current list of order statuses are as follows:

- Submitted [Pre-existing] (orders that teams have submitted, but tech team has not looked at yet)
- Ready for Pickup [Pre-existing] (order has been packaged by tech team, ready for pick up)
- Picked Up [Pre-existing] (team has received order) 
- Cancelled [Pre-existing] (Note: In current main branch, orders that were cancelled by either tech team or hacking team are grouped in 1 category. Do we want to seperate these 2 cases?)
- Returned [Pre-existing]
- **Pending [NEW] (Denote orders that are currently being packed). There is probably a better name for this (Eg: "Packing?)**
- **Rejected [NEW] (Denotes orders manually denied by IEEE team)**

### Work done so far:

1. I've re-coloured the Updated Time UI tag on the Admin/Participant table to be pink instead of blue (to avoid colour confusion with the Pending tag) 

2. I've added the Pending tag to all currently packing orders on Admin/Participant table 

3. Made sure that the Pending and Rejected status are in the Admin Order page (via the OrderFilter component)

4. I've modified the code for Django admin site accordingly to include both the Pending and Rejected status 

### Work left to be done 

1. Discuss and agree to new order status names **[DONE]**
3. Finish code for button to transition Pending orders to Ready for Pickup **[DONE]**
4. Potentially change other files I missed**[DONE]**
5. Modify and/or write new unit tests

## Unit Tests Created

- N/A, most likely need to modify mockData.tsx to pass all tests 

## Known bugs

1. No matter how many times I attempt to refresh, on the admin side, the order filter number count for: Ready for Pickup, Returned, Pending, Rejected are always 0

2. The `Handles double row click event` test in `OrdersTable.tsx` is failing, unsure why 

### Screenshots of changes on HSS site 

1. On Admin side 

Screenshot 2024-01-02 at 1 38 56 AM

3. On participant side 
Screenshot 2024-01-02 at 1 40 49 AM

4. On Django site 

Screenshot 2024-01-02 at 1 41 30 AM


Mustaballer:
Hey @carmen-chau, great start! I will be addressing your next steps below.

1. Regarding the new status names the ones you made are great, I think its fine if we remove rejected because cancelled is good for both cases. So rejected tag isn't needed imo.
2. I'm wondering if you are able to change the order status on the Orders page?(Edit dw about this)
3. For the bug we will probably need to modify some code related to the Orders Filters, but you can put this bug on hold for now and work on the things you can do rn.

hackathon_site/hardware/serializers.py
```diff
@@ -160,7 +160,7 @@ def get_team_code(obj: Order):
 
 class OrderChangeSerializer(OrderListSerializer):
     change_options = {
-        "Submitted": ["Cancelled", "Ready for Pickup"],
+        "Submitted": ["Cancelled", "Ready for Pickup"], # TODO: Unsure about whether I need to add the status "Pending" and "Rejected" to the list
```
Mustaballer: Yeah you should add Pending Status here, I believe it means from submitted status it can changed to  cancelled or ready for pickup. I'm thinking now that instead of pending we call it "Preparing" so that means admins are actively building the order.

hackathon_site/dashboard/frontend/src/api/helpers.ts
```diff
@@ -182,11 +184,14 @@ export const sortCheckedOutOrders = (
...
     let ready_orders = [];
     let submitted_orders = [];
+    let packing_orders = []; // Added new array to ensure sorting of orders being packed accomodates for the orders with "packing" tag. This likely helps with the order display on the admin + participant table
     for (let order of orders) {
         if (order.status === "Ready for Pickup") {
             ready_orders.push(order);
-        } else {
+        } else if (order.status === "Submitted") {
             submitted_orders.push(order);
+        } else {
```
Mustaballer: change to ```else if (order.status === "Packing")```

carmen-chau:
Current bugs with the ticket:

1. On the participant side, the "Cancel Order" button for "Packing Orders" does not work. If you try pressing the "Delete Order" button after the Pop-up modal, the following pop-up will show. I know this has something to do with the `cancelOrderThunk` in `orderSlice.ts` but am unsure of where to go from here. 

Screenshot 2024-01-14 at 7 18 10 PM

2. The Swagger Endpoint for `/hardware/orders/{id}/` doesn't _exactly_ work when you try to change the status of a packing order to "Ready for Pickup". Although the order status **does**  end up changing, you likely will get an error when running the website (something to do with `orders.foreach()`? 
Screenshot 2024-01-14 at 7 19 33 PM

Screenshot 2024-01-14 at 7 27 09 PM

3. On the same topic/endpoint as the 2nd point, this is the error message when changing order from "Packing" => "Cancelled"
Screenshot 2024-01-14 at 7 25 25 PM

